PR #125, opened 2022-10-25
Add support for nested_type field in tjson Converter

### LINKED ISSUE #121: panic: cannot generate crd for resource
### What happened?
I'm trying to convert private cloud provider to upjet provider. 
but on generation stage i am facing error:
panic: cannot generate crd for resource salt_compute_instance: cannot build types for Instance: cannot build the Types: cannot infer type from schema of field disks: invalid schema type TypeInvalid
there is my  config/schema.json file :
https://pastebin.com/kiM1GMq6

### How can we reproduce it?
i'll prepare custom repo with Makefile if it's neeeded, cause of private terraform provider  this may take some time ( and i'm not sure if it needed)

I will be overly grateful if you tell me how to solve this error, if the problem is in the json that the provider generates, please tell me what exactly is wrong.

turkenh:

It looks like we never had a provider schema using `nested_type` field instead if using `type` field for a collection type.
This PR fixes #121 by adding support for `nested_type` in tfjson converter code that we are using to convert tfjson schema to sdk v2 schema.

Fixes #121 

I have:

I could successfully generate all resources using the schema provided in #121.
I have also verified that consuming this change does not cause any difference in the generated schema for `upbound/provider-aws`,` upbound/provider-gcp` and `upbound/provider-azure` to make sure that we are not breaking something there.

### DIFF
--- a/pkg/types/conversion/tfjson/tfjson.go
+++ b/pkg/types/conversion/tfjson/tfjson.go
@@ -83,10 +83,19 @@ func tfJSONAttributeToV2Schema(attr *tfjson.SchemaAttribute) *schemav2.Schema {
 		Deprecated:  deprecatedMessage(attr.Deprecated),
 		Sensitive:   attr.Sensitive,
 	}
-	if err := schemaV2TypeFromCtyType(attr.AttributeType, v2sch); err != nil {
-		panic(err)
+	if attr.AttributeNestedType != nil {
+		if err := schemaV2TypeFromNestedType(attr.AttributeNestedType, v2sch); err != nil {
+			panic(err)
+		}
+		return v2sch
 	}
-	return v2sch
+	if !attr.AttributeType.Equals(cty.NilType) {
+		if err := schemaV2TypeFromCtyType(attr.AttributeType, v2sch); err != nil {
+			panic(err)
+		}
+		return v2sch
+	}
+	panic("attribute neither has \"type\" nor \"nested-type\" defined")
 }
 
 func tfJSONBlockTypeToV2Schema(nb *tfjson.SchemaBlockType) *schemav2.Schema { //nolint:gocyclo
@@ -145,6 +154,36 @@ func tfJSONBlockTypeToV2Schema(nb *tfjson.SchemaBlockType) *schemav2.Schema { //
 	}
 	v2sch.Elem = res
 	return v2sch
+}
+
+func schemaV2TypeFromNestedType(ntyp *tfjson.SchemaNestedAttributeType, schema *schemav2.Schema) error {
+	switch ntyp.NestingMode {
+	case tfjson.SchemaNestingModeSingle:
+		schema.Type = schemav2.TypeList
+		schema.MinItems = 0
+		schema.MaxItems = 1
+	case tfjson.SchemaNestingModeGroup:
+		schema.Type = schemav2.TypeList
+		schema.MinItems = 1
+		schema.MaxItems = 1
+	case tfjson.SchemaNestingModeList:
+		schema.Type = schemav2.TypeList
+	case tfjson.SchemaNestingModeSet:
+		schema.Type = schemav2.TypeSet
+	case tfjson.SchemaNestingModeMap:
+		schema.Type = schemav2.TypeMap
+	default:
+		return errors.Errorf("unknown nesting mode: %s", ntyp.NestingMode)
+	}
+
+	res := &schemav2.Resource{}
+	res.Schema = make(map[string]*schemav2.Schema, len(ntyp.Attributes))
+	for key, attr := range ntyp.Attributes {
+		res.Schema[key] = tfJSONAttributeToV2Schema(attr)
+	}
+	schema.Elem = res
+	schema.ConfigMode = schemav2.SchemaConfigModeAttr
+	return nil
 }
 
 func schemaV2TypeFromCtyType(typ cty.Type, schema *schemav2.Schema) error { //nolint:gocyclo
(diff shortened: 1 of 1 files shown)


pkg/types/conversion/tfjson/tfjson.go
```diff
@@ -83,10 +83,19 @@ func tfJSONAttributeToV2Schema(attr *tfjson.SchemaAttribute) *schemav2.Schema {
...
+		return v2sch
 	}
-	return v2sch
+	if !attr.AttributeType.Equals(cty.NilType) {
+		if err := schemaV2TypeFromCtyType(attr.AttributeType, v2sch); err != nil {
+			panic(err)
+		}
+		return v2sch
+	}
+	panic("attribute neither has \"type\" nor \"nested-type\" defined")
```
ulucinar, line 98: nit:
```suggestion
	panic(`attribute neither has "type" nor "nested-type" defined`)
```

pkg/types/conversion/tfjson/tfjson.go
```diff
@@ -83,10 +83,19 @@ func tfJSONAttributeToV2Schema(attr *tfjson.SchemaAttribute) *schemav2.Schema {
...
-	if err := schemaV2TypeFromCtyType(attr.AttributeType, v2sch); err != nil {
-		panic(err)
+	if attr.AttributeNestedType != nil {
+		if err := schemaV2TypeFromNestedType(attr.AttributeNestedType, v2sch); err != nil {
+			panic(err)
+		}
+		return v2sch
 	}
-	return v2sch
+	if !attr.AttributeType.Equals(cty.NilType) {
```
ulucinar, line 92: Does `cty.NilType` mean that no attribute type has been set?

pkg/types/conversion/tfjson/tfjson.go
```diff
@@ -147,6 +156,36 @@ func tfJSONBlockTypeToV2Schema(nb *tfjson.SchemaBlockType) *schemav2.Schema { //
 	return v2sch
 }
 
+func schemaV2TypeFromNestedType(ntyp *tfjson.SchemaNestedAttributeType, schema *schemav2.Schema) error {
+	switch ntyp.NestingMode {
```
ulucinar, line 160: nit: Could you please leave a comment explaining the source of truth for these conversion cases?

pkg/types/conversion/tfjson/tfjson.go
```diff
@@ -147,6 +156,36 @@ func tfJSONBlockTypeToV2Schema(nb *tfjson.SchemaBlockType) *schemav2.Schema { //
...
+		schema.MinItems = 1
+		schema.MaxItems = 1
+	case tfjson.SchemaNestingModeList:
+		schema.Type = schemav2.TypeList
+	case tfjson.SchemaNestingModeSet:
+		schema.Type = schemav2.TypeSet
+	case tfjson.SchemaNestingModeMap:
+		schema.Type = schemav2.TypeMap
+	default:
+		return errors.Errorf("unknown nesting mode: %s", ntyp.NestingMode)
```
ulucinar, line 176: Especially, if the nesting mode string turns out to be an empty string:
nit:
```suggestion
		return errors.Errorf("unknown nesting mode: %q", ntyp.NestingMode)
```

ulucinar (approved):
Thanks @turkenh, lgtm.

ulucinar:
@sergenyalcin, could you please trigger the CLA bot's job once more. Maybe removing & adding this repo back might help?

IvanTurgenev1:
@turkenh found a bug with nested_type: single:
example:
there is part of config.schema: 
```
 "floating_ip": {
                                "nested_type": {
                                    "attributes": {
                                        "allocate_new": {
                                            "type": "bool",
                                            "description_kind": "plain",
                                            "required": true
                                        },
                                        "description": {
                                            "type": "string",
                                            "description_kind": "plain",
                                            "optional": true
                                        }
                                    },
                                    "nesting_mode": "single"
                                },
 ```

and part of generated crd: 
```
 floatingIp:
                    items:
                      properties:
                        allocateNew:
                          type: boolean
                        description:
                          type: string
                      required:
                      - allocateNew
                      type: object
                    type: array
                    ```
                    
but  with nested_type single it's should be object, not array

and main.tf.json should be 
```
        "floating_ip": 
          {
            "allocate_new": true,
            "description": "123"
          }
        ,
```
not like this 
```
        "floating_ip": [
          {
            "allocate_new": true,
            "description": "123"
          }
        ],
```

pkg/types/conversion/tfjson/tfjson.go
turkenh: Yes, we didn't hit such a case until recently but it looks like it can be nil and only have `AttributeNestedType`.

turkenh:
@IvanTurgenev1 this was my understanding from the json documentation [here](https://github.com/hashicorp/terraform-json/blob/e6b203c3cb12469bcf829be00d16379c595008ed/schemas.go#L131). 
I would need to dig further to see what we need here.

IvanTurgenev1:
> @IvanTurgenev1 this was my understanding from the json documentation [here](https://github.com/hashicorp/terraform-json/blob/e6b203c3cb12469bcf829be00d16379c595008ed/schemas.go#L131). I would need to dig further to see what we need here.
Yeah, but this implementation is relevant for old version of sdk v2
In new terraform sdk netsed_type single should be object..
https://github.com/hashicorp/terraform-plugin-framework/issues/85

turkenh:
Digging further, I believe we have a more fundamental issue here, which is, `Upjet` is currently based on `sdkv2`, whereas this field, `nested_type` is introduced with the [terraform plugin framework](https://github.com/hashicorp/terraform-plugin-framework) which is an alternative to `sdkv2`: https://developer.hashicorp.com/terraform/plugin/which-sdk

So, I believe adding support for this would require more work than a simple conversion trick as I attempted here. 

Closing this PR in favor of this issue: https://github.com/upbound/upjet/issues/130

PR #135, opened 2022-11-06
types: slices and maps of secret key selector should not have pointer

muvaf:

`*map[string]v1.SecretKeySelector` and `*[]v1.SecretKeySelector` are not compatible with marshaling and slices & maps are already pointers, so, no need to make them pointer of pointers to have them optional in OpenAPI spec. The only functional change in this PR is to not make those two cases pointer but only `string` one.

Also opened https://github.com/upbound/upjet/issues/134 since map with struct value type is actually anti-pattern. So we should decide whether we should switch to something else now or later. @turkenh Do you know of CRDs that have sensitive map values like this one in other providers? I wonder how much breaking change we'd inflict if we were to switch to either of the options proposed in the issue.

Part of https://github.com/upbound/upjet/issues/126

I have:

Confirmed that https://github.com/upbound/upjet/issues/126 is fixed with this PR. It also results in rename & retype of `PasswordSecretRef *[]v1.SecretKeySelector` in `User` of `elasticache` to `PasswordSecretRefs []v1.SecretKeySelector`, which is a breaking change in the schema due to rename. I need to check if it was working in the first place.

### DIFF
--- a/pkg/types/field.go
+++ b/pkg/types/field.go
@@ -150,31 +150,34 @@ func NewSensitiveField(g *Builder, cfg *config.Resource, r *resource, sch *schem
 		// Data will be stored in connection details secret
 		return nil, true, nil
 	}
-	sfx := "SecretRef"
-	cfg.Sensitive.AddFieldPath(fieldPathWithWildcard(f.TerraformPaths), "spec.forProvider."+fieldPathWithWildcard(f.CRDPaths)+sfx)
-	// todo(turkenh): do we need to support other field types as sensitive?
-	if f.FieldType.String() != "string" && f.FieldType.String() != "*string" && f.FieldType.String() != "[]string" &&
-		f.FieldType.String() != "[]*string" && f.FieldType.String() != "map[string]string" && f.FieldType.String() != "map[string]*string" {
+
+	f.TFTag = "-"
+	var sfx string
+	switch f.FieldType.String() {
+	case "string", "*string":
+		f.FieldType = typeSecretKeySelector
+		if f.Schema.Optional {
+			f.FieldType = types.NewPointer(f.FieldType)
+		}
+		sfx = "SecretRef"
+	case "[]string", "[]*string":
+		f.FieldType = types.NewSlice(typeSecretKeySelector)
+		sfx = "SecretRefs"
+	case "map[string]string", "map[string]*string":
+		f.FieldType = types.NewMap(types.Universe.Lookup("string").Type(), typeSecretKeySelector)
+		sfx = "SecretRefMap"
+	default:
+		// todo(turkenh): do we need to support other field types as sensitive?
 		return nil, false, fmt.Errorf(`got type %q for field %q, only types "string", "*string", []string, []*string, "map[string]string" and "map[string]*string" supported as sensitive`, f.FieldType.String(), f.FieldNameCamel)
 	}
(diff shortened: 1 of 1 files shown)


turkenh:
> @turkenh Do you know of CRDs that have sensitive map values like this one in other providers? I wonder how much breaking change we'd inflict if we were to switch to either of the options proposed in the issue.

Not top of my mind. You can print effected resources by running generate against a custom change.

turkenh:
Looks like unit tests are failing with a relevant issue, can you check that?

muvaf:
@turkenh I'll run and see the number. Meanwhile, it'd be great if you can share your thoughts on https://github.com/upbound/upjet/issues/134 . Ideally, I'd like to switch to either of those options instead of having to go through a breaking change later with `aws_glue_connection`.

turkenh:
> Confirmed that https://github.com/upbound/upjet/issues/126 is fixed with this PR. It also results in rename & retype of PasswordSecretRef *[]v1.SecretKeySelector in User of elasticache to PasswordSecretRefs []v1.SecretKeySelector, which is a breaking change in the schema due to rename. I need to check if it was working in the first place.

@muvaf are there anything else you did to get the issue fixed with this PR? Because I cannot get it fixed with this PR as well:

```
panic: cannot run goimports for apis folder: ./glue/v1beta1/zz_connection_types.go:45:55: expected ';', found '/'
./glue/v1beta1/zz_connection_types.go:74:1: expected '}', found 'type'
: exit status 2
```

Generates the following
```
ConnectionPropertiesSecretRefMap map[string]github.com/crossplane/crossplane-runtime/apis/common/v1.SecretKeySelector%!(EXTRA string=v1) `json:"connectionPropertiesSecretRefMap,omitempty" tf:"-"`
```

srust:
Hi,

I am attempting to create a Linode provider using upjet and hitting this issue as well. Specifically for the "linode_instance" resource via https://registry.terraform.io/providers/linode/linode/latest/docs/resources/instance

The errors are similar to here:

```
14:22:23 [ .. ] go generate linux_amd64
panic: cannot run goimports for apis folder: ./instance/v1alpha1/zz_instance_types.go:226:48: expected ';', found '/'
./instance/v1alpha1/zz_instance_types.go:235:1: expected '}', found 'type'
./instance/v1alpha1/zz_instance_types.go:425:48: expected ';', found '/'
./instance/v1alpha1/zz_instance_types.go:454:1: expected '}', found 'type'
: exit status 2
```

```
  226 StackscriptDataSecretRef *map[string]github.com/crossplane/crossplane-runtime/apis/common/v1.SecretKeySelector%!(EXTRA string=_v1common) `json:"stackscriptDataSecretRef,omitempty" tf:"-      "`
  425 StackscriptDataSecretRef *map[string]github.com/crossplane/crossplane-runtime/apis/common/v1.SecretKeySelector%!(EXTRA string=_v1common) `json:"stackscriptDataSecretRef,omitempty" tf:"-      "`
```

In this case the StackscriptDataSecretRef becomes

  202     StackscriptDataSecretRefMap map[string]_v1.SecretKeySelector `json:"stackscriptDataSecretRefMap,omitempty" tf:"-"`
  389     StackscriptDataSecretRefMap map[string]_v1.SecretKeySelector `json:"stackscriptDataSecretRefMap,omitempty" tf:"-"`

But it is able to generate successfully with this patch.

turkenh:
Closing this since the underlying problem fixed by https://github.com/upbound/upjet/pull/138

PR #137, opened 2022-11-10
Add config.Provider.GetSkippedResourceNames

ulucinar:

This PR proposes to add a `config.Provider.GetSkippedResourceNames` method that returns a list of Terraform resource names that are available in the Terraform provider schema but that are either skipped or not included so that they are not generated as managed resources.

I've also considered:
- Directly generating the diff file when a path is configured (via provider configuration). But generating this diff, in my opinion, would be a supporting function (not among the core code generation pipelines), so instead of configuring a file path to store the diff and having Upjet generate it, I preferred a change to make Upjet be able to directly supply the required diff information
- Another approach would be not to touch Upjet but instead do the diff calculation in the providers themselves. In the provider, we already have the JSON schema from which we can extract the set of Terraform resources available and also the names of the Terraform resources for which we have generated the corresponding MRs (in the generated files under `apis`). But I found this approach too indirect and requiring further processing for a directly available piece of information. Thus I went with the small change proposed in this PR.

I have:

Tested against upbound/provider-aws.

### DIFF
--- a/pkg/config/provider.go
+++ b/pkg/config/provider.go
@@ -85,6 +85,12 @@ type Provider struct {
 	// can add "aws_waf.*" to the list.
 	SkipList []string
 
+	// skippedResourceNames is a list of Terraform resource names
+	// available in the Terraform provider schema, but
+	// not in the include list or in the skip list, meaning that
+	// the corresponding managed resources are not generated.
+	skippedResourceNames []string
+
 	// IncludeList is a list of regex for the Terraform resources to be
 	// included. For example, to include "aws_shield_protection_group" into
 	// the generated resources, one can add "aws_shield_protection_group$".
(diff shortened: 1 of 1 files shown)


pkg/config/provider.go
```diff
@@ -219,7 +227,8 @@ func NewProvider(schema []byte, prefix string, modulePath string, metadata []byt
 		if !matches(name, p.IncludeList) {
 			continue
 		}
-
+		// resource is to be generated, so remove it from the skipped list
+		p.skippedResourceNames = p.skippedResourceNames[:len(p.skippedResourceNames)-1]
```
turkenh, line 231: While this looked nonintuitive at first place, I could understand the why we did like that.

Another approach could be to combine if statements into one and only add the resource to skip like just before _continue_.

turkenh (approved):

pkg/config/provider.go
ulucinar: Done.

ulucinar:
Closing/reopening the PR in an attempt to retrigger the CLA check...

PR #138, opened 2022-11-11
Handle sensitive input maps via SecretReference

turkenh:

This PR fixes the issues we are having with sensitive input fields of type map by using a `SecretReference` as the data type.

Fixes https://github.com/upbound/upjet/issues/134

I have:

Tested with `aws_glue_connection` resource and [this PR](https://github.com/upbound/provider-aws/pull/142) using the below manifests:

```yaml
apiVersion: glue.aws.upbound.io/v1beta1
kind: Connection
metadata:
  annotations:
    meta.upbound.io/example-id: glue/v1beta1/connection
  labels:
    testing.upbound.io/example-name: example
  name: example
spec:
  forProvider:
    catalogId: "123456789012"
    connectionPropertiesSecretRef:
      name: example-secret
      namespace: upbound-system
    region: us-west-1
---
apiVersion: v1
kind: Secret
metadata:
  name: example-secret
  namespace: upbound-system
type: Opaque
stringData:
  JDBC_CONNECTION_URL: "jdbc:mysql://my-db/exampledatabase"
  PASSWORD: "examplepassword"
  USERNAME: "exampleusername"
```

### DIFF
--- a/pkg/resource/sensitive.go
+++ b/pkg/resource/sensitive.go
@@ -206,40 +206,38 @@ func GetSensitiveParameters(ctx context.Context, client SecretClient, from runti
 
 			switch k := v.(type) {
 			case map[string]any:
-				if hasMapValue(k) {
-					sel := &map[string]v1.SecretKeySelector{}
-					if err = pavedJSON.GetValueInto(expandedJSONPath, sel); err != nil {
+				_, ok := k["key"]
+				if !ok {
+					// This is a special case where we have a "SecretReference" without a selected "key". This happens
+					// when there is an input field of type map[string]string (or map[string]*string).
+					// In this case, we need to get the entire secret data and fill it in the terraform state as a map.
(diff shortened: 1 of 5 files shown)


muvaf (approved):
LGTM! Could you also drop a comment in regards to how many CRDs are affected by this change in the official providers and which ones would be breaking changes?

turkenh:
> LGTM! Could you also drop a comment in regards to how many CRDs are affected by this change in the official providers and which ones would be breaking changes?

@muvaf n**o CRDs effected** with this change since there were no configured resource with a sensitive input of type map since it was failing indeed. Double checked this by running make generate in all 3 providers consuming this change.

muvaf:
Thanks @turkenh !

eljohnson92:
@turkenh @muvaf would you expect this PR would also handle secrets that are arrays? we have a use case where we have an array we would like to mark the contents of as secrets

PR #148, opened 2023-01-04
Rename migration.Converter.ComposedTemplates as Composition

ulucinar:

This PR proposes changes to the `migration.Converter` interface so that:
- `migration.Converter.ComposedTemplates` is renamed to `Composition`
- `migration.Converter.Composition` accepts the `spec.patchSets` of a Composition, and 
- Allows conversion of the `spec.patchSets` of Composition's by the registered migration converters.

Also we have the following improvements for the migration converters:
  - Include references to patch sets in the set of patches for migration targets by default (assumption is patch sets are reusable and thus probably applicable to the migration targets, so include refs to them by default). Target schemas are checked and if a referenced patchset does not conform to the target's schema, it's removed from the set of patches.
  - Remove patches from targets if the patch target/source field does not exist or is of a different type in the target, by default
  - Include patches in targets if the patch target/source field exists and is of the same type in the target, by default

I have:

### DIFF
--- a/pkg/migration/converter.go
+++ b/pkg/migration/converter.go
@@ -78,6 +78,9 @@ func sanitizeResource(m map[string]any) map[string]any {
 	return m
 }
 
+// removeNilValuedKeys removes nil values from the specified map so that
+// the serialized manifest do not contain corresponding superfluous YAML
+// nulls.
 func removeNilValuedKeys(m map[string]interface{}) {
 	for k, v := range m {
 		if v == nil {
--- a/pkg/migration/converter.go
+++ b/pkg/migration/converter.go
@@ -74,7 +74,27 @@ func sanitizeResource(m map[string]any) map[string]any {
 	if len(metadata) == 0 {
 		delete(m, "metadata")
 	}
+	removeNilValuedKeys(m)
 	return m
+}
+
+func removeNilValuedKeys(m map[string]interface{}) {
+	for k, v := range m {
+		if v == nil {
+			delete(m, k)
+			continue
+		}
+		switch c := v.(type) {
+		case map[string]any:
+			removeNilValuedKeys(c)
+		case []any:
+			for _, e := range c {
+				if cm, ok := e.(map[string]interface{}); ok {
+					removeNilValuedKeys(cm)
+				}
+			}
+		}
+	}
 }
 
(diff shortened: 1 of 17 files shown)


ytsarev (approved):
Using the code from this PR, I successfully created a set of working converters, so I believe it is mergeable.

Thanks a lot for the outstanding work, @ulucinar!

pkg/migration/converter.go
```diff
@@ -74,9 +74,29 @@ func sanitizeResource(m map[string]any) map[string]any {
 	if len(metadata) == 0 {
 		delete(m, "metadata")
 	}
+	removeNilValuedKeys(m)
 	return m
 }
 
+func removeNilValuedKeys(m map[string]interface{}) {
```
sergenyalcin, line 81: As far as I understand, we are cleaning here by traversing the map. I guess this is something we do to avoid seeing nil values in the manifest.

pkg/migration/patches.go
```diff
@@ -0,0 +1,235 @@
...
+	// (like mismatches in quotes, etc.)
+	regexMap     = regexp.MustCompile(`(.+)\[["|']?\D+["|']?]`)
+	regexJSONTag = regexp.MustCompile(`([^,]+)?(,.+)?`)
+)
+
+const (
+	jsonTagInlined = ",inline"
+)
+
+func removeInvalidPatches(c runtime.ObjectCreater, gvkSource, gvkTarget schema.GroupVersionKind, patchSets []xpv1.PatchSet, targetTemplate *xpv1.ComposedTemplate) error {
```
sergenyalcin, line 41: After reading the code, I think I understand what it's doing. However, even if we are not exporting, we can leave comments that tell roughly what the functions do

sergenyalcin (approved):
Thanks @ulucinar I left comments about documentation.

pkg/migration/converter.go
ulucinar: Yes, correct. Added a comment explaining the purpose.

pkg/migration/patches.go
ulucinar: Thanks @sergenyalcin, this is a good point. Documented each function in the patch-filtering library.

PR #149, opened 2023-01-19
Add a test-case for config.GetExternalNameFromTemplated where the external-name template variable is not available

ulucinar:

This PR adds a new test-case for `config.GetExternalNameFromTemplated`, where the template expression passed as a parameter to the function, does not contain the `{{ .external_name }}` variable. The expected behavior in this case is that the ID string is returned intact, i.e., the ID is the external-name.

I have:

`make reviewable`


sergenyalcin (approved):
Thanks @ulucinar LGTM!

PR #150, opened 2023-01-19
Add check for native provider upgrade

sergenyalcin:

I have:

This PR was tested on provider-azure by using this change.

- Created a ResourceGroup in azure 3.8.0 version
- Then the provider version was upgraded to 3.38.0
- Then checked everything is okay.

### DIFF
--- a/pkg/terraform/store.go
+++ b/pkg/terraform/store.go
@@ -158,9 +158,9 @@ func (ws *WorkspaceStore) Workspace(ctx context.Context, c resource.SecretClient
 	if err := fp.EnsureTFState(ctx); err != nil {
 		return nil, errors.Wrap(err, "cannot ensure tfstate file")
 	}
-	isNeedProviderUpgrade, err := fp.isNeedProviderUpgrade()
+	isNeedProviderUpgrade, err := fp.needProviderUpgrade()
 	if err != nil {
-		return nil, errors.Wrap(err, "cannot upgrade provider")
+		return nil, errors.Wrap(err, "cannot check if a Terraform dependency update is required")
 	}
 	if err := fp.WriteMainTF(); err != nil {
 		return nil, errors.Wrap(err, "cannot write main tf file")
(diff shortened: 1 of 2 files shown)


pkg/terraform/store.go
```diff
@@ -158,9 +158,22 @@ func (ws *WorkspaceStore) Workspace(ctx context.Context, c resource.SecretClient
...
+		return nil, errors.Wrap(err, "cannot upgrade provider")
+	}
 	if err := fp.WriteMainTF(); err != nil {
 		return nil, errors.Wrap(err, "cannot write main tf file")
 	}
+	if isNeedProviderUpgrade {
+		cmd := w.executor.CommandContext(ctx, "terraform", "init", "-upgrade", "-input=false")
+		cmd.SetDir(w.dir)
+		out, err := cmd.CombinedOutput()
+		w.logger.Debug("init ended", "out", string(out))
```
ulucinar, line 172: nit: We had better distinguish this debug log with the regular `terraform init`'s debug log to prevent ambiguity (if the Terraform CLI output already contains this information, then we can also rely on that):
```suggestion
		w.logger.Debug("init -upgrade ended", "out", string(out))
```

ulucinar (commented):
Thanks @sergenyalcin for working on this. I think we are good to go with this PR after handling one error case mentioned in the comments. 

Just to have discussed some alternatives:
One thing to consider is previously, if `main.tf.json` becomes corrupt, we are able to recover as we are overwriting it. But now, we will not be able to recover. I think we have preferred to parse the provider version from `main.tf.json` because we have the JSON unmarshaler at our disposal. 

One alternative is to parse the version from the `.terraform.lock.hcl` file. But this file is in HCL syntax. Running something like `grep version .terraform.lock.hcl` is sufficient but how robust is this and we had better avoid a new fork with each workspace reconciliation. If HCL parsing libraries are at our disposal in the repo, then we can easily & reliably parse the locked version from the lock file. 

Another alternative is to run:
```shell
❯ terraform version -json
{
  "terraform_version": "1.2.1",
  "platform": "darwin_arm64",
  "provider_selections": {
    "registry.terraform.io/hashicorp/azurerm": "3.38.0"
  },
  "terraform_outdated": true
}
```
, and to parse the version from the output JSON document. But, again this involves another fork at each workspace reconciliation, which we had better avoid.

ulucinar (approved):
Thanks @sergenyalcin, lgtm.

PR #164, opened 2023-02-24
Include full state under status.atProvider

turkenh:

This PR changes schema generation to include the entire state under `status.atProvider` as proposed in the [Observe Only design doc](https://github.com/crossplane/crossplane/pull/3531). In other words, `status.atProvider` will become superset of `spec.forProvider`.

I have:

See this PR https://github.com/upbound/provider-aws/pull/571


pkg/types/builder.go
```diff
@@ -277,6 +277,12 @@ func (r *resource) addParameterField(f *Field, field *types.Var) {
 }
 
 func (r *resource) addObservationField(f *Field, field *types.Var) {
+	for _, obsF := range r.obsFields {
+		if obsF.Name() == field.Name() {
+			// If the field is already added, we don't add it again.
```
ulucinar, line 282: Why do observation fields get added multiple times? Is it related to [this code segment](https://github.com/upbound/upjet/blob/b89baca4ae24c8fbd8eb403c353ca18916093e5e/pkg/types/builder.go#L206)? The comment above that segment states that it prevents elimination of fields in the observation type. If we are unconditionally adding every field as an observation field, can we remove that segment? And if we did so, would we still need this duplication check?

ulucinar (commented):
Thanks @turkenh. Left some comments for you to consider.

pkg/types/field.go
```diff
@@ -204,10 +204,12 @@ func (f *Field) AddToResource(g *Builder, r *resource, typeNames *TypeNames) {
...
 	field := types.NewField(token.NoPos, g.Package, f.FieldNameCamel, f.FieldType, false)
-	switch {
-	case IsObservation(f.Schema):
-		r.addObservationField(f, field)
-	default:
+
+	// Note(turkenh): We want atProvider to be a superset of forProvider, so
+	// we always add the field as an observation field and then add it as a
+	// parameter field if it's not an observation (only) field, i.e. parameter.
+	r.addObservationField(f, field)
```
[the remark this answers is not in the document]
turkenh: Good call! 
Yes, it doesn't make sense to include the field as an observation in that case. I'll update the PR accordingly.

pkg/types/builder.go
turkenh: Removing that code segment and the check here didn't work OOB. Ended up with diffs like below:

Screen Shot 2023-04-05 at 17 21 26

I need to investigate further and remember where those previous additions are coming from.

sergenyalcin (approved):
Thanks @turkenh LGTM! About @ulucinar's [comment](https://github.com/upbound/upjet/pull/164/files#r1157967752) I have not a strong insight. But if you will continue to discussing on this, we may also wait his approval.

pkg/types/builder.go
turkenh: I added a comment on where those previous additions are coming from.

ulucinar: Thank you @turkenh for giving it a try. I believe we can rename those types into common ones once we understand the implications as they are now shared between `spec.forProvider` and `status.atProvider`, when the `status.atProvider` becomes a superset.

ulucinar (approved):
Thank you @turkenh, lgtm.

PR #170, opened 2023-03-03
Add upjet runtime Prometheus metrics

### LINKED ISSUE #167: Expose Prometheus Metrics from Upjet Runtime

In the context of https://github.com/upbound/provider-aws/issues/325, we need more visibility especially into the scheduler runtime component of upjet.

We could expose some Prometheus metrics from upjet runtime. Some suggested metrics are:
- \# of Terraform CLI + Terraform provider processes running
- \# of Terraform workspaces being actively reconciled at a moment
- Duration of each Terraform CLI invocation
- ttr

ulucinar:

Fixes #167 

This PR adds the following Prometheus metrics to the upjet runtime. These are upjet runtime metrics, meaning that they are exposed by a provider while reconciling its managed resources via upjet:
- `upjet_terraform_cli_duration`: This is a histogram metric and reports statistics, in seconds, on how long it takes a Terraform CLI invocation to complete.
- `upjet_terraform_active_cli_invocations`: This is a gauge metric and it's the number of active (running) Terraform CLI invocations.
- `upjet_terraform_running_processes`: This is a gauge metric and it's the number of running Terraform CLI and Terraform provider processes.
- `upjet_resource_ttr`: This is a histogram metric and it measures, in seconds, the time-to-readiness for managed resources.

`terraform.Operation.MarkStart` now atomically checks for any previous ongoing operation before starting a new one, and
`terraform.Operation.{Start,End}Time` no longer return pointers that could potentially be used to modify the shared state outside of critical sections.

The following labels are available for the exposed runtime metrics:
- `upjet_terraform_cli_duration`: `subcommand` and `mode`.
  - `subcommand`: The `terraform` subcommand that's run, e.g., `init`, `apply`, `plan`, `destroy`, etc.
  - `mode`: The execution mode of the Terraform CLI, one of `sync` (so that the CLI was invoked synchronously as part of a reconcile loop), `async` (so that the CLI was invoked asynchronously, the reconciler goroutine will poll and collect results in future).
- `upjet_terraform_active_cli_invocations`:  `subcommand` and `mode`.
  - `subcommand`: The `terraform` subcommand that's run, e.g., `init`, `apply`, `plan`, `destroy`, etc.
  - `mode`: The execution mode of the Terraform CLI, one of `sync` (so that the CLI was invoked synchronously as part of a reconcile loop), `async` (so that the CLI was invoked asynchronously, the reconciler goroutine will poll and collect results in future).
- `upjet_terraform_running_processes`: `type`
  - `type`: Either `cli` for Terraform CLI (the `terraform` process) processes or `provider` for the Terraform provider processes. Please note that this is a best effort metric that may not be able to precisely catch & report all relevant processes. We may, in the future, improve this if needed by for example watching the `fork` system calls. But currently, it may prove to be useful to watch rouge Terraform provider processes.
- `upjet_resource_ttr`: `group`, `version`, `kind`
  - `group`, `version`, `kind` labels record the [API group, version and kind](https://kubernetes.io/docs/reference/using-api/api-concepts/) for the managed resource, whose [time-to-readiness](https://github.com/crossplane/terrajet/issues/55#issuecomment-929494212) measurement is captured.

### Notes on the concurrency-related changes:
1. `terraform.Operation.MarkStart` now atomically checks for ongoing async operations and reserves the "operation slot" (by recording the start time): We were previously checking whether there's an ongoing async operation in a critical section, exiting out of the critical section and then entering another section where we do the reservation like follows:
```go
	if w.LastOperation.IsRunning() {
		return errors.Errorf("%s operation that started at %s is still running", w.LastOperation.Type, w.LastOperation.StartTime().String())
	}
	w.LastOperation.MarkStart("apply")
```
From a theoretical perspective this does not look right but in fact, [the above section is never executed by two concurrent goroutines (on the same operation)](https://github.com/kubernetes/client-go/blob/bfa72fd2d36785ecf275bfcb138cfe881fa46bb1/util/workqueue/doc.go#L20) and thus is safe, as long as the controller-runtime behaves according to this assumption. But nevertheless, this PR proposes to change `MarkStart` so that it atomically checks and reserves the slot because:
  - It's conceptually simpler. The relevant section above now looks like:
```golang
	if !w.LastOperation.MarkStart("apply") {
		return errors.Errorf("%s operation that started at %s is still running", w.LastOperation.Type, w.LastOperation.StartTime().String())
	}
``` 
  - It's easier to reason about and to prove its correctness as we align with the theory. You don't need to make assumptions about how the controller-runtime behaves and in the very unlikely case that this assumption does not hold (because of a bug in client-go or controller-runtime, or because of a change in upjet), we will still be safe. 

2. `terraform.Operation.{Start,End}Time` no longer return pointers that could potentially be used to modify the shared state outside of critical sections: Not sure if this has practical implications but again from a theoretical point of view, it's good practice to read the data in a critical section, make a copy of it, and return that snapshot copy so that its clients will not have a chance to modify the shared state outside of a critical section. 

I have:

- [ ] We'd also like to document what custom metrics are exposed from upjet runtime

10 `userpool.cognitoidp` resources from `upbound/provider-aws` were provisioned, reconciled with a poll interval of 1m twice after acquiring the `Ready=True` status condition, and they were finally destroyed. Here are some sample screenshots from the Prometheus UI:

- `upjet_terraform_active_cli_invocations` gauge metric showing the sync & async `terraform init/apply/plan/destroy` invocations:
image

- `upjet_terraform_running_processes` gauge metric showing both `cli` and `provider` labels:
image

- `upjet_terraform_cli_duration` histogram metric, showing average Terraform CLI running times for the last 5m:
image

The medians (0.5-quantiles) for these observations aggregated by the mode and Terraform subcommand being invoked:
image

- `upjet_resource_ttr` histogram metric, showing average resource TTR for the last 10m:
image

The median (0.5-quantile) for these TTR observations:
image

### DIFF
--- a/pkg/metrics/metrics.go
+++ b/pkg/metrics/metrics.go
@@ -1,0 +1,90 @@
+// Copyright 2023 Upbound Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package metrics
+
+import (
+	"github.com/prometheus/client_golang/prometheus"
+	"sigs.k8s.io/controller-runtime/pkg/metrics"
+)
+
+const (
+	promNSUpjet     = "upjet"
+	promSysTF       = "terraform"
+	promSysResource = "resource"
+)
+
+// ExecMode is the Terraform CLI execution mode label
+type ExecMode int
+
+const (
+	// ModeSync represents the synchronous execution mode
+	ModeSync ExecMode = iota
+	// ModeASync represents the asynchronous execution mode
+	ModeASync
+)
+
+// String converts an execMode to string
+func (em ExecMode) String() string {
+	switch em {
+	case ModeSync:
+		return "sync"
+	case ModeASync:
+		return "async"
+	default:
+		return "unknown"
+	}
+}
+
+var (
+	// CLITime is the Terraform CLI execution times histogram.
+	CLITime = prometheus.NewHistogramVec(prometheus.HistogramOpts{
+		Namespace: promNSUpjet,
+		Subsystem: promSysTF,
+		Name:      "cli_duration",
+		Help:      "Measures in seconds how long it takes a Terraform CLI invocation to complete",
+		Buckets:   []float64{1.0, 3, 5, 10, 15, 30, 60, 120, 300},
+	}, []string{"subcommand", "mode"})
+
+	// CLIExecutions are the active number of terraform CLI invocations.
+	CLIExecutions = prometheus.NewGaugeVec(prometheus.GaugeOpts{
+		Namespace: promNSUpjet,
+		Subsystem: promSysTF,
+		Name:      "active_cli_invocations",
+		Help:      "The number of active (running) Terraform CLI invocations",
+	}, []string{"subcommand", "mode"})
+
+	// TFProcesses are the active number of
+	// terraform CLI & Terraform provider processes running.
+	TFProcesses = prometheus.NewGaugeVec(prometheus.GaugeOpts{
+		Namespace: promNSUpjet,
+		Subsystem: promSysTF,
+		Name:      "running_processes",
+		Help:      "The number of running Terraform CLI and Terraform provider processes",
+	}, []string{"type"})
+
+	// TTRMeasurements are the time-to-readiness measurements for
+	// the managed resources.
+	TTRMeasurements = prometheus.NewHistogramVec(prometheus.HistogramOpts{
+		Namespace: promNSUpjet,
+		Subsystem: promSysResource,
+		Name:      "ttr",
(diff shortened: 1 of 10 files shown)


ulucinar:
I've removed the `workspace` label from the metrics.

sergenyalcin (approved):
Thanks @ulucinar LGTM!

AaronME:
Thanks for looping me in, @Piotr1215 !

@ulucinar This looks good to me! Thank you for tackling this.

AaronME (approved):

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,90 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package metrics
+
+import (
+	"github.com/prometheus/client_golang/prometheus"
```
negz, line 18: Consider this non-blocking, but I'd prefer to use OpenTelemetry to expose Prom metrics. I believe we mostly use Otel for Upbound things internally, and it would open a path to use one SDK for all observability (i.e. traces and logs too).

We've held off on this in the past waiting to see what controller-runtime would do per https://github.com/kubernetes-sigs/controller-runtime/issues/305.

ulucinar: Thanks @negz for the pointer. Makes sense to me.
Let's proceed with the Prometheus metrics for now as the controller-runtime still makes use of them. I have not checked if it's possible with OpenTelemetry metrics but it was convenient to register upjet's custom metrics with the controller-runtime's registry.

PR #153, opened 2023-01-26
Add Patch migration step type

ulucinar:
### Description of your changes

This PR proposes to add the `Patch` step type to the migration plan and converts the following steps to this new type:
- pause-managed
- pause-composites
- edit-composites
- edit-claims
- deletion-policy-orphan
- start-managed
- start-composites

As an example, the generated manifest for pausing a managed resource in the `pause-managed` migration plan step now looks like the following:
```yaml
apiVersion: fakesourceapi/v1alpha1
kind: VPC
metadata:
  annotations:
    crossplane.io/paused: "true"
  name: sample-vpc
```

I have:

- [x] Read and followed Crossplane's [contribution process].
- [x] Run `make reviewable` to ensure this PR is ready for review.
- [ ] Added `backport release-x.y` labels to auto-backport this PR if necessary.

### How has this code been tested

Tested manually using a sample migrator.

[contribution process]: https://git.io/fj2m9

PR #173, opened 2023-03-13
Add migration.PlanGeneratorOption to configure a PlanGenerator

ulucinar:

Depends on: #153 

This PR introduces the `migration.PlanGeneratorOption` to configure a `PlanGenerator` in the migrators. Currently, the available configuration options are:
- `migration.WithErrorOnInvalidPatchSchema`: The migrators will now, by default, not error if issues are encountered while doing a patch schema check against the migration sources & targets and will exclude the patches in the migration target `ComposedTemplate`s after logging the issues. A migrator can revert back to the previous behavior by setting this option on the `PlanGenerator`.
- `migration.WithSkipGVKs`: Allows the migrators to configure a list of GVKs that will not be converted during a migration. Wildcard GVKs can be specified by either leaving the API group, version or kind empty. For example, if the kind of a GVK in the skip list is empty, then all kinds in the specified GV will be skipped. List of skipped GVKs override any converters registered with the `PlanGenerator`'s registry, i.e., even if there's a converter registered for a GVK and the GVK is also in the skip list, then the converter will *not* be run. Resources associated with skipped GVKs are included in the generated migration resources with no further checks or filtering.

This PR also fixes an issue observed when a resource type containing a nested `runtime.RawExtension` field is being migrated. Migration now continues without any further checks on `runtime.RawExtension`s. Currently, the framework just traverses the type hierarchy and being able to follow the structure specified by the `RawExtension` would require us to take the actual values into consideration so initially we just stop checking and filtering at the level a `RawExtension` is encountered.

I have:

Manually validated against a migrator converting provider-kubernetes `Object`s. `Object`s have a `runtime.RawExtension` at path `spec.forProvider.manifest`.

### DIFF
--- a/pkg/migration/patches.go
+++ b/pkg/migration/patches.go
@@ -16,6 +16,7 @@ package migration
 
 import (
 	"fmt"
+	"log"
 	"reflect"
 	"regexp"
 	"strings"
@@ -51,7 +52,8 @@ func isConverted(convertedPS []string, psName string) bool {
 // templates inherit patches from migration source templates by default, and
 // this function is responsible for removing patches (including references to
 // patch sets) that do not conform to the target composed template's schema.
-func removeInvalidPatches(c runtime.ObjectCreater, gvkSource, gvkTarget schema.GroupVersionKind, patchSets []xpv1.PatchSet, targetTemplate *xpv1.ComposedTemplate, convertedPS []string) error { //nolint:gocyclo // complexity (11) just above the threshold (10)
+func (pg *PlanGenerator) removeInvalidPatches(gvkSource, gvkTarget schema.GroupVersionKind, patchSets []xpv1.PatchSet, targetTemplate *xpv1.ComposedTemplate, convertedPS []string) error { //nolint:gocyclo // complexity (11) just above the threshold (10)
+	c := pg.registry.scheme
 	source, err := c.New(gvkSource)
 	if err != nil {
 		return errors.Wrapf(err, "failed to instantiate a new source object with GVK: %s", gvkSource.String())
@@ -88,7 +90,15 @@ func removeInvalidPatches(c runtime.ObjectCreater, gvkSource, gvkTarget schema.G
 		for _, p := range patches {
 			ok, err := assertPatchSchemaConformance(p, s, target)
 			if err != nil {
-				return errors.Wrap(err, "failed to check whether the patch conforms to the target schema")
+				err := errors.Wrap(err, "failed to check whether the patch conforms to the target schema")
+				if pg.ErrorOnInvalidPatchSchema {
+					return err
+				}
+				log.Printf("Excluding the patch from the migration target because conformance checking has failed with: %v\n", err)
+				// if we could not check the patch's schema conformance
+				// and the plan generator is configured not to error,
+				// assume the patch does not conform to the schema
+				ok = false
 			}
 			if !ok {
 				keep = false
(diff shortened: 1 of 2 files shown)


ytsarev (commented):
* Checked that with `migration.WithErrorOnInvalidPatchSchema(false)`  the migrator under test is reporting errors without the blocking failure:
```
2023/04/12 22:50:33 Excluding the patch from the migration target because conformance checking has failed with: failed to check whether the patch conforms to the target schema: failed to assert patch schema for path: spec.forProvider.tags.uid: failed to assert source struct field kind at path: uid: type is not a struct:
2023/04/12 22:50:33 Excluding the patch from the migration target because conformance checking has failed with: failed to check whether the patch conforms to the target schema: failed to assert patch schema for path: spec.forProvider.tags.deployment_guid: failed to assert source struct field kind at path: deployment_guid: type is not a struct:
2023/04/12 22:50:33 Excluding the patch from the migration target because conformance checking has failed with: failed to check whether the patch conforms to the target schema: failed to assert patch schema for path: spec.forProvider.tags.app: failed to assert source struct field kind at path: app: type is not a struct:
 echo $?
0

```

* Tried to skip specific GVK but unfortunately failed
Skip config:
```
+       pg := migration.NewPlanGenerator(converters.Registry, source, target, migration.WithErrorOnInvalidPatchSchema(false), migration.WithSkipGVKs(schema.GroupVersionKind{
+               Group:   "ec2.aws.crossplane.io",
+               Kind:    "SecurityGroup",
+               Version: "v1beta1",
```
The migrated Composition still has 

```
 - base:
      apiVersion: ec2.aws.upbound.io/v1beta1
      kind: SecurityGroup
```

I added verbosity to associated function
```
if pg.isGVKSkipped(gvkSource) {
+               fmt.Println("skipping", gvkSource)
                return nil
        }
```

and pretty sure that it was treated as skipped
```
go run cmd/migrator/main.go
skipping ec2.aws.crossplane.io/v1beta1, Kind=SecurityGroup
skipping ec2.aws.crossplane.io/v1beta1, Kind=SecurityGroup
skipping ec2.aws.crossplane.io/v1beta1, Kind=SecurityGroup
...
```
but the migrated SecurityGroup is still the in the resulting Composition

ytsarev:
My expectation is that the target Composition would have an intact source 'ec2.aws.crossplane.io' in this specific example.

More aggressive example is when source Composition has some special resource base, e.g. from some private provider

Consider this to be a part of the source Composition
```
      - name: customWorkers
        base:
⚠         apiVersion: custom.jet.crossplane.io/v1alpha1
          kind: Aws
```
The migration will fail with
```
main: error: Failed to generate the migration plan: failed to generate the migration plan: failed to migrate the composition: failed to migrate the composed templates of the composition: failed to set the defaults on the migration target composed template: failed to instantiate a new source object with GVK: custom.jet.crossplane.io/v1alpha1, Kind=Aws: no kind "Aws" is registered for version "custom.jet.crossplane.io/v1alpha1" in scheme "pkg/runtime/scheme.go:100"
```
even when we have current skip GVK config of
```
schema.GroupVersionKind{Group: "jet.crossplane.io"}))
```

So full conversion skip but *keeping* the original source config in the target Composition would make sense here.

ulucinar:
> The migration will fail with

> main: error: Failed to generate the migration plan: failed to generate the migration plan: failed to migrate the composition: failed to migrate the composed templates of the composition: failed to set the defaults on the migration target composed template: failed to instantiate a new source object with GVK: custom.jet.crossplane.io/v1alpha1, Kind=Aws: no kind "Aws" is registered for version "custom.jet.crossplane.io/v1alpha1" in scheme "pkg/runtime/scheme.go:100"
> even when we have current skip GVK config of

> schema.GroupVersionKind{Group: "jet.crossplane.io"}))

I think why the skip configuration did not behave as expected is because the API group of the `ComposedTemplate` is `custom.jet.crossplane.io`, whereas the skip configuration specifies `jet.crossplane.io`. If any of the components of a GVK is specified, we need an exact match on the specified components for skipping. A postfix won't work. 

We can also employ regular expressions here so that the specification language we use can be more powerful.

ytsarev:
@ulucinar thanks for the clarification!

The switch to exact group specification `schema.GroupVersionKind{Group: "custom.jet.crossplane.io"}))` helped to configure migrator to skip/ignore the gvk. But in this case, I would love to see it in its original state in the target Composition but it is not there. 

In other words, I would expect skipping migration to mean 'transfer the GVK base as it is to target Composition' instead of 'omit the GVK fully'

ytsarev (approved):
We synced up with @ulucinar  and I just double checked the skipping mechanism on my side - it actually transfers the unchanged resource base to the target composition.

So everything behaves as expected, thanks for the amazing job, Alper!

sergenyalcin (approved):
Thanks @ulucinar LGTM!

PR #178, opened 2023-03-22
Add terraform.ProviderScheduler

ulucinar:

Related to: https://github.com/upbound/provider-aws/issues/325

This PR adds `terraform.ProviderScheduler` interface and three implementations of it (`terraform.NoOpProviderScheduler`, a noop implementation, `terraform.SharedProviderScheduler` that shares Terraform provider processes among multiple reconciliation loops with a configured TTL, and a `terraform.WorkspaceProviderScheduler` that shares a Terraform provider process between the CLI invocations done in the context of a single reconciliation loop). Providers may opt to reenable the shared gRPC server runtime based on these schedulers, which properly isolate the forked Terraform providers to prevent some external resource leakage issues we had observed in the past with the shared server runtime. We have also performed a set of external resource leakage tests with this runtime, which will be discussed below.

I have:

We are still evaluating the performance characteristics of this runtime. We are also collecting feedback from the community members who gave the provider packages consuming this runtime a try. Relevant issues are here:
- https://github.com/upbound/provider-aws/issues/576
- https://github.com/upbound/provider-azure/issues/404
- https://github.com/upbound/provider-gcp/issues/255

We have also performed two long-running tests (one 2d 16h long, the other 4d 16h long) with a modified https://github.com/upbound/platform-ref-aws Crossplane configuration package (modified to depend on a `upbound/provider-aws` package that consumes this runtime). No leaks were observed. We also could not observe external resource leaks during an experiment with this runtime with 210 `cognitoidp.UserPool` resources spanning over 7 AWS regions, with 30 MRs per region. In this experiment, we also configured the poll interval to 1m to increase the likelihood of a race condition. This experiment lasted for 19h. Through another set of experiments involving 2 AWS regions conducted at the Terraform layer (not involving a Crossplane provider or upjet or the schedulers being discussed here) to stress test the isolation principle we've implemented. After 12h, no external resource leaks were observed.


pkg/terraform/store.go
```diff
@@ -57,6 +55,48 @@ type ProviderRequirement struct {
 // ProviderConfiguration holds the setup configuration body
 type ProviderConfiguration map[string]any
 
+// ToProviderHandle converts a provider configuration to a handle
+// for the provider scheduler.
+func (pc ProviderConfiguration) ToProviderHandle() (ProviderHandle, error) {
```
sergenyalcin, line 60: I think we are using the ProviderConfig hash, for deciding whether we need a new tf-provider process to, right? By this way, we will resolve the resource leak issues that we observed before.

sergenyalcin (commented):
Thanks @ulucinar I left a few comments to be sure that I understand the some parts of PR.

sergenyalcin:
We have a ttl value as 100. And also the ttlBudget is 0.1. As I understand, this means that, the native provider plugin process will be restarted after every 110 usage. For clusters with a low number of MRs, this means that the process will survive for a long time. Considering the memory leak problem on the Terraform side, how would it be to set a timeout for the relevant process? We can cover this in other iterations as well.

sergenyalcin:
What about using `ttlMargin` instead of `ttlBudget` to avoid ambiguity.

pkg/terraform/workspace.go
```diff
@@ -97,27 +134,51 @@ type CallbackFn func(error, context.Context) error
...
+}
+
 // ApplyAsync makes a terraform apply call without blocking and calls the given
 // function once that apply call finishes.
 func (w *Workspace) ApplyAsync(callback CallbackFn) error {
 	if !w.LastOperation.MarkStart("apply") {
 		return errors.Errorf("%s operation that started at %s is still running", w.LastOperation.Type, w.LastOperation.StartTime().String())
 	}
 	ctx, cancel := context.WithDeadline(context.TODO(), w.LastOperation.StartTime().Add(defaultAsyncTimeout))
+	w.providerInUse.Increment()
```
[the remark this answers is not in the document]
ulucinar: In the async execution cases, we increment the in-use counter via the reconciler goroutine to make sure that the reservations are actually performed by the reconciler goroutine. The `Workspace.runTF` might be executed by the reconciler goroutine (either if the execution mode for an MR is sync or if we always execute a Terraform CLI always in sync mode, e.g., `terraform apply -refresh-only` invocations), or it might be executed by a worker goroutine asynchronously. In these async cases, we would like to make sure that the goroutine who drives the scheduling decisions (i.e., the goroutine which drives the scheduler implementation) is actually the goroutine responsible for making the reservations (incrementing the in-use count makes a reservation on the shared plugin process, the scheduler is not allowed to replace it as long as it's actively in use).

pkg/terraform/store.go
ulucinar: Yes, correct. `ProviderHandle` is basically a hash of the Terraform provider configuration block. Clients pass a `ProviderHandle` to the scheduler while making reservation requests to identify their requests.

sergenyalcin:
Some test results that are done via this PR:

https://github.com/upbound/provider-aws/issues/576#issuecomment-1485259533
https://github.com/upbound/provider-aws/issues/576#issuecomment-1474137489
https://github.com/upbound/provider-aws/issues/576#issuecomment-1482747981

ulucinar:
> nit: What about using `ttlMargin` instead of `ttlBudget` to avoid ambiguity? We may also consider changing this in the next iteration.

Done.

sergenyalcin (approved):
Thanks @ulucinar LGTM! This is an important milestone in the context of addressing the performance issues.

przysiadZeSztanga:
Testing that with 500MR and settings like (--max-reconcile-rate=5)
CPU usage is significantly lower than on provider version 0.31 and 0.30 but I see tons of errors like:

`observe failed: cannot schedule a native provider during observe: xxxxxxxxxxxxxxx:
      cannot schedule native Terraform provider process: native provider reuse budget
      has been exceeded: invocationCount: 113, ttl: 100'`
      
  Also max rate seems not be affecting process, I see lots of terraform apply running in parallel:
  
  `18948     1 2000     S     817m   5%   7   0% terraform apply -auto-approve -input=false -lock=false -json
19247     1 2000     S     817m   5%   6   0% terraform apply -auto-approve -input=false -lock=false -json
18718     1 2000     S     817m   5%   1   0% terraform apply -auto-approve -input=false -lock=false -json
18757     1 2000     S     817m   5%   5   0% terraform apply -auto-approve -input=false -lock=false -json
19409     1 2000     S     817m   5%   2   0% terraform apply -auto-approve -input=false -lock=false -json
19001     1 2000     S     817m   5%   4   0% terraform apply -auto-approve -input=false -lock=false -json
19076     1 2000     S     884m   6%   2   0% terraform apply -auto-approve -input=false -lock=false -json
18887     1 2000     S     818m   5%   7   0% terraform apply -auto-approve -input=false -lock=false -json
18926     1 2000     S     817m   5%   0   0% terraform apply -auto-approve -input=false -lock=false -json
18797     1 2000     S     817m   5%   2   0% terraform apply -auto-approve -input=false -lock=false -json
19300     1 2000     S     817m   5%   6   0% terraform apply -auto-approve -input=false -lock=false -json
19655     1 2000     S     817m   5%   4   0% terraform apply -auto-approve -input=false -lock=false -json
19105     1 2000     S     817m   5%   1   0% terraform apply -auto-approve -input=false -lock=false -json
19699     1 2000     S     817m   5%   4   0% terraform apply -auto-approve -input=false -lock=false -json
18978     1 2000     S     817m   5%   2   0% terraform apply -auto-approve -input=false -lock=false -json
19210     1 2000     S     817m   5%   6   0% terraform apply -auto-approve -input=false -lock=false -json
19592     1 2000     S     817m   5%   6   0% terraform apply -auto-approve -input=false -lock=false -json
19170     1 2000     S     817m   5%   1   0% terraform apply -auto-approve -input=false -lock=false -json
19552     1 2000     S     817m   5%   1   0% terraform apply -auto-approve -input=false -lock=false -json
19759     1 2000     S     817m   5%   1   0% terraform apply -auto-approve -input=false -lock=false -json
19726     1 2000     S     817m   5%   4   0% terraform apply -auto-approve -input=false -lock=false -json
18836     1 2000     S     817m   5%   1   0% terraform apply -auto-approve -input=false -lock=false -json
19683     1 2000     S     817m   5%   0   0% terraform apply -auto-approve -input=false -lock=false -json
19044     1 2000     S     817m   5%   4   0% terraform apply -auto-approve -input=false -lock=false -json
18689     1 2000     S     817m   5%   6   0% terraform apply -auto-approve -input=false -lock=false -json
19103     1 2000     S     817m   5%   5   0% terraform apply -auto-approve -input=false -lock=false -json
18628     1 2000     S     817m   5%   2   0% terraform apply -auto-approve -input=false -lock=false -json`

PR #179, opened 2023-03-24
Consume DesiredStateChanged predicate

turkenh:

This PR:
- Bumps the crossplane runtime to the latest with DesiredStateChanged predicate: https://github.com/crossplane/crossplane-runtime/pull/372
- Pass it as an event filter in the generated controllers.
- Uses annotations to trigger a reconciliation after an async operation is completed.

I have:

- [ ] Read and followed Crossplane's [contribution process].

Create an s3 bucket and try deleting it when empty and non-empty.


turkenh:
During implementation, I realized that the current async operation callback mechanism does not backoff with failures and this PR will only keep it working as before instead of fixing it, opened a separate issue for that: https://github.com/upbound/upjet/issues/180

ulucinar:
We had sync'ed with @turkenh about this PR and had agreed to consume the predicates as event filters in https://github.com/upbound/upjet/pull/231 on top of the reconcile request requeue mechanism we had introduced. Closing this PR...

PR #195, opened 2023-05-01
Deprecate config.BasePackages.Controller in favor of ControllerMap

ulucinar:

`config.BasePackages.Controller` does not convey which subpackage a controller should be added to. One option would be to rely on a path component of the controller (e.g., if the controller being registered is `internal/controller/eks/clusterauth`, then the subpackage could be `eks` as deduced from the path) but the proposed `ControllerMap` makes this more explicit and makes it easier to override/remap if needed. 

The proposed ControllerMap is a map from the controller's path to the subpackage name for the controller. `Controller` array is also deprecated in favor of `ControllerMap`.

Note: The deprecated `BasePackages.Controller` API still exists and behaves exactly as is (if not, it will be treated as a bug), and the new `BasePackages.ControllerMap` API is an optional one. In summary, this PR is intended to be backwards-compatible.

I have:

Tested via https://github.com/upbound/provider-aws/pull/680.

### DIFF
--- a/pkg/config/common.go
+++ b/pkg/config/common.go
@@ -13,6 +13,19 @@ import (
 	tjname "github.com/upbound/upjet/pkg/types/name"
 )
 
+const (
+	// PackageNameConfig is the name of the provider subpackage that contains
+	// the base resources (e.g., ProviderConfig, ProviderConfigUsage,
+	// StoreConfig. etc.).
+	// TODO: we should be careful that there may also exist short groups with
+	// these names. We can consider making these configurable by the provider
+	// maintainer.
+	PackageNameConfig = "config"
+	// PackageNameMonolith is the name of the backwards-compatible
+	// provider subpackage that contains the all the resources.
+	PackageNameMonolith = "monolith"
+)
+
 // Commonly used resource configurations.
 var (
 	DefaultBasePackages = BasePackages{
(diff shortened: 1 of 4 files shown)


pkg/config/common.go
```diff
@@ -21,9 +34,9 @@ var (
 			"apis/v1alpha1",
 			"apis/v1beta1",
 		},
-		Controller: []string{
```
turkenh, line 24: Should we also keep the default value of `Controller` for backward compatibility?

pkg/config/provider.go
```diff
@@ -44,7 +44,9 @@ func (cc ResourceConfiguratorChain) Configure(r *Resource) {
 // and controllers. Typically, we expect to see ProviderConfig packages here.
 type BasePackages struct {
```
turkenh, line 45: Is this `BasePackages` the same as the config package we configured with `PackageNameConfig`? 
Could we name that variable as `PackageNameBase="config"` or at least have a comment indicating this?

turkenh (approved):

pkg/config/provider.go
ulucinar: Done. Renamed the `const` from `PackageNameConfig` to `PackageNameFamily`.

PR #212, opened 2023-06-07
Add support for category-based manifest pre-processors

ulucinar:

This PR adds support for category-based manifest pre-processors which can optionally be registered with a migration registry and are called before any converters are called. They cannot mutate resources (this is the job of converters) and they can only be used to extract information from the manifests belonging to a specific category such as the managed resources or compositions. Pre-processors can be used to initialize converters with information extracted from the manifests emitted by a source.

Following changes are also introduced in this PR:
- Support for `Lock.pkg` converters
- Use [`meta.RESTMapper`](https://pkg.go.dev/k8s.io/apimachinery/pkg/api/meta#RESTMapper) for converting from GVKs to GVRs in `migration.KubernetesSource` and remove all references of `meta.UnsafeGuessKindToResource`.
- Fix a panic when no converters act on an observed `Configuration.meta`

I have:

### DIFF
--- a/pkg/migration/configurationmetadata_steps.go
+++ b/pkg/migration/configurationmetadata_steps.go
@@ -42,7 +42,7 @@ const (
 	errConfigurationMetadataOutput = "failed to output configuration JSON merge document"
 )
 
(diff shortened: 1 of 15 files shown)


sergenyalcin (approved):
Thanks @ulucinar LGTM!

PR #217, opened 2023-06-14
Add support for categorical converters

ulucinar:

This PR adds support for the categorical converters, which allows converters to work on the whole category of resources, such as the `managed` resources. The migration registry allows registration of category converters via the `Registry.RegisterCategoricalConverter` function and categorical converters receive and act on unstructured objects.

It also introduces support for categorical preprocessors: A registered (with the migration registry) preprocessor is guaranteed to be run before any registered converters are run. Preprocessors can currently be registered for the [CRD categories](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#categories) and they can be used to alter (to supply parameters to) the behavior of any converters run downstream.

We also introduce support for converters of various types such as Crossplane's `Configuration.pkg`, `Configuration.meta`, `Provider.pkg`, `Lock.pkg`, etc., which will be employed in various migrators targeting conversions from (community or official) monoliths to provider families.

We also add the `migration.PlanExecutor` supporting `migration.ExecutorCallback`. A `PlanExecutor` is responsible for driving the execution of the `Step`s of a migration `Plan` using its configured `Executor`. The repo currently hosts a single `Executor` implementation: `migration.ForkExecutor`, which can execute migration plan steps by forking child processes. `ExecutorCallback` implementations can be used to be notified:
- Just before a migration `Step` is executed. This allows a callback implementation to skip or continue with the execution of the `Step` or may be used to cancel the execution of the `Plan`.
- After a successful execution of a migration `Step`. This allows an implementation to be notified of a successful step execution and the execution of the `Plan` can be cancelled.
- After a `Step` execution failure:  This allows an implementation to be notified of a failed step execution and the `Step` execution may be repeated (for instance after editing any resource manifests or patch documents the `Step` refers to), skipped, or the `Plan` execution can be cancelled. 

I have:

The PR has been tested using https://github.com/upbound/extensions-migration/pull/12. 


sergenyalcin (approved):
Thanks @ulucinar LGTM!

PR #224, opened 2023-06-29
Support granular managment policies

### LINKED ISSUE #220: Support for granular management policies
As described in [granular management policies design](https://github.com/crossplane/crossplane/blob/952b124c6d8b75d827df84b77b6783c8c3f88ee0/design/one-pager-ignore-changes.md), we are changing the implemtation of the ObserveOnly alpha feature by introducing granular Management polices. 

For Upjet, we would need to update:
- [Observe logic](https://github.com/upbound/upjet/blob/018e6404baef55366ddd86d8d9cbf04c9bcb2426/pkg/controller/external.go#L164)
- [CEL generations rule logic](https://github.com/upbound/upjet/blob/018e6404baef55366ddd86d8d9cbf04c9bcb2426/pkg/types/builder.go#L136)

lsviben:

Updated the ObserveOnly  management policy support to the [new design](https://github.com/crossplane/crossplane/blob/312c84a9c63a0cb90fc4cbb709db6e553745dfdb/design/one-pager-ignore-changes.md) which 
incorporates granular management policies.

The main change is that instead of enum based `spec.managementPolicy: FullControl/ObserveOnly/OrphanOnDelete`, now we have an more granular array of actions `spec.managementPolicy:["Observe", "Create", "Update", "LateInitialize", "Delete"]`.

So as upjet was already supporting ObserveOnly with Import, we needed to do minor changes on conditionals to make sure that `spec.managementPolicy:["Observe"]` is still recognized as ObserveOnly. In fact, all combinations which don't contain Create or Update actions can be handled similarly.

Furthermore, as there is now a management policy which skips late initialziation, I noticed that upjet was using late initialization for updating critical annotations on resources. Those annotations are usually set after Create, but as upjet offers async creation, and most resources use that (all in aws-provider for example), they are not set after Create, but in Observe, with using late-init to update the resource. 

In this case, when late-init is off, we update the annotations manually in the Observe step. We could consider doing this in all situations and stop relying on late-init to do it for us. Or handle this differently.

Fixes #220 

I have:

Tested it in the same process as https://github.com/crossplane/crossplane-runtime/pull/456 as they are connected. More info on that issue.


go.mod
```diff
@@ -2,6 +2,9 @@ module github.com/upbound/upjet
 
 go 1.19
```
jeanduplessis, line 3: Since this is a substantial update to Upjet I wonder if if now is a good time to go up to go 1.20 as well

go.mod
```diff
@@ -17,21 +20,21 @@ require (
...
-	k8s.io/client-go v0.26.3
-	k8s.io/utils v0.0.0-20221128185143-99ec85e7a448
-	sigs.k8s.io/controller-runtime v0.14.6
-	sigs.k8s.io/json v0.0.0-20220713155537-f223a00ba0e2
+	k8s.io/api v0.27.2
+	k8s.io/apimachinery v0.27.2
+	k8s.io/client-go v0.27.2
+	k8s.io/utils v0.0.0-20230505201702-9f6742963106
+	sigs.k8s.io/controller-runtime v0.15.0
+	sigs.k8s.io/json v0.0.0-20221116044647-bc3834ca7abd
```
jeanduplessis, line 37: Are all of these dependency updates needed for the PR? 
I only ask as I saw the Renovate PR to bump these caused our pipelines to all fail https://github.com/upbound/provider-aws/pull/759

pkg/controller/external.go
```diff
@@ -157,11 +160,15 @@ func (e *external) Observe(ctx context.Context, mg xpresource.Managed) (managed.
 
 	// Note(turkenh): We don't need to check if the management policies are
 	// enabled or not because the crossplane-runtime's managed reconciler already
-	// does that for us. In other words, if the management policy is set to
-	// "ObserveOnly" without management policies being enabled, the managed
+	// does that for us. In other words, if the management policies are set to
```
jeanduplessis, line 163:
```suggestion
	// does that for us. In other words, if the management policies are set
```

go.mod
lsviben: it is because the controller-runtime was updated in xp-runtime and that pulls all of the other stuff. Ill make a separate PR for that to make this PR leaner

go.mod
lsviben: Sure i can add that

go.mod
lsviben: actually, ill leave it for now. When https://github.com/crossplane/crossplane-runtime/pull/456 gets merged, we can do a xp-runtime release, so that here we get the dependency on a released version, instead of a commit

lsviben: > Are all of these dependency updates needed for the PR? I only ask as I saw the Renovate PR to bump these caused our pipelines to all fail [upbound/provider-aws#759](https://github.com/upbound/provider-aws/pull/759)

Now i checked, when testing i had to update one line [here](https://github.com/upbound/provider-aws/blob/1ef9fd0c17313478f30b9bcd3d48701e0b1ec9f7/internal/controller/providerconfig/config.go#L33) for `provider-aws` . Probably its failing because of that:

```
-               Watches(&source.Kind{Type: &v1beta1.ProviderConfigUsage{}}, &resource.EnqueueRequestForProviderConfig{}).
+               Watches(&v1beta1.ProviderConfigUsage{}, &resource.EnqueueRequestForProviderConfig{}).
```

pkg/controller/external.go
```diff
@@ -155,13 +159,19 @@ func (e *external) Observe(ctx context.Context, mg xpresource.Managed) (managed.
...
+	// does that for us. In other words, if the management policies are set
+	// without management policies being enabled, the managed
 	// reconciler will error out before reaching this point.
 	// https://github.com/crossplane/crossplane-runtime/pull/384/files#diff-97300a2543f95f5a2ada3560bf47dd7334e237e27976574d15d1cddef2e66c01R696
-	if tr.GetManagementPolicy() == xpv1.ManagementObserveOnly {
+	// Note (lsviben) We are only using import instead of refresh if the
+	// management policies do not contain create or update as they need the
+	// required fields to be set, which is not the case for import.
+	if tr.GetManagementPolicies() != nil &&
+		!policySet.HasAny(xpv1.ManagementActionCreate, xpv1.ManagementActionUpdate, xpv1.ManagementActionAll) {
```
turkenh, line 174: Wondering if we could use the newly introduced `ManagementPoliciesResolver` in runtime here?

lsviben: Well, we would need to add the enabled flag and deletion policy into it. For these simple cases it seemed to me like its not needed

go.mod
turkenh: In general, I would prefer doing such a version upgrade in a separate/dedicated PR not to mix things up.

pkg/controller/external.go
```diff
@@ -155,13 +159,19 @@ func (e *external) Observe(ctx context.Context, mg xpresource.Managed) (managed.
...
-	// "ObserveOnly" without management policies being enabled, the managed
+	// does that for us. In other words, if the management policies are set
+	// without management policies being enabled, the managed
 	// reconciler will error out before reaching this point.
 	// https://github.com/crossplane/crossplane-runtime/pull/384/files#diff-97300a2543f95f5a2ada3560bf47dd7334e237e27976574d15d1cddef2e66c01R696
-	if tr.GetManagementPolicy() == xpv1.ManagementObserveOnly {
+	// Note (lsviben) We are only using import instead of refresh if the
+	// management policies do not contain create or update as they need the
+	// required fields to be set, which is not the case for import.
+	if tr.GetManagementPolicies() != nil &&
```
turkenh, line 173: It looks like we can drop `tr.GetManagementPolicies() != nil` here?

pkg/controller/external.go
```diff
@@ -199,18 +209,38 @@ func (e *external) Observe(ctx context.Context, mg xpresource.Managed) (managed.
...
+	// to true so that the reconciler updates the managed resource. This
+	// behavior conflicts with management policies in which LateInitialize is
+	// turned off. To circumvent this, we are checking if the management policy
+	// does not contain LateInitialize and if it does not, we are updating the
+	// annotations manually.
 	annotationsUpdated, err := resource.SetCriticalAnnotations(tr, e.config, tfstate, string(res.State.GetPrivateRaw()))
 	if err != nil {
 		return managed.ExternalObservation{}, errors.Wrap(err, "cannot set critical annotations")
 	}
+	if annotationsUpdated && !policySet.HasAny(xpv1.ManagementActionLateInitialize, xpv1.ManagementActionAll) {
```
turkenh, line 227: nit: It could be more readable if we define a var like `policyHasLateInit` and use it here and [below](https://github.com/upbound/upjet/pull/224/files#diff-65e234e18bcdf00d056617ba597cbc9988035cbfe73bc28158a85b188222eb30R239)

```go
policyHasLateInit := policySet.HasAny(xpv1.ManagementActionLateInitialize, xpv1.ManagementActionAll)
if annotationsUpdated && !policyHasLateInit {
```

and 

```go
if policyHasLateInit { 
```

turkenh (approved):
Thanks @lsviben, looking good!

go.mod
lsviben: Makes sense to do it in a separate PR :+1:

pkg/controller/external.go
```diff
@@ -155,13 +159,18 @@ func (e *external) Observe(ctx context.Context, mg xpresource.Managed) (managed.
...
-	// "ObserveOnly" without management policies being enabled, the managed
+	// does that for us. In other words, if the management policies are set
+	// without management policies being enabled, the managed
 	// reconciler will error out before reaching this point.
 	// https://github.com/crossplane/crossplane-runtime/pull/384/files#diff-97300a2543f95f5a2ada3560bf47dd7334e237e27976574d15d1cddef2e66c01R696
-	if tr.GetManagementPolicy() == xpv1.ManagementObserveOnly {
+	// Note (lsviben) We are only using import instead of refresh if the
+	// management policies do not contain create or update as they need the
+	// required fields to be set, which is not the case for import.
+	if !policySet.HasAny(xpv1.ManagementActionCreate, xpv1.ManagementActionUpdate, xpv1.ManagementActionAll) {
```
ulucinar, line 173: Isn't this going to result in an import if, for example, the management policy is set only to `Delete`?  Do we rely on the `managed.defaultSupportedManagementPolicies` here (looks like the only supported default policies not containing these are the `Observe` and the empty policy)? I assume with the empty policy we don't expect to reach here and for the remaining `Observe`, import is the right thing to do.

But what will happen, if for instance, the provider configures the supported management policies to include some other set? Do we rely specifically on the default supported policies here?

pkg/controller/external.go
```diff
@@ -199,18 +208,39 @@ func (e *external) Observe(ctx context.Context, mg xpresource.Managed) (managed.
...
+	// behavior conflicts with management policies in which LateInitialize is
+	// turned off. To circumvent this, we are checking if the management policy
+	// does not contain LateInitialize and if it does not, we are updating the
+	// annotations manually.
 	annotationsUpdated, err := resource.SetCriticalAnnotations(tr, e.config, tfstate, string(res.State.GetPrivateRaw()))
 	if err != nil {
 		return managed.ExternalObservation{}, errors.Wrap(err, "cannot set critical annotations")
 	}
+	policyHasLateInit := policySet.HasAny(xpv1.ManagementActionLateInitialize, xpv1.ManagementActionAll)
+	if annotationsUpdated && !policyHasLateInit {
```
ulucinar, line 227: So we are explicitly doing an update here for the sole case when:
- Management policies are enabled
- The MR's management policy lacks the `LateInitialize` policy

In this case, because in the current implementation we rely on the managed reconciler doing an update when the `managed.ExternalObservation.ResourceLateInitialized` is set to `True` (and I assume in the new crossplane-runtime, if late-initialization is disallowed, we will skip an update even though we return `ResourceLateInitialized: True` here), we have to update the resource. 

Is this correct? And is the only case we are doing this for is the case described above?

pkg/controller/external.go
```diff
@@ -199,18 +208,39 @@ func (e *external) Observe(ctx context.Context, mg xpresource.Managed) (managed.
...
+		if err := e.kube.Update(ctx, mg); err != nil {
```
ulucinar, line 228: This may have performance implications because it may result in conflicts when/if the managed reconciler also decides to a spec update. 

I don't think this is super important because I assume the number of times `annotationsUpdated` will be `True` will probably be small but I'm 100% sure as it's not just the critical annotation keys but also the values. Just leaving a note here for future reference. What do you think?

pkg/types/builder.go
```diff
@@ -133,7 +133,7 @@ func (g *Builder) AddToBuilder(typeNames *TypeNames, r *resource) (*types.Named,
 
 	for _, p := range r.topLevelRequiredParams {
 		g.validationRules += "\n"
-		g.validationRules += fmt.Sprintf(`// +kubebuilder:validation:XValidation:rule="self.managementPolicy == 'ObserveOnly' || has(self.forProvider.%s)",message="%s is a required parameter"`, p, p)
+		g.validationRules += fmt.Sprintf(`// +kubebuilder:validation:XValidation:rule="!('*' in self.managementPolicies || 'Create' in self.managementPolicies || 'Update' in self.managementPolicies) || has(self.forProvider.%s)",message="%s is a required parameter"`, p, p)
```
ulucinar, line 136: Are we sure that for a `Delete` operation, we are fine with not setting a required parameter? Did we validate this? I think it's not possible to have delete-only resource right, at least with the default policy sets? Looks like `Delete` policy always comes with the `Create` policy, is this correct? 

If so, are we relying on the default supported policy sets?

ulucinar (commented):
Thank you @lsviben for the PR. Left some comments for my understanding. And thank you for bearing with me while I'm trying to understand these important changes better.

go.mod
jeanduplessis: Alper implemented the upgrade to 1.20 here: https://github.com/upbound/upjet/pull/236

lsviben: Awesome, thanks @ulucinar !

pkg/controller/external.go
lsviben: As in crossplane-runtime, we decided not to support & test initially all possible combinations. It's true that Provider developers can create their own set of supported management polices, but if something is not working for them out of the box, I would expect them to create a ticket to support some combination or contribute it.  

The options which are left when we remove Create and Update are (and not currently supported by default):
```
["Observe", "LateInitialize", "Delete"]
["Observe", "LateInitialize"]
["Observe", "Delete"]
["Delete", "LateInitialize"]
["Delete"]
["LateInitialize"]
```

Supported by default:
```
["Observe"]
[""]
```
From my perspective any without "Observe" does not make a lot of sense (or I dont see the use case) so what is left is :
```
["Observe", "LateInitialize", "Delete"]
["Observe", "LateInitialize"]
["Observe", "Delete"]
```

I dont see why `["Observe", "Delete"]` should not work. Only thing I see is that if we add `LateInitialize` that the late initialization wont happen. I can add the code for late initialization to Import if that is what we want. Not sure if we want to support this right away though.

pkg/controller/external.go
lsviben: Yeah, this is the case. 

Crossplane-runtime reconciler[ "expects"](https://github.com/crossplane/crossplane-runtime/blob/8e87b2c47ee705f88b1224bff7b3eb8d1a071760/pkg/reconciler/managed/reconciler.go#L1030) the critical annoatations to be updated after a Create. But, as Upjet mostly does [ApplyAsync](https://github.com/upbound/upjet/blob/accab08e93515620340fa2b3576fefbf14d70071/pkg/controller/external.go#L315), it doesnt set the Critical Annotations, instead it kinda misuses the LateInitialization mechanism to update the managed resource with the annotations. 

So yeah, this just makes sure that the annotations are still updated even if `LateInitialization` is off. So it should only affect resources in which:
- managementPolicies are enabled
- `LateInitialization` is off

pkg/controller/external.go
lsviben: Similar to the answer [above](https://github.com/upbound/upjet/pull/224#discussion_r1269912443). It should only affect resources in which:
- managementPolicies are enabled
- `LateInitialization` is off

But good point to keep it in mind :+1: for possible performance implications later on.

pkg/types/builder.go
lsviben: Aswered a bit [here](https://github.com/upbound/upjet/pull/224#discussion_r1269906549), but now I see that possibly delete wont work if we dont have required fields, not sure how Terraform reacts to that. As I mentioned I didn't test all unsupported combinations. All supported policies have `Delete` in a combination with either `Create` or `Update`.

pkg/controller/external.go
ulucinar: I think what we have implemented here is correct, given that what's left for here is the `{Observe}` and the empty set, `{}` (in which case we will just pause reconciliation I assume). What worries me is that we have an API to override the set of supported policies for an upjet-based provider, and this implementation makes assumptions on the set of supported policies. In principle, I think it would be safer to disallow (at least for now until we have a good use case) the API that allows changing the supported policies but I don't certainly think that it's a blocker for this PR.

pkg/controller/external.go
ulucinar: Yes, I believe why we initially relied on setting `LateInitialized: True` in the first place (as you mentioned above) is to prevent conflicts with the managed reconciler's updates by trying to have the managed reconciler as the sole entity doing these updates. Let's continue with this implementation and we will need to act if we observe any adverse affects.

pkg/types/builder.go
ulucinar: > ... not sure how Terraform reacts to that...
This is also my concern here. I can help validating the behavior. 

Another concern is the validation rules (build-time entities) are not aware of if management policies are actually enabled or not (the decision is only available at runtime). But with these "static" rules, we are conditioning whether, for instance, management policy for a resource is set to `Observe`.

Image a case where the management policy is set to `Observe` and the management policies feature is disabled at runtime. For now, we are relying on the crossplane-runtime to error when the management policies are not enabled but a non-default (non `*`) management policy is declared. I don't think this is a blocker for this PR, just trying to understand the implications so that we can maintain the new feature.

ulucinar (approved):
Thanks @lsviben for the detailed clarifications, lgtm.

PR #235, opened 2023-07-20
Add description field to the migration plan

sergenyalcin:

This PR adds description field to the migration plan for better user interface

I have:

Generated and reviewed migration plan


pkg/migration/testdata/plan/generated/configurationv1_pkg_migration_plan.yaml
```diff
@@ -224,5 +246,6 @@ spec:
       manualExecution:
         - "kubectl patch --type='merge' -f deletion-policy-delete/sample-vpc.vpcs.fakesourceapi_v1alpha1.yaml --patch-file deletion-policy-delete/sample-vpc.vpcs.fakesourceapi_v1alpha1.yaml"
       type: Patch
+      description: "Setting the deletion policies of Managed Resources whose deletion policy is set to Orphan at the beginning of the migration, to Delete again"
```
ulucinar, line 249:
```suggestion
      description: "Setting the deletion policies of Managed Resources whose deletion policy had been set to Orphan at the beginning of the migration process, back to Delete"
```

pkg/migration/plan_executor.go
```diff
@@ -69,6 +71,12 @@ func WithExecutorCallback(cb ExecutorCallback) PlanExecutorOption {
 	}
 }
 
+func WithStartIndex(startIndex int) PlanExecutorOption {
```
ulucinar, line 74: Let's also add documentation for this.

pkg/migration/plan_executor.go
```diff
@@ -103,14 +111,15 @@ func NewPlanExecutor(plan Plan, executors []Executor, opts ...PlanExecutorOption
 
 func (pe *PlanExecutor) Execute() error { //nolint:gocyclo // easier to follow this way
 	ctx := make(map[string]any)
-	for i := 0; i < len(pe.plan.Spec.Steps); i++ {
+	for i := pe.StartIndex; i < len(pe.plan.Spec.Steps); i++ {
```
ulucinar, line 114: Do we have access to a logger in `migration.PlanExecutor`? If so, we may consider adding at least a debug statement that will convey the step index we are starting execution at.

pkg/migration/plan_executor.go
```diff
@@ -103,14 +111,15 @@ func NewPlanExecutor(plan Plan, executors []Executor, opts ...PlanExecutorOption
...
 		var r CallbackResult
 		if pe.callback != nil {
 			r = pe.callback.StepToExecute(pe.plan.Spec.Steps[i], i)
 			switch r.Action {
 			case ActionCancel:
 				return nil
 			case ActionSkip:
+				pe.LastSuccessfulStep = i
```
ulucinar, line 122: How do we use `LastSucccessfulStep`? If a step is skipped, is it still a successful one?

pkg/migration/plan_executor.go
```diff
@@ -26,9 +26,11 @@ const (
 // PlanExecutor drives the execution of a plan's steps and
 // uses the configured `executors` to execute those steps.
 type PlanExecutor struct {
-	executors []Executor
-	plan      Plan
-	callback  ExecutorCallback
+	executors          []Executor
+	plan               Plan
+	callback           ExecutorCallback
+	LastSuccessfulStep int
```
ulucinar, line 32: Where do we read `LastSuccessfulStep`?

ulucinar (commented):
Thank you @sergenyalcin for introducing the step descriptions that will help folks give more context into what the migrator is doing and for what purpose. 

We may also consider having a verbose mode for the plan executor, in which we will dump the description and the outputs from the executed steps. If the executor is not verbose, then we may consider just dumping the description strings.

And as also mentioned in the review comments, we can also consider giving some more context in the description texts, such the the provider names being installed/waited, files being applied, image URLs being pushed/replaced, etc. I believe this also depends whether we will have a mode of execution where we just dump the descriptions (an option for a cleaner output in my opinion). Let's discuss.

pkg/migration/plan_executor.go
sergenyalcin: We read this value in the family-migrator. Please see this [PR](https://github.com/upbound/extensions-migration/pull/18).

pkg/migration/plan_executor.go
sergenyalcin: I left that up to the reviews to talk about. In family-migrator, we return the skip action for the steps that the user has run manually. We should consider successful steps that are run manually and that the user validates it.

However, the skip action is a much more general action. Therefore, we can specify another type of action for such situations. (For manually performed steps)

pkg/migration/plan_executor.go
sergenyalcin: We don't have access to a logger. One option can be adding a log to callback implementation.

pkg/migration/plan_executor.go
sergenyalcin: Or the another option can be that changing the name of the `LastSucccessfulStep` to `LastPerformedStep`.

This field is used for caching mechanism for plan. Please see this [PR](https://github.com/upbound/extensions-migration/pull/18)

sergenyalcin:
@ulucinar I made the changes you suggested about the wording in the descriptions. However, I would like to say a few things about the fact that it contains more information. 

It may be good to provide as much information as possible in the Description, but I think it is sufficient to set the summary information as a title in its current form. An example output is in this [PR](https://github.com/upbound/extensions-migration/pull/18). (Of course, this is an older version, the wording has been improved.)

PR #237, opened 2023-07-24
Support initProvider 

### LINKED ISSUE #225: initProvider support in upjet
As described in [granular management policies design](https://github.com/crossplane/crossplane/blob/952b124c6d8b75d827df84b77b6783c8c3f88ee0/design/one-pager-ignore-changes.md), we are adding a new field `spec.initProvider` which will complement `spec.forProvider` during the external Resource Create. This will allow users to specify some fields that they would like or need(required fields) to set for the resource creation, but that they dont want Crossplane to control and later keep updated. 

This behaviour also complements the granular managementPolicies, especially those which are skipping Late Initialziation.

Example usage for EKS NodeGroup:
```
  initProvider:
    scalingConfig:
      - desiredSize: 1
  forProvider:
    region: us-east-1
    scalingConfig:
      maxSize: 5
      minSize: 1
```

lsviben:

Adds support for the new `initProvider` alpha feature, which is a part of the management
policies feature ([design](https://github.com/crossplane/crossplane/blob/master/design/one-pager-ignore-changes.md))

#### Runtime changes

As Upjet uses Terraform, it needs to create the `main.tf` file and workspace to run commands against it. It does it in the [Connect step](https://github.com/crossplane/crossplane-runtime/blob/8e87b2c47ee705f88b1224bff7b3eb8d1a071760/pkg/reconciler/managed/reconciler.go#L839) of the managed reconciler. The creation of the workspace already fills up the main.tf with the resource in question. This means that in the [Create](https://github.com/crossplane/crossplane-runtime/blob/8e87b2c47ee705f88b1224bff7b3eb8d1a071760/pkg/reconciler/managed/reconciler.go#L983) and [Update](https://github.com/crossplane/crossplane-runtime/blob/8e87b2c47ee705f88b1224bff7b3eb8d1a071760/pkg/reconciler/managed/reconciler.go#L1096) steps the passed resource is not actually the one which we Create/Update, as it has already been set in the `Connect`. This means that at the point of creating the `main.tf` file, we dont know if it will be used for Update or Create. So to fulfill the purpose of `initProvider` which is to have fields that are only used during Create, we need to use available TF tools - ignore_changes lifecycle hook, to mark the fields which should only be used for creation.

So if the management policies are enabled, when creating the FileProducer that holds the parameters of the `main.tf` file:
-  we calculate the fields which should be in the `ignore_changes` by going through the `spec.initProvider` and `spec.forProvider` , and looking for fields set in `spec.initProvider` but not in `spec.forProvider`
- we merge the `spec.initProvider` to `spec.forProvider` 

TODO:
currently `spec.initProvider` fields overwrite the `spec.forProvider` fields if both are set (slices and maps merge), due to how mergo works. In practice, this will be a problem if the user tries to set a field/map key in both. For our current use cases we dont have this situation, and it can be resolved easily by removing the `initProvider` conflicting key if the user wants to use `forProvider` for it. But we should take a look and fix that.
Issue: https://github.com/upbound/upjet/issues/240

#### Code generation

As mentioned in the design, added a new field in the resource `spec` , `spec.initProvider`. It will be used to set some fields of the `spec.forProvider` which should be used just in Create. We are not copying `spec.forProvider` 1:1 because some field types are not suited to be replaced during Create. 

The code generation recognizes a few different field types, based on resource config or terraform API information:
- **Identifier** fields, which are used to identify resources in addition to names. Good example is AWS `region`. Configured by provider options
- **Reference** fields, which allow to resolve cross resource references into some fields. The resolving of references takes place in the beginning of the managed reconciler loop. Configured by provider options. Example is:

```go
	// ARN of the certificate authority.
	// +crossplane:generate:reference:type=CertificateAuthority
	CertificateAuthorityArn *string `json:"certificateAuthorityArn,omitempty" tf:"certificate_authority_arn,omitempty"`

	// Reference to a CertificateAuthority to populate certificateAuthorityArn.
	CertificateAuthorityArnRef *v1.Reference `json:"certificateAuthorityArnRef,omitempty" tf:"-"`
```
- **Sensitive** fields, are those marked in terraform to contain sensitive information, such as passwords. We are replacing them with secret references to avoid having sensitive information in resource yaml's. 
- **Regular** fields, are all other fields for the resource, generated through parsing the fields of the terraform resource schema.

For `initProvider` we are just interested in the **Regular** fields. Because:
- Identifiers specify the resource, we cannot create the resource in `region: us-west-1` and expect to later managed it in `us-east-3`.
- References because they are references to other resources. Not sure if there is an use case where a resource depends on one resource during creation, but during lifetime another. It can easily be changed in the forProvider if needed. 
- Sensitive fields similarly to above, as they are in the end also references to secrets. 
- fields which have a `tf: "-"` tag. Its the fields above, but added it for precaution, as we are merging initProvider and forProvider and calculating `ignore_changes` based on the terraform fields.

We could in theory add reference and sensitive fields to the `initProvider` but as there is not use case for that and their usage could incur bugs, I would rather leave them out unless they are needed. 

#### CEL rules

All required `spec.forProvider` fields, which are also `spec.initProvider` fields are now optional, because they can be replaced with `initProvider` fields. We already have a functionality to mark the top non-identifier required fields as optional and set CEL rules to check if they are set, or if ObserveOnly managementPolicy is [active](https://github.com/upbound/upjet/blob/main/pkg/types/builder.go#L134). This only affects the top level fields, as it can easily mark required nested structs. 
```yaml
// +kubebuilder:validation:XValidation:rule="!('*' in self.managementPolicies || 'Create' in self.managementPolicies || 'Update' in self.managementPolicies)",message="%s is a required parameter"
```

Now because of initProvider, we can also check if the field is in `spec.initProvider`

```yaml
// +kubebuilder:validation:XValidation:rule="!('*' in self.managementPolicies || 'Create' in self.managementPolicies || 'Update' in self.managementPolicies || has(self.forProvider.%s))",message="%s is a required parameter"
```

But because now we merge required nested structs from `spec.initProvider` to `spec.forProvider` we need to expand the CEL rules to also mark the nested fields as required either in `spec.initProvider` or `spec.forProvider`. This is something still TODO, issue: https://github.com/upbound/upjet/issues/239. 

Fixes #225 

I have:

The generation code has been tested with [generating provider aws](https://github.com/upbound/provider-aws/pull/785/commits/fe906d63156b015775b12c48b8ee4ce173f251e4)

Tested with provider-aws([PR](https://github.com/upbound/provider-aws/pull/785)), creating a Node Group from the [design example](https://github.com/crossplane/crossplane/blob/master/design/one-pager-ignore-changes.md#eks-nodegroup). 
with some added fields to test the ignore_changes generation:

```yaml
apiVersion: eks.aws.upbound.io/v1beta1
kind: NodeGroup
metadata:
  name: sample-eks-ng
spec:
  managementPolicies: ["Observe", "Create", "Update", "Delete"]
  initProvider:
    tags:
      someTag: tag
    forceUpdateVersion: false
    scalingConfig:
      - desiredSize: 1
  forProvider:
    clusterNameRef:
      name: sample-eks-cluster
    nodeRoleArnRef:
      name: sample-node-role
    region: us-west-1
    scalingConfig:
      - maxSize: 4
        minSize: 1
    subnetIdRefs:
      - name: sample-subnet1
      - name: sample-subnet2
```

Resulting `main.tf.json`:
```json
"resource": {
    "aws_eks_node_group": {
      "sample-eks-ng": {
        "cluster_name": "sample-eks-cluster",
        "force_update_version": false,
        "lifecycle": {
          "ignore_changes": [
            "force_update_version",
            "scaling_config[0].desired_size",
            "tags[\"someTag\"]"
          ],
          "prevent_destroy": true
        },
        "node_group_name": "sample-eks-ng",
        "node_role_arn": "arn:aws:iam::609897127049:role/sample-node-role",
        "scaling_config": [
          {
            "desired_size": 1,
            "max_size": 4,
            "min_size": 1
          }
        ],
        "subnet_ids": [
          "subnet-034a2e1a7ffdd3e7f",
          "subnet-044de3d67a78caa77"
        ],
        "tags": {
          "crossplane-kind": "nodegroup.eks.aws.upbound.io",
          "crossplane-name": "sample-eks-ng",
          "crossplane-providerconfig": "default",
          "someTag": "tag"
        }
      }
    }
  },
  "terraform": {
    "required_providers": {
      "aws": {
        "source": "hashicorp/aws",
        "version": "4.56.0"
      }
    }
  }
```

DynamoDB Table
```yaml
spec:
  managementPolicies: ["Observe", "Create", "Update", "Delete"]
  initProvider:
    writeCapacity: 20
    readCapacity: 19
  forProvider:
    region: us-west-1
    attribute:
    - name: UserId
      type: S
    - name: GameTitle
      type: S
    - name: TopScore
      type: "N"
    billingMode: PROVISIONED
    globalSecondaryIndex:
    - hashKey: GameTitle
      name: GameTitleIndex
      nonKeyAttributes:
      - UserId
      projectionType: INCLUDE
      rangeKey: TopScore
      readCapacity: 10
      writeCapacity: 10
    hashKey: UserId
    rangeKey: GameTitle
```
```json
"resource": {
    "aws_dynamodb_table": {
      "example": {
        "attribute": [
          {
            "name": "UserId",
            "type": "S"
          },
          {
            "name": "GameTitle",
            "type": "S"
          },
          {
            "name": "TopScore",
            "type": "N"
          }
        ],
        "billing_mode": "PROVISIONED",
        "global_secondary_index": [
          {
            "hash_key": "GameTitle",
            "name": "GameTitleIndex",
            "non_key_attributes": [
              "UserId"
            ],
            "projection_type": "INCLUDE",
            "range_key": "TopScore",
            "read_capacity": 10,
            "write_capacity": 10
          }
        ],
        "hash_key": "UserId",
        "lifecycle": {
          "ignore_changes": [
            "read_capacity",
            "write_capacity"
          ],
          "prevent_destroy": true
        },
        "name": "example",
        "range_key": "GameTitle",
        "read_capacity": 19,
        "tags": {
          "crossplane-kind": "table.dynamodb.aws.upbound.io",
          "crossplane-name": "example",
          "crossplane-providerconfig": "default"
        },
        "write_capacity": 20
      }
    }
  },
```

### DIFF
--- a/pkg/pipeline/templates/terraformed.go.tmpl
+++ b/pkg/pipeline/templates/terraformed.go.tmpl
@@ -72,6 +72,16 @@ import (
         return json.TFParser.Unmarshal(p, &tr.Spec.ForProvider)
     }
 
+    // GetInitParameters of this {{ .CRD.Kind }}
+    func (tr *{{ .CRD.Kind }}) GetInitParameters() (map[string]any, error) {
+        p, err := json.TFParser.Marshal(tr.Spec.InitProvider)
+        if err != nil {
+            return nil, err
+        }
+        base := map[string]any{}
+        return base, json.TFParser.Unmarshal(p, &base)
+    }
+
     // LateInitialize this {{ .CRD.Kind }} using its observed tfState.
     // returns True if there are any spec changes for the resource.
     func (tr *{{ .CRD.Kind }}) LateInitialize(attrs []byte) (bool, error) {
--- a/pkg/resource/ignored.go
+++ b/pkg/resource/ignored.go
@@ -1,0 +1,62 @@
+/*
+Copyright 2023 Upbound Inc.
+*/
+
+package resource
+
+import (
+	"fmt"
+	"sort"
+)
+
+// GetTerraformIgnoreChanges returns a sorted Terraform `ignore_changes`
+// lifecycle meta-argument expression by looking for differences between
+// the `initProvider` and `forProvider` maps. The ignored fields are the ones
+// that are present in initProvider, but not in forProvider.
+func GetTerraformIgnoreChanges(forProvider, initProvider map[string]any) []string {
+	ignored := getIgnoredFieldsMap("%s", forProvider, initProvider)
+	// sort the ignored fields so that we can compare them easily
+	sort.Strings(ignored)
+	return ignored
+}
+
+func getIgnoredFieldsMap(format string, forProvider, initProvider map[string]any) []string {
+	ignored := []string{}
+
+	for k := range initProvider {
+		if _, ok := forProvider[k]; !ok {
+			ignored = append(ignored, fmt.Sprintf(format, k))
+		} else {
+			// both are the same type so we dont need to check for forProvider type
+			if _, ok = initProvider[k].(map[string]any); ok {
+				ignored = append(ignored, getIgnoredFieldsMap(fmt.Sprintf(format, k)+"[%q]", forProvider[k].(map[string]any), initProvider[k].(map[string]any))...)
+			}
+			// if its an array, we need to check if its an array of maps or not
+			if _, ok = initProvider[k].([]any); ok {
+				ignored = append(ignored, getIgnoredFieldsArray(fmt.Sprintf(format, k), forProvider[k].([]any), initProvider[k].([]any))...)
+			}
+
+		}
+	}
+	return ignored
+}
+
+func getIgnoredFieldsArray(format string, forProvider, initProvider []any) []string {
+	ignored := []string{}
+	for i := range initProvider {
+		// Construct the full field path with array index and prefix.
+		fieldPath := fmt.Sprintf("%s[%d]", format, i)
+		if i < len(forProvider) {
+			if _, ok := initProvider[i].(map[string]any); ok {
+				ignored = append(ignored, getIgnoredFieldsMap(fieldPath+".%s", forProvider[i].(map[string]any), initProvider[i].(map[string]any))...)
+			}
+			if _, ok := initProvider[i].([]any); ok {
+				ignored = append(ignored, getIgnoredFieldsArray(fieldPath+"%s", forProvider[i].([]any), initProvider[i].([]any))...)
+			}
+		} else {
+			ignored = append(ignored, fieldPath)
+		}
+
+	}
+	return ignored
+}
--- a/pkg/resource/ignored_test.go
+++ b/pkg/resource/ignored_test.go
@@ -1,0 +1,298 @@
+/*
+Copyright 2023 Upbound Inc.
+*/
+
+package resource
+
+import (
+	_ "embed"
+	"testing"
+
+	"github.com/google/go-cmp/cmp"
+)
+
+func TestGetIgnoredFields(t *testing.T) {
+	type args struct {
+		initProvider map[string]any
+		forProvider  map[string]any
+	}
+	type want struct {
+		ignored []string
+	}
+	cases := map[string]struct {
+		reason string
+		args
+		want
+	}{
+		"InitProviderEmpty": {
+			reason: "Should return the empty ignored fields when initProvider is empty",
+			args: args{
+				forProvider: map[string]any{
+					"singleField": "bob",
+				},
+			},
+			want: want{
+				ignored: []string{},
+			},
+		},
+		"ForProviderEmpty": {
+			reason: "Should return the ignored fields when forProvider is empty",
+			args: args{
+				initProvider: map[string]any{
+					"singleField": "bob",
+				},
+			},
+			want: want{
+				ignored: []string{"singleField"},
+			},
+		},
+		"SingleField": {
+			reason: "Should return the single ignored field when forProvider has a different field",
+			args: args{
+				initProvider: map[string]any{
+					"singleField": "bob",
+				},
+				forProvider: map[string]any{
+					"otherField": "bob",
+				},
+			},
+			want: want{
+				ignored: []string{"singleField"},
+			},
+		},
+		"SingleFieldSame": {
+			reason: "Should not return the single ignored field when forProvider has same fields",
+			args: args{
+				initProvider: map[string]any{
+					"singleField": "bob",
+				},
+				forProvider: map[string]any{
+					"singleField": "bob",
+				},
+			},
+			want: want{
+				ignored: []string{},
+			},
+		},
+		"NestedArrayField": {
+			reason: "Should return the ignored field when init and for provider have nested arrays with different fields",
+			args: args{
+				initProvider: map[string]any{
+					"array": []any{
+						map[string]any{
+							"singleField": "bob",
+							"same":        "same",
+						},
+					},
+				},
+				forProvider: map[string]any{
+					"array": []any{
+						map[string]any{
+							"otherField": "bob",
+							"same":       "not same value",
+						},
+					},
+				},
+			},
+			want: want{
+				ignored: []string{"array[0].singleField"},
+			},
+		},
+		"NestedArrayFieldMultiple": {
+			reason: "Should return the multiple ignored fields when init and for provider have nested arrays with different fields",
+			args: args{
+				initProvider: map[string]any{
+					"array": []any{
+						map[string]any{
+							"singleField": "bob",
+							"same":        "same",
+						},
+						map[string]any{
+							"nextField": "bob",
+						},
+					},
+				},
+				forProvider: map[string]any{
+					"array": []any{
+						map[string]any{
+							"otherField": "bob",
+							"same":       "not same value",
+						},
+					},
+				},
+			},
+			want: want{
+				ignored: []string{"array[0].singleField", "array[1]"},
+			},
+		},
+		"NestedArraySameField": {
+			reason: "Should not the ignored field when init and for provider have nested arrays with same fields",
+			args: args{
+				initProvider: map[string]any{
+					"array": []any{
+						map[string]any{
+							"same": "same",
+						},
+					},
+				},
+				forProvider: map[string]any{
+					"array": []any{
+						map[string]any{
+							"same": "not same value",
+						},
+					},
+				},
+			},
+			want: want{
+				ignored: []string{},
+			},
+		},
+		"NestedMapKey": {
+			reason: "Should return the ignored field when init and for provider have nested maps with different keys",
+			args: args{
+				initProvider: map[string]any{
+					"map": map[string]any{
+						"key":  "bob",
+						"same": "same",
+					},
+				},
+				forProvider: map[string]any{
+					"map": map[string]any{
+						"otherKey": "bob",
+						"same":     "not same value",
+					},
+				},
+			},
+			want: want{
+				ignored: []string{"map[\"key\"]"},
+			},
+		},
+		"NestedMapSameKey": {
+			reason: "Should not return the ignored field when init and for provider have nested maps with same keys",
+			args: args{
+				initProvider: map[string]any{
+					"map": map[string]any{
+						"same": "same",
+					},
+				},
+				forProvider: map[string]any{
+					"map": map[string]any{
+						"same": "not same value",
+					},
+				},
+			},
+			want: want{
+				ignored: []string{},
+			},
+		},
+		"MissingMapField": {
+			reason: "Should return the ignored field when init and for provider have different nested maps",
+			args: args{
+				initProvider: map[string]any{
+					"map": map[string]any{
+						"key": "bob",
+					},
+				},
+				forProvider: map[string]any{
+					"othermap": map[string]any{
+						"key": "bob",
+					},
+				},
+			},
+			want: want{
+				ignored: []string{"map"},
+			},
+		},
+		"MissingArrayField": {
+			reason: "Should return the ignored field when init and for provider have different nested arrays",
+			args: args{
+				initProvider: map[string]any{
+					"array": []any{
+						"bob",
+					},
+				},
+				forProvider: map[string]any{
+					"otherArray": []any{
+						"bob",
+					},
+				},
+			},
+			want: want{
+				ignored: []string{"array"},
+			},
+		},
+		"MissingMultipleFields": {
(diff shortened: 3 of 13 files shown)


lsviben:
Checked the size change for the CRDs in provider-aws:

```bash
~/go/provider-aws/package (main) [stash: 1]$ du -sh crds
27M	crds

~/go/provider-aws/package (granular_management_policies) [stash: 1]$ du -sh crds
31M	crds
```
Which is around a ~15% increase to the CRDs size. Ill check for others as well when I generate them

pkg/resource/ignored.go
```diff
@@ -0,0 +1,60 @@
...
+package resource
+
+import (
+	"fmt"
+	"sort"
+)
+
+// GetIgnorableFields returns a sorted list of ignorable fields by looking for
+// differences between the `initProvider` and `forProvider` maps. The ignored
+// fields are the ones that are present in initProvider, but not in forProvider.
```
turkenh, line 14: Should we error out when the same field is defined both in `initProvider` and `forProvider` or silently ignore the `initProvider` one by overriding it with the one from `forProvider` ? It looks like you implemented the latter, but wondering if we should rather error out considering there is no reason to have such a configuration.

pkg/resource/ignored_test.go
```diff
@@ -0,0 +1,271 @@
...
+						map[string]any{
+							"same": "not same value",
+						},
+					},
+				},
+			},
+			want: want{
+				ignored: []string{},
+			},
+		},
```
turkenh, line 122: I just tried the following:

Screenshot 2023-07-24 at 14 28 21

But got:

```
    ignored_test.go:270: GetIgnorableFields() got = [array[0].singleField], want [array[0].singleField array[1].newField]
--- FAIL: TestGetIgnoredFields (0.00s)
```

pkg/types/builder.go
```diff
@@ -282,23 +314,51 @@ func (r *resource) addParameterField(f *Field, field *types.Var) {
...
-	} else {
+	// Note(lsviben): Only identifier fields should be required,
+	// as they are unique for spec.forProvider.
+	if req && f.Identifier {
 		// Required fields should not have omitempty tag in json tag.
 		// TODO(muvaf): This overrides user intent if they provided custom
 		// JSON tag.
 		r.paramTags = append(r.paramTags, fmt.Sprintf(`json:"%s" tf:"%s"`, strings.TrimSuffix(f.JSONTag, ",omitempty"), f.TFTag))
+	} else {
+		r.paramTags = append(r.paramTags, fmt.Sprintf(`json:"%s" tf:"%s"`, f.JSONTag, f.TFTag))
```
turkenh, line 336: Not sure exactly where but looks like we lost `+kubebuilder:validation:Required` for identifier arguments: https://github.com/upbound/provider-aws/pull/785/files#diff-7a8cb7fd3694d62f4b9b059131d2d57c12546cab7142c46c26c1d08ee87dc7e2L40

Screenshot 2023-07-24 at 14 43 42

turkenh:
> Added code generation for spec.initProvider which is based on spec.forProvider with some exceptions:
> 
> - does not inclued Identifiers - they should be spec.forProvider specific due to their nature
> - does not include references
> - does not include sensitive fields
> - does not include non-terraform fields
> 
> CEL rules are now updated to generate the required check in spec.forProvider.<path?> and spec.initProvider.<path?` if the field is not one of the above.

@lsviben I am a bit lost with when (i.e. managementPolicy) we use which (required/optional vs CEL) validation on where (`initProvider` vs `forProvider`) for different parameter types (identifier/required/optional). It would be great if you could provide a summary of what to expect in that regard.

pkg/resource/ignored.go
lsviben: ideally if the same field is defined in both, we would just disregard the 'initProvider' field. But as we use mergo.Merge, to achieve the goal of merging maps and arrays, we use the option mergo.WithSliceDeepCopy which includes overrides. I wanted to play a bit more with it to see if we can get the desired effect, but not sure if possible.

If we error out, it would fix that problem at least :D. Not sure which is the better UX though

pkg/resource/ignored_test.go
lsviben: good catch, fixed. The result should be 
```go 
[]string{"array[0].singleField", "array[1]"}
```

because there is not equivalent `forProvider` array entry, so the whole array element should be ignored.

pkg/types/builder.go
lsviben: Seems i removed setting the required comment. Reverted it now and modified to be required if its required and the field is not an initProvider field

turkenh: Ok, now I am seeing the required marker is back for identifier but optional ones are still gone. Any reason not to keep `// +kubebuilder:validation:Optional` markers for optional parameters?

pkg/resource/ignored.go
turkenh: I cannot think of a meaningful reason behind defining a parameter in both places. But, if that happens, one could expect the one under `forProvider` should override the one under `initProvider`. I am fine as long as we could implement one of the two options.

turkenh (approved):
LGTM, except I want to make sure to minimize the diff caused by this PR by putting `// +kubebuilder:validation:Optional` markers back as I stated [here](https://github.com/upbound/upjet/pull/237#discussion_r1275339906).

pkg/types/builder.go
lsviben: Returned them. 

We could also put the `// +kubebuilder:validation:Optional` on `initProvider` and `atProvider` fields if we want. `atProvider` did not have them earlier though. Not sure what difference it makes, it seems they are optional with or without them

pkg/types/builder.go
```diff
@@ -282,23 +315,48 @@ func (r *resource) addParameterField(f *Field, field *types.Var) {
 	// - req => required
 	// - !f.Identifier => not identifiers - i.e. region, zone, etc.
 	// - len(f.CanonicalPaths) == 1 => top level, i.e. not a nested field
+	// TODO (lsviben): We should add CEL rules for all required fields,
+	// not just the top level ones, due to having all forProvider
+	// fields now optional. CEL rules should check if a field is
+	// present either in forProvider or initProvider.
```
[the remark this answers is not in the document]
lsviben: Added an issue to tackle this: https://github.com/upbound/upjet/issues/239

pkg/terraform/files.go
```diff
@@ -52,14 +62,38 @@ func NewFileProducer(ctx context.Context, client resource.SecretClient, dir stri
...
+		if err != nil {
+			return nil, errors.Wrap(err, "cannot get init parameters")
+		}
+
+		// get fields which should be in the ignore_changes lifecycle block
+		fp.ignored = resource.GetIgnorableFields(params, initParams)
+
+		// TODO(lsviben): Currently initProvider fields override forProvider
+		// fields if set. It should be the other way around.
+		// merge the initProvider and forProvider parameters
```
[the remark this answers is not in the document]
lsviben: Added an issue to tackle this: https://github.com/upbound/upjet/issues/240

pkg/terraform/files.go
```diff
@@ -52,14 +62,39 @@ func NewFileProducer(ctx context.Context, client resource.SecretClient, dir stri
...
+		// https://github.com/upbound/upjet/issues/240
+		err = mergo.Merge(&params, initParams, mergo.WithSliceDeepCopy)
+		if err != nil {
+			return nil, errors.Wrap(err, "cannot merge init and forProvider parameters")
```
ulucinar, line 94:
```suggestion
			return nil, errors.Wrapf(err, "cannot merge the spec.initProvider and spec.forProvider parameters for the resource %q", tr.GetName())
```

pkg/types/builder.go
```diff
@@ -31,8 +32,9 @@ type Generated struct {
 	Types    []*types.Named
 	Comments twtypes.Comments
 
-	ForProviderType *types.Named
-	AtProviderType  *types.Named
+	ForProviderType  *types.Named
+	InitProviderType *types.Named
```
ulucinar, line 36: Not for this PR, but should we consider putting the generation of the `spec.initProvider` API behind a configuration parameter (make the generation of the API optional)?

How do we feel about the following question? Will all the providers generated by upjet need this API on their MRs? From our previous discussions, my understanding is that not every MR (or provider in the higher level) may need this API. If this is the case, then we can consider opening an issue in upjet so that we can parameterize the code generation pipeline.

pkg/types/field.go
```diff
@@ -231,24 +235,54 @@ func (f *Field) AddToResource(g *Builder, r *resource, typeNames *TypeNames) {
...
+//
+// Because of how upjet works, the main.tf file is created and filled
+// in the Connect step of the reconciliation. So we merge the initProvider
+// and forProvider there and write it to the main.tf file. So fields that are
+// not part of terraform are not included in this merge, plus they cant be
+// ignored through ignore_changes. References similarly get resolved in
+// an earlier step, so they cannot be included as well. Plus probably they
+// should also not change for Create and Update steps.
+func (f *Field) isInit() bool {
+	return !f.Identifier && f.Reference == nil && f.TFTag != "-"
```
ulucinar, line 283: We had an offline discussion with @lsviben on this. Please see the above discussion on the aggregation APIs.

pkg/pipeline/templates/terraformed.go.tmpl
```diff
@@ -72,6 +72,25 @@ import (
...
+        p, err := json.TFParser.Marshal(tr.Spec.InitProvider)
+        if err != nil {
+            return nil, err
+        }
+        base := map[string]any{}
+        return base, json.TFParser.Unmarshal(p, &base)
+    }
+
+    // SetInitParameters for this {{ .CRD.Kind }}
+    func (tr *{{ .CRD.Kind }}) SetInitParameters(params map[string]any) error {
```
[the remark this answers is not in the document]
lsviben: Mostly for completeness, i can remove it.

pkg/types/builder.go
lsviben: I would say that probably because of tags it will be needed for all resources. But as an alpha feature we can see how it is used. We could add a configuration parameter not to generate initProviders for some fields, and just put an empty struct for.

PR #231, opened 2023-07-14
Requeue the reconcile request on async update error

ulucinar:
### Description of your changes

Fixes #176 
Fixes #180

This PR employs [the new `DesiredStatusChanged` predicate](https://github.com/crossplane/crossplane-runtime/pull/372) predicate from the crossplane-runtime as an event filter to prevent too-frequent reconciles on error conditions or on updates to the external create [pending](https://github.com/crossplane/crossplane-runtime/blob/4374a3b5d1ff58214b118a7cea641230bfbc785f/pkg/meta/meta.go#L45)/[failed](https://github.com/crossplane/crossplane-runtime/blob/4374a3b5d1ff58214b118a7cea641230bfbc785f/pkg/meta/meta.go#L59) annotations. It also allows the [async callback functions](https://github.com/upbound/upjet/blob/b70acc1bcb6d42499f0310b391f07694d0a97448/pkg/terraform/workspace.go#L131) to requeue a reconcile request with an exponential back-off, when the async operation fails. 

I have:

- [x] Read and followed Crossplane's [contribution process].
- [x] Run `make reviewable` to ensure this PR is ready for review.
- [ ] Added `backport release-x.y` labels to auto-backport this PR if necessary.

### How has this code been tested

Tested the PR with https://github.com/upbound/provider-aws/pull/789.
When, for example, the external connector's `Connect` call fails, the requeue request is exponentially backed-off with a cap of 1 min. Here are the sample timestamps from an experiment with a `Connect` call failing:
```
2023-07-26T22:40:35+03:00
2023-07-26T22:40:36+03:00
2023-07-26T22:40:37+03:00
2023-07-26T22:40:41+03:00
2023-07-26T22:40:49+03:00
2023-07-26T22:41:05+03:00
2023-07-26T22:41:37+03:00
2023-07-26T22:42:37+03:00
2023-07-26T22:43:37+03:00
```

To test whether we exponentially back-off when we error in asynchronous callbacks, I provisioned an S3 bucket, put a file into it, and attempted to delete the bucket. Here are the timestamps belonging to the (failing) reconciliation attempts while deleting the `Bucket.s3` MR:
```
2023-07-26T23:03:35+03:00
2023-07-26T23:03:42+03:00
2023-07-26T23:03:48+03:00
2023-07-26T23:03:57+03:00
2023-07-26T23:04:06+03:00
2023-07-26T23:04:27+03:00
2023-07-26T23:05:04+03:00
2023-07-26T23:06:09+03:00
[...]

PR #241, opened 2023-07-27
Explicitly Queue a Reconcile Request if a Shared Provider has Expired

### LINKED ISSUE #192: Confusing "budget exceeded (ttl)" message in shared schdeuler
After enabling shared schdeuler, some logs about ttl budget appear in providers.

```
cannot schedule a native provider during observe: 4626d819-cc8a-4660-8b1b-4a05476a6122: cannot schedule native Terraform provider process: native provider reuse budget has been exceeded: invocationCount: 121, ttl: 100
```

These logs get confusing at times. In addition, these logs have been encountered many times on the community side as they are also visible in MR statuses.

We may consider fixing this confusing message in upjet side.

ulucinar:

Fixes #192 

We currently return an error if a [terraform.SharedProvider](https://github.com/upbound/upjet/blob/cd958ec5d17665415fe1ca31724bb8ffada9e235/pkg/terraform/provider_runner.go#L79) has already expired and an external client requests service from the expired shared provider. This results in the MR being reconciled by the external client to acquire a `Synced: False` status condition. This is a temporary situation that's resolved once the expired shared provider is drained and the [terraform.SharedProviderScheduler](https://github.com/upbound/upjet/blob/cd958ec5d17665415fe1ca31724bb8ffada9e235/pkg/terraform/provider_scheduler.go#L117) replaces it.

However, until the shared provider is replaced, the MRs entering into the `Synced: False` state become confusing. Before the `handler.EventHandler` implementation from #231, we had to return an error to the managed reconciler to requeue a timely reconciliation request. However, we can now explicitly requeue an exponentially backed-off reconciliation request so that the external client will retry to get service from the provider in a timely manner without having to wait for the poll period (default 10 min for the official providers) and without forcing the MR falling into the `Synced: False` state.

The external client will retry to schedule a shared runner up to 20 times without erroring, exponentially backing-off after each failure. And after this initial 20 retries, it will report the error to the managed reconciler and keep trying.

I have:

Tested via https://github.com/upbound/provider-aws/pull/805.

### DIFF
--- a/pkg/controller/external.go
+++ b/pkg/controller/external.go
@@ -8,6 +8,8 @@ import (
 	"context"
 	"time"
 
+	tferrors "github.com/upbound/upjet/pkg/terraform/errors"
+
 	xpv1 "github.com/crossplane/crossplane-runtime/apis/common/v1"
 	"github.com/crossplane/crossplane-runtime/pkg/logging"
 	"github.com/crossplane/crossplane-runtime/pkg/reconciler/managed"
@@ -17,6 +19,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	"github.com/upbound/upjet/pkg/config"
+	"github.com/upbound/upjet/pkg/controller/handler"
 	"github.com/upbound/upjet/pkg/metrics"
 	"github.com/upbound/upjet/pkg/resource"
 	"github.com/upbound/upjet/pkg/resource/json"
@@ -34,8 +37,13 @@ const (
 	errStartAsyncDestroy = "cannot start async destroy"
 	errApply             = "cannot apply"
 	errDestroy           = "cannot destroy"
-	errScheduleProvider  = "cannot schedule native Terraform provider process"
+	errScheduleProvider  = "cannot schedule native Terraform provider process, please consider increasing its TTL with the --provider-ttl command-line option"
 	errUpdateAnnotations = "cannot update managed resource annotations"
+)
+
+const (
+	rateLimiterScheduler = "scheduler"
+	retryLimit           = 20
 )
 
 // Option allows you to configure Connector.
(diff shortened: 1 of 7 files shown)


pkg/controller/external.go
```diff
@@ -149,10 +174,20 @@ func (e *external) Observe(ctx context.Context, mg xpresource.Managed) (managed.
...
+	requeued, err := e.scheduleProvider(mg.GetName())
+	if err != nil {
 		return managed.ExternalObservation{}, errors.Wrapf(err, "cannot schedule a native provider during observe: %s", mg.GetUID())
 	}
+	if requeued {
+		// return a noop for Observe after requeuing the reconcile request
+		// for a retry.
+		return managed.ExternalObservation{
+			ResourceExists:   true,
+			ResourceUpToDate: true,
```
sergenyalcin, line 186: nit: I do not have a strong opinion, but I think the case of a resource that has never been scheduled and that of a resource that has been scheduled and reconciled several times (at least within the application) can be evaluated differently.

So, we may consider returning an error for a resource that has never been scheduled.

sergenyalcin (approved):
Thanks @ulucinar LGTM. I left a nit comment for further discussions

pkg/controller/external.go
ulucinar: Thank you @sergenyalcin for considering this. I agree with your point but I also feel a bit uncomfortable with continuing to produce unsync'ed MRs even for certain cases since this change is a UX change.

As we have discussed, I also feel uncomfortable for returning a direct success from the external client's `Create` function to the managed reconciler but the situation is similar to what we do in the async mode of operation, i.e., we already return an immediate success by just starting a goroutine that may very well fail to initiate a creation request for the external resource.

I propose we continue with supressing the unsync'ed state of MRs and if folks complain about prolonging wait times than we already have a solution for this: If you don't want to wait, please give more memory to the native provider and increase the `provider-ttl` command-line parameter. 

Maybe we will just end up implementing a scheduler that will not deny requests coming to an expired runner but will fork a new one as needed at the cost of running multiple runners in parallel and increasing the memory footprint. Let's try the approach implemented here first as we already have the `provider-ttl` parameter.

sergenyalcin (approved):
Thanks @ulucinar LGTM!

PR #257, opened 2023-08-17
revert leaving optional for required init fields

lsviben:

Related to https://github.com/upbound/provider-gcp/issues/355 .

In https://github.com/upbound/upjet/pull/237, the required fields in `spec.forProvider` were set as optional due to them being replacable by the fields in `spec.initProvider` (just the fields that are in `spec.initProvider` not all required fields like identifiers).

This also added, or rather didnt remove, the `omitempty` json tag for those fields. This causes boolean fields to not be properly late-initialized, which in turn causes the resource to be stuck in `Synced: FALSE` status as it does not have all the required fields.

So this PR reverts the change in which the `omitempty` tag was not being removed for fields that are required && initProvider field. Instead it removes the `omitempty` for all fields that are required.

Fixes #

I have:

Ran `make generate` on provider-gcp and tested running locally with a kind cluster.

The resource created is the cluster from the bug issue: https://github.com/upbound/provider-gcp/issues/355#issuecomment-1672889502
```yaml
apiVersion: container.gcp.upbound.io/v1beta1
kind: Cluster
metadata:
  annotations:
    meta.upbound.io/example-id: container/v1beta1/cluster
  labels:
    testing.upbound.io/example-name: cluster
  name: cluster
spec:
  forProvider:
    location: europe-north1
    ipAllocationPolicy:
      - {}
    enableAutopilot: true
```

The result is:
```bash
kubectl get -f cluster.yaml 
NAME      READY   SYNCED   EXTERNAL-NAME   AGE
cluster   True    True     cluster         89m
```

### DIFF
--- a/pkg/types/builder.go
+++ b/pkg/types/builder.go
@@ -306,13 +306,13 @@ func newTopLevelRequiredParam(path string, includeInit bool) *topLevelRequiredPa
 }
 
 func (r *resource) addParameterField(f *Field, field *types.Var) {
-	req := !f.Schema.Optional
+	requiredBySchema := !f.Schema.Optional
 	// Note(turkenh): We are collecting the top level required parameters that
 	// are not identifier fields. This is for generating CEL validation rules for
 	// those parameters and not to require them if the management policy is set
 	// Observe Only. In other words, if we are not creating or managing the
 	// resource, we don't need to provide those parameters which are:
-	// - req => required
+	// - requiredBySchema => required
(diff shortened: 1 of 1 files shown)


turkenh (approved):
Thanks @lsviben 💪

lsviben:
@ulucinar @sergenyalcin 

would adding a label according to ` Added backport release-x.y labels to auto-backport this PR if necessary.` work in this repo? Ideally this would be backported and then patched to the provider releases

haarchri (approved):
LGTM

PR #260, opened 2023-08-18
remove overwrite from initProvider-forProvider merge

### LINKED ISSUE #240: Improve spec.initProvider to spec.forProvider merge
Introduced with https://github.com/upbound/upjet/pull/237
When management policies are enabled, we are using the `spec.initProvider` field, and merging it to `spec.forProvider`. In addition, 
we are calculating the fields which are part of `spec.initProvider` but not `spec.forProvider` and setting those fields into `ignore_changes` lifecycle hooks.

The issue here is that we are using [mergo](https://github.com/imdario/mergo) in a way that `spec.initProvider` fields override `spec.forProvder` fields:

```go
// TODO(lsviben): Currently initProvider fields override forProvider
// fields if set. It should be the other way around.
// merge the initProvider and forProvider parameters
err = mergo.Merge(&params, initParams, mergo.WithSliceDeepCopy)
if err != nil {
	return nil, errors.Wrap(err, "cannot merge init and forProvider parameters")
}
```

This is due to using `mergo.WithSliceDeepCopy` which is on the other hand needed to merge arrays. 

Ideally, `spec.forProvider` fields would take precedence over the `spec.initProvider` ones. Maybe there is an option with mergo to do something like that, or it wont be difficult to write ourselves.

But, as we don't recommenced or expect the same fields to be set, we could also error out if they set the same fields.

lsviben:

This resolves a TODO left from https://github.com/upbound/upjet/pull/237 where during the `spec.initProvider` to `spec.forProvider` merge the fields in `spec.initProvider` would overwrite the `spec.forProvider` ones, instead
of skipping those fields. 

Although we would expect and suggest not to set the same fields for both, it still good to fix this behaviour.

To merge the 2, we used [mergo](https://github.com/darccio/mergo), with the option SliceDeepCopy to merge
the slices from initProvider to forProvider.

```go
err = mergo.Merge(&params, initParams, mergo.WithSliceDeepCopy)
```

Looking deeper, we can see the the `mergo.WithSliceDeepCopy` also [sets the Overwrite to true](https://github.com/darccio/mergo/blob/1475161b8f00ae26a13f5baa3debc0e3d9a85582/merge.go#L368-L372)

So simply setting the overwrite to false does the trick.
```go
err = mergo.Merge(&params, initParams, mergo.WithSliceDeepCopy, func(c *mergo.Config) {
	c.Overwrite = false
})
```

Fixes #240 

I have:

- [] ~Added `backport release-x.y` labels to auto-backport this PR if necessary.~

Extended the unit test to test for overwrites.

### DIFF
--- a/pkg/terraform/files.go
+++ b/pkg/terraform/files.go
@@ -85,11 +85,14 @@ func NewFileProducer(ctx context.Context, client resource.SecretClient, dir stri
 		// get fields which should be in the ignore_changes lifecycle block
 		fp.ignored = resource.GetTerraformIgnoreChanges(params, initParams)
 
-		// TODO(lsviben): Currently initProvider fields override forProvider
-		// fields if set. It should be the other way around.
-		// merge the initProvider and forProvider parameters
-		// https://github.com/upbound/upjet/issues/240
-		err = mergo.Merge(&params, initParams, mergo.WithSliceDeepCopy)
+		// Note(lsviben): mergo.WithSliceDeepCopy is needed to merge the
+		// slices from the initProvider to forProvider. As it also sets
+		// overwrite to true, we need to set it back to false, we don't
+		// want to overwrite the forProvider fields with the initProvider
+		// fields.
+		err = mergo.Merge(&params, initParams, mergo.WithSliceDeepCopy, func(c *mergo.Config) {
+			c.Overwrite = false
+		})
 		if err != nil {
 			return nil, errors.Wrapf(err, "cannot merge the spec.initProvider and spec.forProvider parameters for the resource %q", tr.GetName())
 		}
--- a/pkg/terraform/files_test.go
+++ b/pkg/terraform/files_test.go
@@ -424,22 +424,25 @@ func TestWriteMainTF(t *testing.T) {
 						"param": "paramval",
 						"array": []any{
 							map[string]any{
-								"other": "val",
+								"other": "val1",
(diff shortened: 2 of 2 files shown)


ulucinar (approved):
Thanks @lsviben, lgtm.
Looks like we need to fix the reporting of some tests in `pkg/terraform/files_test.go` but this is not a blocker for this PR. E.g.,
```
diff --git a/pkg/terraform/files_test.go b/pkg/terraform/files_test.go
index ec38af8..353fd31 100644
--- a/pkg/terraform/files_test.go
+++ b/pkg/terraform/files_test.go
@@ -488,8 +488,8 @@ func TestWriteMainTF(t *testing.T) {
                        if err = json.JSParser.Unmarshal([]byte(tc.want.maintf), &wantJson); err != nil {
                                t.Errorf("cannot unmarshal want main.tf.json: %v", err)
                        }
-                       if diff := cmp.Diff(res, wantJson, test.EquateConditions()); diff != "" {
-                               t.Errorf("\n%s\nWriteMainTF(...): -want error, +got error:\n%s", tc.reason, diff)
+                       if diff := cmp.Diff(wantJson, res); diff != "" {
+                               t.Errorf("\n%s\nWriteMainTF(...): -want main.tf.json, +got main.tf.json:\n%s", tc.reason, diff)
                        }
                })
        }
```

PR #262, opened 2023-08-22
Queue a reconcile request after marking the resource as ready

ulucinar:

With this PR, right after marking a managed resource as ready (adding the `Ready=True` status condition), we queue a reconcile request because the status updates will no longer trigger a reconciliation and there may be a need for a subsequent reconcile for late-initialization. Currently, we need to wait for the poll interval after marking a resource as ready before it's late-initialized as we prioritize marking a resource as ready over late-initializing it unlike the community providers (this is at least the case for the community AWS provider). 

We are also planning to change this behavior and simplify the logic implemented [here](https://github.com/upbound/upjet/blob/e620c6228964310a1535e6d55030b85bfaac8b61/pkg/controller/external.go#L295) in a future PR.

I have:

Please see the comment [below](https://github.com/upbound/upjet/pull/262#pullrequestreview-1591054002).

### DIFF
--- a/pkg/controller/external.go
+++ b/pkg/controller/external.go
@@ -43,6 +43,7 @@ const (
 
 const (
 	rateLimiterScheduler = "scheduler"
+	rateLimiterStatus    = "status"
 	retryLimit           = 20
 )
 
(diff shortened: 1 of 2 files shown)


sergenyalcin (approved):
I tested this PR against to the provider-gcp and I validated a reconciliation is requested after making the resource as ready.

Thanks @ulucinar LGTM.

ulucinar:
Thank you @sergenyalcin!

PR #269, opened 2023-09-01
CEL rules extension

### LINKED ISSUE #239: Extend CEL rules to nested structs
With the introduction of https://github.com/upbound/upjet/pull/237 , we have made all the generated `spec.forProvider` fields which are also `spec.initProvider` fields non-required in the schema, as they can be set either in one or the other. This means that there is no validation currently for these fields.

With [CEL rules](https://kubernetes.io/docs/reference/using-api/cel/), we could make checks which validate if the field is set either in `spec.forProvider` or `spec.initProvider`. We already have such rules, but only for top level parameters, as they were introduced with ObserveOnly. Now we should go into nested structs, which makes the path generation for CEL a bit tricky.

lsviben:

**Disclaimer** :warning: 
I am not entirely happy with this solution and am wondering if we should merge this. Putting it out for discussion and explaining my thought process below

### Intro
In https://github.com/upbound/upjet/pull/237 we made all non-init fields optional, so that they
can be set either in `spec.forProvider` or `spec.initProvider`. For top level fields, we replaced the "required" validation with
CEL rules. Such as:

```yaml
// +kubebuilder:validation:XValidation:rule="!('*' in self.managementPolicies || 'Create' in self.managementPolicies || 'Update' in self.managementPolicies) || has(self.forProvider.title) || has(self.initProvider.title)",message="title is a required parameter"
```

But we didn't have a solution for the nested fields. So the purpose of this PR is to add CEL validation for nested fields that should be required.

This proved to be tricky due to 2 factors:
- how we are generating nested fields based on Terraform schemas
- [CEL language specification ](https://github.com/google/cel-spec/blob/master/doc/langdef.md)

Starting this task I had expected to generate something like:

```yaml
... has(self.forProvider.someMap.nestedField) || ...
OR
... has(self.forProvider.someList[*].nestedField) || ...
```
But I was very wrong :cold_sweat: .

### Problem explanation
Firstly, we generate almost every nested struct as an array. Even though the resource expects and needs only one element of that array. I think its due to how terraform schemas are translated into our types, as they mark all of the nested structs as lists of nested structs. I didn't go too deep into this as it's highly unlikely that we would make some changes there and break our API. 

Example from upbound/provider-gcp Cluster:

```go
	// Per-cluster configuration of Node Auto-Provisioning with Cluster Autoscaler to
	// automatically adjust the size of the cluster and create/delete node pools based
	// on the current needs of the cluster's workload. See the
	// guide to using Node Auto-Provisioning
	// for more details. Structure is documented below.
	ClusterAutoscaling []ClusterAutoscalingParameters `json:"clusterAutoscaling,omitempty" tf:"cluster_autoscaling,omitempty"`
	
...
	type ClusterAutoscalingParameters struct {

	// Contains defaults for a node pool created by NAP. A subset of fields also apply to
	// GKE Autopilot clusters.
	// Structure is documented below.
	AutoProvisioningDefaults []AutoProvisioningDefaultsParameters `json:"autoProvisioningDefaults,omitempty" tf:"auto_provisioning_defaults,omitempty"`

	// Whether node auto-provisioning is enabled. Must be supplied for GKE Standard clusters, true is implied
	// for autopilot clusters. Resource limits for cpu and memory must be defined to enable node auto-provisioning for GKE Standard.
	Enabled *bool `json:"enabled,omitempty" tf:"enabled,omitempty"`
...

```

Secondly, it is how CEL works:
- we need to be mindful of the [scope](https://github.com/google/cel-spec/blob/master/doc/langdef.md#name-resolution)
- wildcards don't work

### Scope
For the scope, it basically means that as we want to reach both `forProvider` and `initProvider`, the CEL rule has to be defined on `spec`. We cannot define a rule in a nested field as it wont have access to `initProvider` and vice-versa. Otherwise this would be very handy and we wouldn't have a problem with field selection as we could just set the rule on the nested struct in question.

### Field selection
On the other hand, we have the problem of how to actually get through to the field we want to check if it is set. There are multiple ways.

Lets imagine we have the following resource:
```yaml
spec:
  forProvider:
    someList:
    - reqField: "I am set"
      nonReqField: "I just chill here"
```
The CEL rule can look like this if we want to check just if the first element exists. For most cases this is ok as we expect only one element. For those that are truly lists we miss
```yaml
has(self.forProvider.someList[0].reqField) || has(self.initProvider.someList[0].reqField)
```

CEL also provides an `all`, `exists`, `exists_one` macros that could be helpful:
```yaml
self.forProvider.someList.all(e, has(e.reqField)) || self.initProvider.someList.all(e, has(e.reqField)) 
```
The problem here is that the check will not work well with true lists, as we could have some elements in initProvider, and some in forProvider, which is narrow edge case, but still worth to keep in mind. If we use `exists` we come back to a similar solution as using `[0]`.
Example where all wont work well.
```yaml
spec:
  initProvider:
    someList:
    - additionalField: "just an init field"
    - reqField: "I am set"
      nonReqField: "I just chill here"
  forProvider:
    someList:
    - reqField: "I am set"
      nonReqField: "I just chill here"
    - additionalField: "my required field is set in initProvider"
```

So lets say that we use the solution with `[0]`:
```yaml
has(self.forProvider.someList[0].reqField) || has(self.initProvider.someList[0].reqField)
```
What if `someList` is optional? The field `reqField` is only required if `someList` actually exists. With the above rule, we would force users to use `someList`. Thats why we could add:
```yaml
!has(self.forProvider.someList) || has(self.forProvider.someList[0].reqField) || has(self.initProvider.someList[0].reqField)
``` 

And as resources have required fields few levels deep, we can get things like this(example is from GCP Provider AccessLevel):
```yaml
!(''*'' in self.managementPolicies || ''Create'' in self.managementPolicies
                || ''Update'' in self.managementPolicies) || !has(self.forProvider.basic)
                || !has(self.forProvider.basic[0].conditions) || !has(self.forProvider.basic[0].conditions[0].devicePolicy)
                || !has(self.forProvider.basic[0].conditions[0].devicePolicy[0].osConstraints)
                || has(self.forProvider.basic[0].conditions[0].devicePolicy[0].osConstraints[0].osType)
                || has(self.initProvider.basic[0].conditions[0].devicePolicy[0].osConstraints[0].osType)
```
which becomes unreadable, but works. 

As I said in the beginning, not sure if we should include this. Maybe somebody will see a better solution. The current PR implements the above. 

Fixes #239 

I have:

Tested with https://github.com/upbound/provider-gcp/. Ran `make generate` with my `upjet`. Created the `/apis/accesscontextmanager/v1beta1/zz_accesslevel_types.go`  CRD in a local kind cluster and tested with a few examples:

Resulting CEL rules:
```yaml
x-kubernetes-validations:
            - message: basic[0].conditions[0].devicePolicy[0].osConstraints[0].osType
                is a required parameter
              rule: '!(''*'' in self.managementPolicies || ''Create'' in self.managementPolicies
                || ''Update'' in self.managementPolicies) || !has(self.forProvider.basic)
                || !has(self.forProvider.basic[0].conditions) || !has(self.forProvider.basic[0].conditions[0].devicePolicy)
                || !has(self.forProvider.basic[0].conditions[0].devicePolicy[0].osConstraints)
                || has(self.forProvider.basic[0].conditions[0].devicePolicy[0].osConstraints[0].osType)
                || has(self.initProvider.basic[0].conditions[0].devicePolicy[0].osConstraints[0].osType)'
            - message: basic[0].conditions is a required parameter
              rule: '!(''*'' in self.managementPolicies || ''Create'' in self.managementPolicies
                || ''Update'' in self.managementPolicies) || !has(self.forProvider.basic)
                || has(self.forProvider.basic[0].conditions) || has(self.initProvider.basic[0].conditions)'
            - message: custom[0].expr[0].expression is a required parameter
              rule: '!(''*'' in self.managementPolicies || ''Create'' in self.managementPolicies
                || ''Update'' in self.managementPolicies) || !has(self.forProvider.custom)
                || !has(self.forProvider.custom[0].expr) || has(self.forProvider.custom[0].expr[0].expression)
                || has(self.initProvider.custom[0].expr[0].expression)'
            - message: custom[0].expr is a required parameter
              rule: '!(''*'' in self.managementPolicies || ''Create'' in self.managementPolicies
                || ''Update'' in self.managementPolicies) || !has(self.forProvider.custom)
                || has(self.forProvider.custom[0].expr) || has(self.initProvider.custom[0].expr)'
            - message: name is a required parameter
              rule: '!(''*'' in self.managementPolicies || ''Create'' in self.managementPolicies
                || ''Update'' in self.managementPolicies) || has(self.forProvider.name)
                || has(self.initProvider.name)'
            - message: parent is a required parameter
              rule: '!(''*'' in self.managementPolicies || ''Create'' in self.managementPolicies
                || ''Update'' in self.managementPolicies) || has(self.forProvider.parent)
                || has(self.initProvider.parent)'
            - message: title is a required parameter
              rule: '!(''*'' in self.managementPolicies || ''Create'' in self.managementPolicies
                || ''Update'' in self.managementPolicies) || has(self.forProvider.ti
```

**Minimal passing validation:**

```yaml
apiVersion: accesscontextmanager.gcp.upbound.io/v1beta1
kind: AccessLevel
metadata:
  name: access-level-service-account
spec:
  forProvider:
    name: name
    parent: parent
    title: title
```
Result: PASSES
Description: contains the 3 required top level fields

**Minimal not passing validation**
```yaml
apiVersion: accesscontextmanager.gcp.upbound.io/v1beta1
kind: AccessLevel
metadata:
  name: access-level-service-account
spec:
  forProvider:
    name: name
    parent: parent
    title: title
```
Result: FAILS
Description: 
```bash
The AccessLevel "access-level-service-account" is invalid: spec: Invalid value: "object": no such key: initProvider evaluating rule: name is a required parameter
```

**Nested validation - one level**
```yaml
apiVersion: accesscontextmanager.gcp.upbound.io/v1beta1
kind: AccessLevel
metadata:
  name: access-level-service-account
spec:
  forProvider:
    basic:
    - conditions:
      - devicePolicy:
        - requireScreenLock: false
    name: name
    parent: parent
    title: title
```
Result: PASSES
Description: sets just `basic[0].conditions[0].devicePolicy[0]`. Does not go further into `osConstraints[]`

### DIFF
--- a/pkg/types/builder.go
+++ b/pkg/types/builder.go
@@ -19,6 +19,7 @@ import (
 	"github.com/crossplane/crossplane-runtime/pkg/fieldpath"
 
 	"github.com/upbound/upjet/pkg/config"
+	"github.com/upbound/upjet/pkg/types/name"
 )
 
 const (
@@ -318,6 +319,7 @@ func newTopLevelRequiredParam(path string, includeInit bool) *topLevelRequiredPa
 
 func (r *resource) addParameterField(f *Field, field *types.Var) {
 	requiredBySchema := !f.Schema.Optional
+
 	// Note(turkenh): We are collecting the top level required parameters that
 	// are not identifier fields. This is for generating CEL validation rules for
 	// those parameters and not to require them if the management policy is set
@@ -335,7 +337,7 @@ func (r *resource) addParameterField(f *Field, field *types.Var) {
 		requiredBySchema = false
 		// If the field is not a terraform field, we should not require it in init,
 		// as it is not an initProvider field.
-		r.topLevelRequiredParams = append(r.topLevelRequiredParams, newTopLevelRequiredParam(f.TransformedName, f.TFTag != "-"))
+		r.topLevelRequiredParams = append(r.topLevelRequiredParams, newTopLevelRequiredParam(f.TransformedName, f.isInit()))
 	}
 
 	// Note(lsviben): Only fields which are not also initProvider fields should have a required kubebuilder comment.
@@ -464,4 +466,12 @@ func sanitizePath(p string) string {
 		}
 	}
 	return p
+}
+
+func constructCELPath(canonicalPath []string) string {
+	var celPath []string
+	for _, p := range canonicalPath {
+		celPath = append(celPath, sanitizePath(name.NewFromCamel(p).LowerCamelComputed))
+	}
+	return strings.Join(celPath, ".")
 }
--- a/pkg/types/builder.go
+++ b/pkg/types/builder.go
@@ -465,7 +465,7 @@ func sanitizePath(p string) string {
 }
 
 func constructCELPath(celPath []string) string {
-	var currentPath []string
+	currentPath := make([]string, 0, len(celPath))
 	// Go through the list of fields, and if a wildcard is encountered,
 	// it means that the previous field is a list, so we add "[0]" to the path.
 	// This is not ideal, but as there is no "list[*]" in CEL, we are at
@@ -493,7 +493,7 @@ func constructCELRules(celPath []string, isInit bool) *celRule {
 		return &celRule{}
 	}
 	var rule []string
-	var currentPath []string
+	currentPath := make([]string, 0, len(celPath))
 
 	// construct forProvider rule.
 	// If the field is a wildcard, it means that the previous field is a list,
--- a/pkg/types/builder.go
+++ b/pkg/types/builder.go
@@ -19,7 +19,6 @@ import (
 	"github.com/crossplane/crossplane-runtime/pkg/fieldpath"
 
 	"github.com/upbound/upjet/pkg/config"
-	"github.com/upbound/upjet/pkg/types/name"
 )
 
 const (
@@ -69,7 +68,7 @@ func NewBuilder(pkg *types.Package) *Builder {
 
 // Build returns parameters and observation types built out of Terraform schema.
 func (g *Builder) Build(cfg *config.Resource) (Generated, error) {
-	fp, ap, ip, err := g.buildResource(cfg.TerraformResource, cfg, nil, nil, false, cfg.Kind)
+	fp, ap, ip, err := g.buildResource(cfg.TerraformResource, cfg, nil, nil, nil, false, cfg.Kind)
 	return Generated{
 		Types:            g.genTypes,
 		Comments:         g.comments,
@@ -80,7 +79,7 @@ func (g *Builder) Build(cfg *config.Resource) (Generated, error) {
 	}, errors.Wrapf(err, "cannot build the Types")
 }
 
-func (g *Builder) buildResource(res *schema.Resource, cfg *config.Resource, tfPath []string, xpPath []string, asBlocksMode bool, names ...string) (*types.Named, *types.Named, *types.Named, error) { //nolint:gocyclo
+func (g *Builder) buildResource(res *schema.Resource, cfg *config.Resource, tfPath, xpPath, celPath []string, asBlocksMode bool, names ...string) (*types.Named, *types.Named, *types.Named, error) { //nolint:gocyclo
 	// NOTE(muvaf): There can be fields in the same CRD with same name but in
 	// different types. Since we generate the type using the field name, there
 	// can be collisions. In order to be able to generate unique names consistently,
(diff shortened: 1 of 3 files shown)


turkenh:
> All in all, our conclusion was that for now we dont have a good solution for the nested structs validation, and that for now it is not pressing to have that, so we could consider dropping this for now until we need it or think of a better solution.

I agree. 
I will close this PR (for now?) and deprioritize the corresponding issue (i.e. drop it from the scope of Beta graduation). /cc @jeanduplessis @ulucinar

PR #272, opened 2023-09-06
remove finalizers before deleting migrated old MRs

erhancagirici:

In MR API migration process, old API MRs have their reconciliation paused, new MRs with the target API is created, then the old API MRs are deleted. However, deletion gets stuck because they have a finalizer and their reconciliation operations are paused.

The MRs have single finalizer named `finalizer.managedresource.crossplane.io`, which ensures the deletion of native cloud provider resource, however in this case it is a no-op, since the deletion policy is `Orphan` and we already would like to keep the cloud provider resource since we are just migrating it.

This PR adds a new migration step for old MRs, that removes the finalizer.
Also, it fixes a bug in the sorting of the steps, when step count to be executed exceeds 10.

I have:

- Run unit tests
- Manually tested with VPC resource

### DIFF
--- a/pkg/migration/api_steps.go
+++ b/pkg/migration/api_steps.go
@@ -35,6 +35,7 @@ const (
 	stepEditComposites
 	stepEditClaims
 	stepDeletionPolicyOrphan
+	stepRemoveFinalizers
 	stepDeleteOldManaged
 	stepStartManaged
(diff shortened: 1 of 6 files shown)


sergenyalcin (approved):
Thanks @erhancagirici LGTM!

PR #274, opened 2023-09-07
Improve CEL message

### LINKED ISSUE #273: Misleading error pointing to spec.initProvider when required parameters are not provided
### What happened?

We are getting misleading/confusing errors when we don't provide a top-level required parameter indicating as if it should have been set under `spec.initProvider`. 

See related discussion: https://github.com/upbound/provider-azure/issues/535#issuecomment-1709472013

### How can we reproduce it?

Trying to create the following manifest:

```yaml
apiVersion: eventhub.azure.upbound.io/v1beta1
kind: EventHub
metadata:
  name: example-eventhub
spec:
  forProvider: {}
```

Fails with:

```
The EventHub "example-eventhub" is invalid:
* spec: Invalid value: "object": no such key: initProvider evaluating rule: messageRetention is a required parameter
* spec: Invalid value: "object": no such key: initProvider evaluating rule: partitionCount is a required parameter
```

lsviben:

Improves the message for toplevel CEL rules, noting the full path of the field.
Also fixes the issue where initProvider would be mentioned in the CEL message due to it
not being set.

Fixes #273 

I have:

- ~[] Added `backport release-x.y` labels to auto-backport this PR if necessary.~

Tested with the example from #273 .

Initial result:
```bash
The EventHub "example-eventhub" is invalid:
* spec: Invalid value: "object": no such key: initProvider evaluating rule: messageRetention is a required parameter
* spec: Invalid value: "object": no such key: initProvider evaluating rule: partitionCount is a required parameter
```

After applying the change:
```bash
The EventHub "example-eventhub" is invalid: 
* spec: Invalid value: "object": spec.forPovider.messageRetention is a required parameter
* spec: Invalid value: "object": spec.forPovider.partitionCount is a required parameter

```

### DIFF
--- a/pkg/types/builder.go
+++ b/pkg/types/builder.go
@@ -151,9 +151,9 @@ func (g *Builder) AddToBuilder(typeNames *TypeNames, r *resource) (*types.Named,
 		g.validationRules += "\n"
 		sp := sanitizePath(p.path)
 		if p.includeInit {
-			g.validationRules += fmt.Sprintf(`// +kubebuilder:validation:XValidation:rule="!('*' in self.managementPolicies || 'Create' in self.managementPolicies || 'Update' in self.managementPolicies) || has(self.forProvider.%s) || has(self.initProvider.%s)",message="%s is a required parameter"`, sp, sp, p.path)
+			g.validationRules += fmt.Sprintf(`// +kubebuilder:validation:XValidation:rule="!('*' in self.managementPolicies || 'Create' in self.managementPolicies || 'Update' in self.managementPolicies) || has(self.forProvider.%s) || (has(self.initProvider) && has(self.initProvider.%s))",message="spec.forProvider.%s is a required parameter"`, sp, sp, p.path)
 		} else {
-			g.validationRules += fmt.Sprintf(`// +kubebuilder:validation:XValidation:rule="!('*' in self.managementPolicies || 'Create' in self.managementPolicies || 'Update' in self.managementPolicies) || has(self.forProvider.%s)",message="%s is a required parameter"`, sp, p.path)
+			g.validationRules += fmt.Sprintf(`// +kubebuilder:validation:XValidation:rule="!('*' in self.managementPolicies || 'Create' in self.managementPolicies || 'Update' in self.managementPolicies) || has(self.forProvider.%s)",message="spec.forProvider.%s is a required parameter"`, sp, p.path)
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


pkg/types/builder.go
```diff
@@ -151,9 +151,9 @@ func (g *Builder) AddToBuilder(typeNames *TypeNames, r *resource) (*types.Named,
...
+			g.validationRules += fmt.Sprintf(`// +kubebuilder:validation:XValidation:rule="!('*' in self.managementPolicies || 'Create' in self.managementPolicies || 'Update' in self.managementPolicies) || has(self.forProvider.%s) || (has(self.initProvider) && has(self.initProvider.%s))",message="spec.forProvider.%s is a required parameter"`, sp, sp, p.path)
 		} else {
-			g.validationRules += fmt.Sprintf(`// +kubebuilder:validation:XValidation:rule="!('*' in self.managementPolicies || 'Create' in self.managementPolicies || 'Update' in self.managementPolicies) || has(self.forProvider.%s)",message="%s is a required parameter"`, sp, p.path)
+			g.validationRules += fmt.Sprintf(`// +kubebuilder:validation:XValidation:rule="!('*' in self.managementPolicies || 'Create' in self.managementPolicies || 'Update' in self.managementPolicies) || has(self.forProvider.%s)",message="spec.forProvider.%s is a required parameter"`, sp, p.path)
```
turkenh, line 156: How would it behave with the following?

```yaml
apiVersion: eventhub.azure.upbound.io/v1beta1
kind: EventHub
metadata:
  name: example-eventhub
spec:
  initProvider: 
    partitionCount: 2
    messageRetention: 1
```

Not sure if this is a valid scenario though 😅 But it should still not fail with `Invalid value: "object": no such key: forProvider` I assume.

lsviben: `spec.forProvider` is required, while `spec.initProvider` is not. That is that behaviour was even happeneing. If we remove `spec.forProvider` we get this:
```bash
The EventHub "example-eventhub" is invalid: 
* spec.forProvider: Required value
* <nil>: Invalid value: "null": some validation rules were not checked because the object was invalid; correct the existing errors to complete validation
```

lsviben: BTW I intentionally didnt put 
```
spec.forProvider.%s or spec.initProvider.%s is a required parameter
```
due to `initProvider still being alpha and to avoid confusing people. But I dont feel strongly about that. My worry is that somebody would set a value in `spec.initProvider` while the alpha feature is not turned on. WDYT?

turkenh: This sounds reasonable to me.

turkenh (approved):
Thanks for the quick fix @lsviben 🙌

Issue #300, opened 2023-09-13
Add merge strategy annotations to managed resources


negz:
### What problem are you facing?

The new `FunctionComposer` added in crossplane/crossplane#4500 uses server-side-apply (https://github.com/crossplane/crossplane/issues/4047). This means that if you're using a Composition in `mode: Pipeline` Crossplane (technically, the API server) is able to correctly merge arrays of objects, among other things.

For this to work, the CRD of the composed resource needs to include [merge strategy OpenAPI extensions](https://kubernetes.io/docs/reference/using-api/server-side-apply/#merge-strategy). For example by default arrays of objects will be replaced entirely - the CRD needs to specify `//listType: map` and a `//listMapKey` to change this behavior.

### How could Crossplane help solve your problem?

We should start adding these annotations to Crossplane types. The core types defined in crossplane-runtime (e.g. `ConditionedStatus`) would be a great start, plus very common types like AWS tags.

negz:
@CodyRay No, I don't think this will work for a `runtime.RawExtension` since the API server has no idea about the structure of the data inside that field and thus SSA can't merge it.

I'm not entirely sure how to handle that case. We could reintroduce `mergeOptions`, but I'm a little reluctant to do that given that in most cases folks should not use it and should use SSA instead.

sttts:
This might be a naive question, but who is the other actor you want to merge with? The provider that completes lists provided by a function?

negz:
@sttts Yes, it's usually the provider. There's still parts of spec that are "shared" - e.g. AWS tags.

CodyRay:
> Could you raise an issue to track your use case WRT merging/appending RawExtension fields?

Would that issue belong in https://github.com/crossplane/upjet or https://github.com/crossplane-contrib/function-patch-and-transform ?

negz:
@CodyRay Option 3 - https://github.com/crossplane/crossplane. 😉

erikgb:
@CodyRay Did you ever create an issue for this?

CodyRay:
I didn't end up creating a ticket. We found some really problematic behavior before we made it to production with `mergeOptions` where array items weren't removed from the destination when they are deleted from the source composite resource. Since I think that in the long term server-side apply is going to be better, we ended up pivoting to use a function and I suspect we'll eventually use https://github.com/crossplane-contrib/function-go-templating or do the appending within our helm chart instead.

PR #301, opened 2023-11-16
Add server-side apply merge strategy markers

negz:

Fixes https://github.com/crossplane/upjet/issues/300 (mostly?)

This commit adds SSA merge strategy markers that allow the API server to granularly merge lists, rather than atomically replacing them. Composition functions use SSA, so this change means that a function can return only the list elements it desires (e.g. tags) and those elements will be merged into any existing elements, without replacing them.

For the moment I've only covered two cases:

* Lists that we know are sets of scalar values (generated from Terraform sets)
* Maps of scalar values (generated from Terraform maps)

I'm hopeful that in both of these cases it _should_ be possible to allow the map or set to be granularly merged, not atomically replaced.

https://kubernetes.io/docs/reference/using-api/server-side-apply/#merge-strategy

I have:

I tried to add a case to `TestBuild`, but then noticed that it appears to be completely unaware of comments. If you'd like this tested, please give me some direction on how and where to do so.

See https://github.com/upbound/provider-aws/pull/968 for a PR generated using this change.

### DIFF
--- a/pkg/types/field.go
+++ b/pkg/types/field.go
@@ -15,6 +15,7 @@ import (
 	"github.com/crossplane/upjet/pkg"
 	"github.com/crossplane/upjet/pkg/config"
 	"github.com/crossplane/upjet/pkg/types/comments"
+	"github.com/crossplane/upjet/pkg/types/markers"
 	"github.com/crossplane/upjet/pkg/types/name"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 	"github.com/pkg/errors"
@@ -151,7 +152,36 @@ func NewField(g *Builder, cfg *config.Resource, r *resource, sch *schema.Schema,
 	f.FieldType = fieldType
 	f.InitType = initType
 
+	AddServerSideApplyMarkers(f)
+
 	return f, nil
+}
+
+// AddServerSideApplyMarkers adds server-side apply comment markers to indicate
+// that scalar maps and sets can be merged granularly, not replace atomically.
+func AddServerSideApplyMarkers(f *Field) {
+	switch f.Schema.Type { //nolint:exhaustive
+	case schema.TypeMap:
+		// A map should always have an element of type Schema.
+		if es, ok := f.Schema.Elem.(*schema.Schema); ok {
+			switch es.Type { //nolint:exhaustive
+			// We assume scalar types can be granular maps.
+			case schema.TypeString, schema.TypeBool, schema.TypeInt, schema.TypeFloat:
+				f.Comment.ServerSideApplyOptions.MapType = ptr.To[markers.MapType](markers.MapTypeGranular)
+			}
+		}
+	case schema.TypeSet:
+		if es, ok := f.Schema.Elem.(*schema.Schema); ok {
+			switch es.Type { //nolint:exhaustive
+			// We assume scalar types can be granular maps.
+			case schema.TypeString, schema.TypeBool, schema.TypeInt, schema.TypeFloat:
+				f.Comment.ServerSideApplyOptions.ListType = ptr.To[markers.ListType](markers.ListTypeSet)
+			}
+		}
+	}
+	// TODO(negz): Can we reliably add SSA markers for lists of objects? Do we
+	// have cases where we're turning a Terraform map of maps into a list of
+	// objects with a well-known key that we could merge on?
 }
 
 // NewSensitiveField returns a constructed sensitive Field object.
--- a/pkg/types/field.go
+++ b/pkg/types/field.go
@@ -152,7 +152,9 @@ func NewField(g *Builder, cfg *config.Resource, r *resource, sch *schema.Schema,
 	f.FieldType = fieldType
 	f.InitType = initType
 
-	AddServerSideApplyMarkers(f)
+	if !sch.Sensitive {
+		AddServerSideApplyMarkers(f)
+	}
 
 	return f, nil
 }
@@ -173,7 +175,7 @@ func AddServerSideApplyMarkers(f *Field) {
 	case schema.TypeSet:
 		if es, ok := f.Schema.Elem.(*schema.Schema); ok {
 			switch es.Type { //nolint:exhaustive
-			// We assume scalar types can be granular maps.
+			// We assume scalar types can be granular sets.
 			case schema.TypeString, schema.TypeBool, schema.TypeInt, schema.TypeFloat:
 				f.Comment.ServerSideApplyOptions.ListType = ptr.To[markers.ListType](markers.ListTypeSet)
 			}
--- a/pkg/types/markers/ssa.go
+++ b/pkg/types/markers/ssa.go
@@ -20,7 +20,7 @@ const (
 	// elements.
 	ListTypeSet ListType = "set"
 
-	// ListTypeSet can be granularly merged, and different managers can own
+	// ListTypeMap can be granularly merged, and different managers can own
 	// different elements in the list. The list can include only nested types
 	// (i.e. objects).
 	ListTypeMap ListType = "map"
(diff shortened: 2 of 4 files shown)


pkg/types/field.go
```diff
@@ -151,9 +152,38 @@ func NewField(g *Builder, cfg *config.Resource, r *resource, sch *schema.Schema,
...
+			switch es.Type { //nolint:exhaustive
+			// We assume scalar types can be granular maps.
+			case schema.TypeString, schema.TypeBool, schema.TypeInt, schema.TypeFloat:
+				f.Comment.ServerSideApplyOptions.ListType = ptr.To[markers.ListType](markers.ListTypeSet)
+			}
+		}
+	}
+	// TODO(negz): Can we reliably add SSA markers for lists of objects? Do we
+	// have cases where we're turning a Terraform map of maps into a list of
+	// objects with a well-known key that we could merge on?
```
negz, line 184: Happy to expand this PR to cover this case too, if it's possible. Are there heuristics I can use to detect TF schemas that become Kubernetes slices-of-structs, where some fields in the struct (e.g. `name`) uniquely identify them?

pkg/types/field.go
```diff
@@ -151,9 +152,38 @@ func NewField(g *Builder, cfg *config.Resource, r *resource, sch *schema.Schema,
...
+			switch es.Type { //nolint:exhaustive
+			// We assume scalar types can be granular maps.
+			case schema.TypeString, schema.TypeBool, schema.TypeInt, schema.TypeFloat:
+				f.Comment.ServerSideApplyOptions.MapType = ptr.To[markers.MapType](markers.MapTypeGranular)
+			}
+		}
+	case schema.TypeSet:
+		if es, ok := f.Schema.Elem.(*schema.Schema); ok {
+			switch es.Type { //nolint:exhaustive
+			// We assume scalar types can be granular maps.
```
negz, line 176: Note to self. 😄 
 
```suggestion
			// We assume scalar types can be granular sets.
```

pkg/types/field.go
```diff
@@ -151,9 +152,38 @@ func NewField(g *Builder, cfg *config.Resource, r *resource, sch *schema.Schema,
...
 
+	AddServerSideApplyMarkers(f)
+
 	return f, nil
 }
 
+// AddServerSideApplyMarkers adds server-side apply comment markers to indicate
+// that scalar maps and sets can be merged granularly, not replace atomically.
+func AddServerSideApplyMarkers(f *Field) {
+	switch f.Schema.Type { //nolint:exhaustive
```
sergenyalcin, line 163: We may consider moving this switch-case to inside the `buildSchema` function. Because there is a similar switch-case block inside of this function.

pkg/types/markers/ssa.go
```diff
@@ -0,0 +1,87 @@
...
+	// ListTypeAtomic means the entire list is replaced during merge. At any
+	// point in time, a single manager owns the list.
+	ListTypeAtomic ListType = "atomic"
+
+	// ListTypeSet can be granularly merged, and different managers can own
+	// different elements in the list. The list can include only scalar
+	// elements.
+	ListTypeSet ListType = "set"
+
+	// ListTypeSet can be granularly merged, and different managers can own
```
sergenyalcin, line 23:
```suggestion
	// ListTypeMap can be granularly merged, and different managers can own
```

sergenyalcin (commented):
Thank you @negz for this PR. I left a few comments.

pkg/types/markers/ssa.go
```diff
@@ -0,0 +1,87 @@
...
+// upjet needs to control.
+// https://kubernetes.io/docs/reference/using-api/server-side-apply/#merge-strategy
+type ServerSideApplyOptions struct {
+	ListType   *ListType
+	ListMapKey []string
+	MapType    *MapType
+	StructType *StructType
+}
+
+func (o ServerSideApplyOptions) String() string {
```
sergenyalcin, line 67: It seems we need to change the `:`s to `=` while generating the markers.

https://github.com/kubernetes/enhancements/blob/master/keps/sig-api-machinery/555-server-side-apply/README.md#lists

pkg/types/field.go
```diff
@@ -151,9 +152,38 @@ func NewField(g *Builder, cfg *config.Resource, r *resource, sch *schema.Schema,
 	f.FieldType = fieldType
 	f.InitType = initType
 
+	AddServerSideApplyMarkers(f)
```
ulucinar, line 155: We need to conditionally add the kubebuilder topology markers for sets & maps if the field is not a sensitive field. For those fields marked as sensitive in their corresponding Terraform schema, we either generate a `*v1.SecretReference` or a `*[]v1.SecretKeySelector`. In this implementation, we will treat sensitive Terraform arguments with type `schema.TypeSet` as `set` lists (lists of scalar items) and sensitive arguments with type `schema.TypeMap` as `granular` maps. But in the first case, they are actually lists of objects (`*[]v1.SecretKeySelector`) and in the second case they are `*v1.SecretReference`, not maps. For sensitive map arguments, we may consider marking them as `granular` structs but I think a controller should update a `v1.SecretReference` atomically:
```suggestion
	if !sch.Sensitive {
		AddServerSideApplyMarkers(f)
	}
```

pkg/types/field.go
ulucinar: Hi @negz,
Here are some statistics regarding the object lists across the generated APIs of `upbound/provider-{aws,azure,gcp}`:
```
provider-aws   ->  map[embedded:1397 list-of-objects:165]
provider-azure -> map[embedded:1165 list-of-objects:471]
provider-gcp    -> map[embedded:1335 list-of-objects:291]
```
Here, `embedded` counts represent object lists with a `MaxItems` constraint of 1 whether that item is optional or not. The importance of this is that we can, in theory, convert these lists (of length 1) into embedded objects. A good example is the `spec.forProvider.vpcConfig` field of `Cluster.eks`, which is also the subject of https://github.com/upbound/provider-aws/issues/975. Due to HCL syntax, such embedded objects are also represented as lists, which is also reflected in our generated APIs. Converting such object lists of length 1 into required (`MinItems == 1`) or optional embedded objects will relieve us from having to specify the list map keys for the list `map` SSA merge strategy. But we still need to be careful about the possibility of the changing `MaxItems` constraints if the semantics of the API could potentially allow this.

The `list-of-objects` counts represent other object lists without a `MaxItems` constraint or a `MaxItems` constraint greater than 1. The question is then how we will determine the index set, i.e., the list map keys for those associative lists. We have also examined 10 object lists belonging to MRs of different API groups and 3-4 of them already had their indexing (naming) keys (such as a unique identifier for the list item) and 6-7 of them did not have a natural index set (no subset of the scalar object fields constitute an index set that can uniquely identify a list item). 

So my current understanding is that it's not possible to reliably determine the associative list object keys in our generated APIs. We will even need to inject some naming fields into the generated CRDs schemas to be able to specify an SSA merge strategy of type `map`, so that the API server can apply its SSA merge semantics for the objects in the list.

On the other hand, if we just inject a naming field that will not overlap with the already existing fields in the list object schema, then it will result in unnormalized APIs where there is already a set of naming fields (index set mentioned above). 

Another complexity we need to handle is any field that goes into the list map keys is not nullable, it has to be either a required field or should have a default value. For injected keys of objects of lists with a length constraint greater than 1 (or no length constraint), we cannot do the defaulting via the kubebuilder markers and we will have to resort to something like defaulting webhooks that will compute a proper unique identifier (index set) for a given list object (two index sets must be equal iff the corresponding objects are equal). So, to do this properly, I believe we will need to have a manual configuration API in upjet through which we can specify the logic to configure & compute an index set. 

My suggestion is to leave the object lists case unresolved for now and introduce a configuration API for this purpose in a follow-up PR so that we can address issues similar to https://github.com/upbound/provider-aws/issues/975 via SSA merge strategies.

ulucinar:
Hi @sergenyalcin,
My suggestion is that we fix the issues spotted in the reviews on this PR's feature branch and merge this PR, as @negz is out for a while and we would like to fix https://github.com/upbound/provider-aws/issues/975. How do you feel about it?

sergenyalcin:
> Hi @sergenyalcin, My suggestion is that we fix the issues spotted in the reviews on this PR's feature branch and merge this PR, as @negz is out for a while and we would like to fix [upbound/provider-aws#975](https://github.com/upbound/provider-aws/issues/975). How do you feel about it?

This plan is perfectly fine with me. As you mentioned, we want to fix the relevant issue (related to AWS eks.Cluster); also, fixing and merging this PR is a chance to include the `server-side apply` to Upjet-based providers.

sergenyalcin (approved):
Thanks @ulucinar LGTM!

ulucinar (approved):

PR #309, opened 2023-12-05
Put default and configured timeouts to InstanceState for using the timeouts while Observe calls

sergenyalcin:
This PR puts default and configured timeouts to `InstanceState` for using the timeouts while `Observe` calls. We had already passed the timeouts to InstanceDiff. This covers the `Create`, `Update`, and `Delete` calls. However, we also need to put the timeouts to `InstanceState` for' Observe' calls. 

We caught the issue while working on the provider-gcp storage.Bucket resource. There is a retry mechanism for Observe calls in the terraform provider for the following reason: 

> There seems to be some eventual consistency errors in some cases, so we want to check a few times to ensure it exists before moving on.

Also, if we do not set timeouts to resource, the TF provider sets the timeouts to 20m by default. When a user tries to create a new bucket, this operation is stuck for 20m on a retry loop in the Observation state and does not return anything. So, this blocks the creation of this resource.

It seems we have already observed this issue before and set a timeout [here](https://github.com/upbound/provider-gcp/blob/95302ba820a66c14b44dc3c887f03c706b1b4612/config/storage/config.go#L27) for the storage.Bucket resource. To activate this overridden timeout, we must pass these values to the InstanceState, as I mentioned above.

Fixes #

I have:

Tested in provider-gcp by using the storage.Bucket resource manually.


sergenyalcin:
@ulucinar As we discussed offline, first, I put the timeouts to the `Connect` function. Then, I checked the situation on the Observe before setting the state. It seems we do not override the timeouts back. I mean, the timeouts remain as we set it. However, being on the safe side, we may consider adding the timeouts again in the `Observe` function.

ulucinar (approved):
Thanks @sergenyalcin, lgtm.

PR #311, opened 2023-12-12
Call the registered schema.CustomizeDiffFunc functions in the Terraform SDK-based external client

ulucinar:

This PR starts passing the registered `schema.CustomizeDiffFunc`s to the `schema.schemaMap.Diff` function so that the available `CustomizeDiff` functions in the underlying Terraform provider are utilized. We had previously considered using these custom diff functions but we could not do so previously.

The strategy is now to start supplying the custom diff functions from the native schema by default but for the providers that we need to validate these registered functions are working as intended, we will just remove, and thus prevent, the custom diff functions from the schema.

I have:

Tested in https://github.com/upbound/provider-gcp/pull/424 with the BucketACL.storage` resource.


sergenyalcin (approved):
Thanks @ulucinar LGTM!

PR #312, opened 2023-12-12
Ignore specific error that returned by expandWildcard function

sergenyalcin:

Related to https://github.com/crossplane/crossplane-runtime/pull/627

The issue was observed while testing the GCP certificatemanager.Certificate resource. While trying to get the connection details of this resource, we try to read the self_managed[*].certificate_pem attribute of this resource. If the self_managed object is nil, then we observe an error:

cannot get connection details: cannot get connection details: cannot expand wildcards: cannot expand wildcards for segments: "self_managed[*].certificate_pem": "self_managed": unexpected wildcard usage'

If the value of this field is nil, then we do not want to return an error because there is nothing to read in the sensitive attribute. So, after returning a specific error from the nil case, we are handling this in Upjet by ignoring this error type.

I have:

Tested against to the GCP certificatemanager.Certificate resource.

### DIFF
--- a/pkg/resource/sensitive.go
+++ b/pkg/resource/sensitive.go
@@ -107,6 +107,9 @@ func GetSensitiveAttributes(from map[string]any, mapping map[string]string) (map
 	for tf := range mapping {
 		fieldPaths, err := paved.ExpandWildcards(tf)
 		if err != nil {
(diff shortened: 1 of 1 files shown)


ulucinar (approved):
Thanks @sergenyalcin, lgtm.

PR #315, opened 2023-12-21
Add reference fields to the InitProvider

### LINKED ISSUE #307: Consider adding reference fields to initProvider

Inspired by https://github.com/upbound/provider-aws/issues/994. 

The `initProvider` fields mirror `forProvider` fields except for Identifiers and References which are exclusive for `forProvider`.

For Identifiers it was decided to be only in `forProvider` because they are fields that should be immutable during the resource lifecycle, so no point to have them in `initProvider`.

For References, it was more of a practical decision due to the reconciler flow in upjet based providers. From the `IsInit()` function comment:
```
// Because of how upjet works, the main.tf file is created and filled
// in the Connect step of the reconciliation. So we merge the initProvider
// and forProvider there and write it to the main.tf file. So fields that are
// not part of terraform are not included in this merge, plus they cant be
// ignored through ignore_changes. References similarly get resolved in
// an earlier step, so they cannot be included as well.
```

As we now are moving towards no-fork Upjet, that doesnt have the TF file, we could reconsider the decision about references as now it may be feasible to do it.

sergenyalcin:

Fixes #307 

This PR supports generation of reference fields and reference resolver functions to the InitProvider.

I have:

This PR tested against to the upjet-based provider-aws

### DIFF
--- a/pkg/types/builder.go
+++ b/pkg/types/builder.go
@@ -274,7 +274,7 @@ func (g *Builder) buildSchema(f *Field, cfg *config.Resource, names []string, r 
 				if paramType.Underlying().String() != emptyStruct {
 					field := types.NewField(token.NoPos, g.Package, f.Name.Camel, types.NewSlice(paramType), false)
 					r.addParameterField(f, field)
-					r.addInitField(f, field, g.Package)
+					r.addInitField(f, field, g, nil)
 				}
 			default:
 				if paramType == nil {
@@ -400,7 +400,7 @@ func (r *resource) addParameterField(f *Field, field *types.Var) {
 	r.paramFields = append(r.paramFields, field)
 }
 
-func (r *resource) addInitField(f *Field, field *types.Var, pkg *types.Package) {
+func (r *resource) addInitField(f *Field, field *types.Var, g *Builder, typeNames *types.TypeName) {
 	// If the field is not an init field, we don't add it.
 	if !f.isInit() {
 		return
@@ -410,10 +410,14 @@ func (r *resource) addInitField(f *Field, field *types.Var, pkg *types.Package) 
 
 	// If the field is a nested type, we need to add it as the init type.
 	if f.InitType != nil {
-		field = types.NewField(token.NoPos, pkg, f.Name.Camel, f.InitType, false)
+		field = types.NewField(token.NoPos, g.Package, f.Name.Camel, f.InitType, false)
 	}
 
 	r.initFields = append(r.initFields, field)
+
+	if f.Reference != nil {
+		r.addReferenceFields(g, typeNames, f, true)
+	}
 }
 
 func (r *resource) addObservationField(f *Field, field *types.Var) {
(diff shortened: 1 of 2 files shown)


pkg/types/field.go
```diff
@@ -401,7 +401,7 @@ func (f *Field) AddToResource(g *Builder, r *resource, typeNames *TypeNames, add
 // an earlier step, so they cannot be included as well. Plus probably they
 // should also not change for Create and Update steps.
 func (f *Field) isInit() bool {
-	return !f.Identifier && f.Reference == nil && (f.TFTag != "-" || f.Injected)
```
ulucinar, line 404: Why do we delete `f.Injected` condition here? For injected fields (for defining associative lists in the context of SSA), we need the TF tag to be `-` so that they are ignored during TF JSON serialization but we still need them to be available in the CRD API. So, I suspect, we need to preserve the condition of `f.Injected` here.

pkg/types/field.go
```diff
@@ -401,7 +401,7 @@ func (f *Field) AddToResource(g *Builder, r *resource, typeNames *TypeNames, add
...
+	return !f.Identifier && f.TFTag != "-"
```
ulucinar, line 404: Please see the comment above. Does the following work?
```suggestion
	return !f.Identifier && (f.TFTag != "-" || f.Injected)
```

pkg/types/builder.go
```diff
@@ -429,10 +433,15 @@ func (r *resource) addObservationField(f *Field, field *types.Var) {
...
 	refFields, refTags := g.generateReferenceFields(paramName, field)
-	r.paramTags = append(r.paramTags, refTags...)
-	r.paramFields = append(r.paramFields, refFields...)
+	if isInit {
+		r.initTags = append(r.initTags, refTags...)
+		r.initFields = append(r.initFields, refFields...)
+	} else {
+		r.paramTags = append(r.paramTags, refTags...)
+		r.paramFields = append(r.paramFields, refFields...)
+	}
```
ulucinar, line 444: I think the reference fields will now always be added to the `spec.forProvider` and `spec.initProvider`, is this correct? If we don't have the branching here, what would cause an issue is the fact that this function is called once for the `forProvider` and once for `initProvider`, is this correct?

pkg/types/field.go
sergenyalcin: Thanks @ulucinar for catching this. After rebasing, I tried to solve the conflicts and it seems this removal is not an intentional change. I put this again.

pkg/types/field.go
sergenyalcin: Thanks @ulucinar for catching this. After rebasing, I tried to solve the conflicts and it seems this removal is not an intentional change. I put this again.
Done!

pkg/types/builder.go
sergenyalcin: Yes the both are correct. Now we always generate the reference fields for both. And if we do not have a branch here, this will cause an issue.

ulucinar (approved):
Thanks @sergenyalcin, lgtm.

PR #316, opened 2023-12-21
Pass full state to GetExternalNameFn function to access field other than ID

sergenyalcin:

This PR passes full state to GetExternalNameFn function to access field other than ID.

I have:

Tested against keyVault.AccessPolicy resource in provider-azure

### DIFF
--- a/pkg/controller/external_nofork.go
+++ b/pkg/controller/external_nofork.go
@@ -545,12 +545,12 @@ func (n *noForkExternal) Observe(ctx context.Context, mg xpresource.Managed) (ma
 // if the external-name of the MR has changed.
 func (n *noForkExternal) setExternalName(mg xpresource.Managed, stateValueMap map[string]interface{}) (bool, error) {
 	id, ok := stateValueMap["id"]
-	if !ok {
+	if !ok || id.(string) == "" {
 		return false, nil
 	}
 	newName, err := n.config.ExternalName.GetExternalNameFn(stateValueMap)
 	if err != nil {
-		return false, errors.Wrapf(err, "failed to get the external-name from ID: %s", id)
+		return false, errors.Wrapf(err, "failed to compute the external-name from the state map of the resource with the ID %s", id)
 	}
 	oldName := meta.GetExternalName(mg)
 	// we have to make sure the newly set external-name is recorded
@@ -602,11 +602,11 @@ func (n *noForkExternal) Create(ctx context.Context, mg xpresource.Managed) (man
 	n.opTracker.SetTfState(newState)
 
 	stateValueMap, err := n.fromInstanceStateToJSONMap(newState)
+	if err != nil {
+		return managed.ExternalCreation{}, errors.Wrap(err, "failed to convert instance state to map")
+	}
 	if _, err := n.setExternalName(mg, stateValueMap); err != nil {
 		return managed.ExternalCreation{}, errors.Wrapf(err, "failed to set the external-name of the managed resource during create")
-	}
-	if err != nil {
-		return managed.ExternalCreation{}, err
 	}
 	err = mg.(resource.Terraformed).SetObservation(stateValueMap)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


pkg/controller/external_nofork.go
```diff
@@ -543,15 +543,14 @@ func (n *noForkExternal) Observe(ctx context.Context, mg xpresource.Managed) (ma
 
 // sets the external-name on the MR. Returns `true`
 // if the external-name of the MR has changed.
-func (n *noForkExternal) setExternalName(mg xpresource.Managed, newState *tf.InstanceState) (bool, error) {
-	if newState.ID == "" {
+func (n *noForkExternal) setExternalName(mg xpresource.Managed, stateValueMap map[string]interface{}) (bool, error) {
+	id, ok := stateValueMap["id"]
+	if !ok {
```
ulucinar, line 548: To be precise, we also need to make sure that the Terraform ID is not an empty string, i.e., it may be in the map but its value could be the empty string:
```suggestion
	if !ok || id.(string) == "" {
```
, with the assumption that if the `id` key exists, its value is a string.

pkg/controller/external_nofork.go
```diff
@@ -543,15 +543,14 @@ func (n *noForkExternal) Observe(ctx context.Context, mg xpresource.Managed) (ma
...
 		return false, nil
 	}
-	newName, err := n.config.ExternalName.GetExternalNameFn(map[string]any{
-		"id": newState.ID,
-	})
+	newName, err := n.config.ExternalName.GetExternalNameFn(stateValueMap)
 	if err != nil {
-		return false, errors.Wrapf(err, "failed to get the external-name from ID: %s", newState.ID)
+		return false, errors.Wrapf(err, "failed to get the external-name from ID: %s", id)
```
ulucinar, line 553: nit:
```suggestion
		return false, errors.Wrapf(err, "failed to compute the external-name from the state map of the resource with the ID %s", id)
```

pkg/controller/external_nofork.go
```diff
@@ -602,10 +601,10 @@ func (n *noForkExternal) Create(ctx context.Context, mg xpresource.Managed) (man
 	}
 	n.opTracker.SetTfState(newState)
 
-	if _, err := n.setExternalName(mg, newState); err != nil {
+	stateValueMap, err := n.fromInstanceStateToJSONMap(newState)
+	if _, err := n.setExternalName(mg, stateValueMap); err != nil {
```
ulucinar, line 605: Let's make sure we check any errors returned from `n.fromInstanceStateToJSONMap`:
```suggestion
	stateValueMap, err := n.fromInstanceStateToJSONMap(newState)
	if err != nil {
		return managed.ExternalCreation{}, errors.Wrap(err, "failed to create the resource")
	}
	if _, err := n.setExternalName(mg, stateValueMap); err != nil {
```

pkg/controller/external_nofork.go
```diff
@@ -602,10 +601,10 @@ func (n *noForkExternal) Create(ctx context.Context, mg xpresource.Managed) (man
...
 		return managed.ExternalCreation{}, errors.Wrapf(err, "failed to set the external-name of the managed resource during create")
 	}
-	stateValueMap, err := n.fromInstanceStateToJSONMap(newState)
 	if err != nil {
 		return managed.ExternalCreation{}, err
 	}
```
ulucinar, line 610: This block would be moved:
```suggestion
```

pkg/controller/external_nofork.go
sergenyalcin: Done!

pkg/controller/external_nofork.go
sergenyalcin: Done!

pkg/controller/external_nofork.go
sergenyalcin: Done!

pkg/controller/external_nofork.go
sergenyalcin: Done!

ulucinar (approved):
Thank you @sergenyalcin, lgtm.

PR #320, opened 2023-12-27
Bump hashicorp/terraform-plugin-sdk to v2.30.0

ulucinar:

This PR bumps the Terraform plugin SDK version to `v2.30.0` and if the `schema.Resource.Schema` is `nil`, populates the resource's native schema by calling the `schema.Resource.SchemaFunc`. 

The PR does not attempt to optimize the memory usage due to large resource schemas with the lazy initialization of such schemas. This could be a future optimization point for us.

Another point is that we may need to branch out the versions we consume in the providers if we figure out issues when we bump the upjet versions in those providers because the bumped TF SDK version may not be compatible with them.

I have:

Tested against https://github.com/upbound/provider-aws/pull/1037 using the `aws_kinesis_firehose_delivery_stream` resource that @turkenf reported.


pkg/config/provider.go
```diff
@@ -286,11 +286,13 @@ func NewProvider(schema []byte, prefix string, modulePath string, metadata []byt
 					"but either config.Provider.TerraformProvider is not configured or the Go schema does not exist for the resource", name))
 			}
 			terraformResource = p.TerraformProvider.ResourcesMap[name]
-			// TODO: we will need to bump the terraform-plugin-sdk dependency to handle
-			// schema.Resource.SchemaFunc
 			if terraformResource.Schema == nil {
-				p.skippedResourceNames = append(p.skippedResourceNames, name)
-				continue
+				if terraformResource.SchemaFunc == nil {
```
ulucinar, line 290: We may even consider panicing on these conditions.

sergenyalcin (approved):
Thanks @ulucinar LGTM!

PR #331, opened 2024-01-19
Add a transformer to remove API group imports for cross-resource references

### LINKED ISSUE #96: Explicit reference causes import cycle error
### What happened?

With one specific GCP resource (`google_service_networking_peered_dns_domain `) configured with an explicit `Network` reference, running `make reviewable` results in an [error](https://github.com/upbound/official-providers/runs/7951516684?check_suite_focus=true#step:9:30) as described in this https://github.com/upbound/official-providers/pull/518#discussion_r943598464. Note that the `Network` reference is not automatically added via Upjet either.

https://github.com/upbound/official-providers/blob/7ac76a2a7d155838a7cc479b6dd8352357b1d79c/provider-gcp/config/cloudplatform/config.go#L93-L101

[Removing](https://github.com/upbound/official-providers/pull/518/commits/12dc7be67b7bb4b7d956d4121f36619ea7fbcae5#diff-7605d33a7111a5a72caa05b139361688d16236f1da6a0bc24a1e9c6943e8b6acL94-L97) the explicit reference allows for successful generation.

### How can we reproduce it?

* Checkout commit 93afafc9361dfdf96bd2aa91d20cd2bdc29bebf3
* `make && cd provider-gcp && make reviewable`

### What environment did it happen in?

* Crossplane Version: `v1.9.0-up.2`
* Provider Name: `provider-gcp`
* Provider Version: development - see https://github.com/upbound/official-providers/pull/518

ulucinar:

Fixes #96

While working in the context of the multi-version CRD & conversion function generation pipelines and #321 & https://github.com/upbound/provider-aws/pull/1075, we observed that selectively moving individual resources to different API versions increases the likelihood of #96 because cross-resource references are more common inside an API group. We observed multiple cases of #96 in https://github.com/upbound/provider-aws/pull/1075 when we bumped the API versions of MRs with breaking API changes, such as the `RoutingProfile.connect` resource. 

This PR employs a dynamic type registry in the calls to the [APIResolver](https://github.com/crossplane/crossplane-runtime/blob/ed7658c003b3780596247f71c19a286f7e7d951a/pkg/reference/reference.go#L304) so that we don't need to import the modules for the individual GVKs in the resolver modules, which effectively eliminates the issue reported in #96. 

While the approach taken here can be applied to the other providers in the ecosystem by implementing the logic in https://github.com/crossplane/crossplane-tools, we have decided to initially implement it in upjet as a transformer (as opposed to being a full-fledged generator). The crossplane-tools' `angryjet` is invoked first to generate the cross-resource reference resolvers (with cross API-group import statements) and then the transformer introduced with this change (`resolver`) can be applied as part of the build pipeline to get rid of the cross API-group import statements, effectively addressing #96 for upjet-based providers.

An example transformation obtained by running `resolver` is as follows:
```go
// ResolveReferences of this Subnet.
func (mg *Subnet) ResolveReferences(ctx context.Context, c client.Reader) error {
	r := reference.NewAPIResolver(c, mg)

	var rsp reference.ResolutionResponse
	var err error

	rsp, err = r.Resolve(ctx, reference.ResolutionRequest{
		CurrentValue: reference.FromPtrValue(mg.Spec.ForProvider.VPCID),
		Extract:      reference.ExternalName(),
		Reference:    mg.Spec.ForProvider.VPCIDRef,
		Selector:     mg.Spec.ForProvider.VPCIDSelector,
		To: reference.To{
			List:    &VPCList{},
			Managed: &VPC{},
		},
	})
	if err != nil {
		return errors.Wrap(err, "mg.Spec.ForProvider.VPCID")
	}
	mg.Spec.ForProvider.VPCID = reference.ToPtrValue(rsp.ResolvedValue)
	mg.Spec.ForProvider.VPCIDRef = rsp.ResolvedReference

```
is transformed into:
```go
// ResolveReferences of this Subnet.
func (mg *Subnet) ResolveReferences(ctx context.Context, c client.Reader) error {
	var m xpresource.Managed
	var l xpresource.ManagedList
	r := reference.NewAPIResolver(c, mg)

	var rsp reference.ResolutionResponse
	var err error
	{
		m, l, err = apisresolver.GetManagedResource("ec2.aws.upbound.io",
			"v1beta1",
			"VPC", "VPCList",
		)
		if err != nil {
			return errors.Wrap(err, "failed to get the reference target managed resource and its list for reference resolution")
		}

		rsp, err = r.Resolve(ctx, reference.ResolutionRequest{
			CurrentValue: reference.FromPtrValue(mg.Spec.ForProvider.VPCID),
			Extract:      reference.ExternalName(),
			Reference:    mg.Spec.ForProvider.VPCIDRef,
			Selector:     mg.Spec.ForProvider.VPCIDSelector,
			To:           reference.To{List: l, Managed: m},
		})
	}
	if err != nil {
		return errors.Wrap(err, "mg.Spec.ForProvider.VPCID")
	}
	mg.Spec.ForProvider.VPCID = reference.ToPtrValue(rsp.ResolvedValue)
	mg.Spec.ForProvider.VPCIDRef = rsp.ResolvedReference

```

TODO:
- [x] Implement unit tests

I have:

All the generated resolver modules for `upbound/provider-aws` have been transformed using `resolver` in https://github.com/upbound/provider-aws/pull/1075. We have also tried moving the `RoutingProfile.connect` resource to version `v1beta2`, leaving all the other resources in the same API group at version `v1beta1`, made sure the import cycle exists between `connect/v1beta1/zz_generated.resolvers.go` & `connect/v1beta2/zz_generated.resolvers.go`, and applied to the transformer to remove the import cycle, and then successfully provision a `RoutingProfile.connect` instance using [this manifest](https://github.com/ulucinar/upbound-provider-aws/blob/e064009b33f5367a6fabc909710efe40e3fdc730/examples/connect/v1beta1/routingprofile.yaml).


pkg/transformers/resolver.go
```diff
@@ -0,0 +1,595 @@
+// SPDX-FileCopyrightText: 2024 The Crossplane Authors <https://crossplane.io>
```
sergenyalcin, line 1: In my local tests, I observe that the place of the first resolver function's code comment changes. This is not a very important issue. However, I thought this place could be reviewed in case it is re-visited.

pkg/transformers/resolver.go
```diff
@@ -0,0 +1,595 @@
...
+												used: false,
+											}
+											// we will parse the import path such as:
+											// github.com/upbound/provider-aws/apis/ec2/v1beta1
+											// and extract the GV information from it.
+											tokens := strings.Split(path, "/")
+											// e.g., v1beta1
+											version = tokens[len(tokens)-1]
+											// e.g., ec2.aws.upbound.io
+											group = fmt.Sprintf("%s.%s", tokens[len(tokens)-2], apiGroupSuffix)
```
sergenyalcin, line 326: Will this statement cause any issues with the Azure-ResourceGroup resource? Because the API group of ResourceGroup is the same as the `apiGroupSuffix.: `azure.upbound.io`. 

I mean, the split statement will be this: `github.com/upbound/provider-azure/apis/azure/v1beta1`

Then, the calculated group will be `azure.azure.upbound.io`.

pkg/transformers/resolver.go
```diff
@@ -0,0 +1,595 @@
...
+										// source. We still transform it for uniformity and
+										// in the future, the source and target might still be
+										// moved to different packages.
+										// The GV information comes from file name in this case:
+										// apis/cur/v1beta1/zz_generated.resolvers.go
+										tokens := strings.Split(filePath, "/")
+										// e.g., v1beta1
+										version = tokens[len(tokens)-2]
+										// e.g., cur.aws.upbound.io
+										group = fmt.Sprintf("%s.%s", tokens[len(tokens)-3], apiGroupSuffix)
```
sergenyalcin, line 348: Is there the above problem here?

sergenyalcin (commented):
Thanks @ulucinar I left a few comments

cmd/resolver/main.go
```diff
@@ -0,0 +1,31 @@
...
+	"github.com/crossplane/upjet/pkg/transformers"
+)
+
+func main() {
+	var (
+		app                     = kingpin.New(filepath.Base(os.Args[0]), "Transformer for the generated resolvers by the crossplane-tools so that cross API-group imports are removed.").DefaultEnvars()
+		apiGroupSuffix          = app.Flag("apiGroupSuffix", "Resource API group suffix, such as aws.upbound.io. The resource API group names are suffixed with this to get the canonical API group name.").Short('g').Required().String()
+		pattern                 = app.Flag("pattern", "List patterns for the packages to process, such as ./...").Short('p').Default("./...").Strings()
+		resolverFilePattern     = app.Flag("resolver", "Name of the generated resolver files to process.").Short('r').Default("zz_generated.resolvers.go").String()
+		ignorePackageLoadErrors = app.Flag("ignoreLoadErrors", "Ignore errors encountered while loading the packages.").Short('s').Bool()
```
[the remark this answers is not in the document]
ulucinar: I feel like the default should be the more strict mode of operation (failing on package loading errors). Though I think I see your point: The purpose of this transformer is to prevent/get rid of circular imports, and any provider benefiting from this transformer should have an import cycle caused by the cross-API group references, so in order to run this transformer, they will need to supply the `-s` option. But still, I'd like the decision to ignore any package loading errors to be on the side of the Crossplane provider repo maintainer who's using this transformer as part of their code generation pipelines. So keeping the default mode strict forces the repo maintainer to show explicit consent while incorporating the transformer. 

The package loading issues we expect are the import cycle errors caused by cross-resource references but what if the API folder has some unexpected (by the transformer) compile errors? I believe it's better to give the responsibility of suppressing package loading errors to the repo maintainer, although from a practical perspective, I agree setting the default of this option to `true` will not make much difference.

pkg/transformers/resolver.go
ulucinar: Thanks for reporting this unwanted behavior. I've tackled with this issue a bit and found [this](https://github.com/golang/go/issues/20744) upstream issue regarding the floating comments when we modify the existing code. Let me open a separate issue to address this as I suspect we would need to properly calculate the position of that single comment in the transformer source code or use a higher level library like [this one](https://github.com/dave/dst) to fix this and currently, to the best of my knowledge, it does cause any functional issues.

pkg/transformers/resolver.go
ulucinar: Thanks @sergenyalcin, good catch indeed! Given this a try with `upbound/provider-azure` as you suggested and what we generate is:
```go
m, l, err = apisresolver.GetManagedResource("azure.azure.upbound.io", "v1beta1", "ResourceGroup", "ResourceGroupList")
		if err != nil {
			return errors.Wrap(err, "failed to get the reference target managed resource and its list for reference resolution")
		}
```

Added the `--api-group-override` command-line option (shorthand `-o`) so that any API group overrides can be configured. Also added the `--api-resolver-package` (shorthand `-a`) command-line option so that we can properly configure the API resolver function's (`GetManagedResource`) package.

So, when we run the transformer with the following command-line options:
```shell
go run github.com/crossplane/upjet/cmd/resolver -g azure.upbound.io -a github.com/upbound/provider-azure/internal/apis -o azure.azure.upbound.io=azure.upbound.io
```
what gets now generated is:
```go
		m, l, err = apisresolver.GetManagedResource("azure.upbound.io", "v1beta1", "ResourceGroup", "ResourceGroupList")
		if err != nil {
			return errors.Wrap(err, "failed to get the reference target managed resource and its list for reference resolution")
		}
```

In a future iteration (especially if some other providers also adopt this transformer), we may also generate the `GetManagedResource` function via upjet for the provider and remove the `--api-resolver-package` option.

pkg/transformers/resolver.go
ulucinar: Yes, we need to address this also. Please see the above comment explaining group overrides configuration and the test with `upbound/provider-azure`. Thank you @sergenyalcin.

sergenyalcin (approved):
Thanks @ulucinar LGTM!

PR #334, opened 2024-01-24
fix diff state not being set to freshly observed state for non-existing resources

erhancagirici:
Fixes diff state not being set to freshly observed (nil) state for non-existing resources.
This causes reconstructed prior state being used at diff that causes parsing issues for fields with validations.

Will fix https://github.com/upbound/provider-aws/issues/1071, after consumed at `provider-aws`

I have:

Consumed at a local provider-aws build and tested with manifests specified in https://github.com/upbound/provider-aws/issues/1071 and issue is not observed.


ulucinar (approved):
Thanks @erhancagirici, lgtm. Please also record the root cause analysis we've done for the `ReplicaGroup.elasticache` `engine_version` issue.

PR #338, opened 2024-02-01
Replace NoFork Terminology with TerraformPluginSDK

ulucinar:

The nofork architecture now has the native implementations for both the Terraform plugin SDK v2 resources and the Terraform plugin framework resources. The repo uses `NoFork` in the file & type names related to the plugin SDK v2 implementation. This PR aims to replace that terminology with the `TerraformPluginSDK` naming. 

This PR also adds a more explanatory error message when immutable fields of a managed resource have changed. We currently do not generate CRD validation rules for immutable fields and thus, if they are updated, we block it at the Terraform layer. The improve the error message for such cases as follows:
`refuse to update the external resource because the following update requires replacing it`

This PR also fixes a race condition on a global state in the Terraform plugin SDK v2 & Terraform plugin framework based external client *tests*. The global state is a `fake.Terraformed` object, which is shared between multiple  test cases. The theory is that changing the names of the test modules caused a reordering of the test executions, which revealed the race. Resolved by not sharing this state between test cases.

I have:

Tested via https://github.com/upbound/provider-aws/pull/1121.


sergenyalcin (approved):
Thanks @ulucinar LGTM!

PR #341, opened 2024-02-09
move RequiresReplace check of plan to Update time instead of Observe & suppress diffs with only computed attributes

erhancagirici:

Moves the RequiresReplace check of the Plan to start of the Update, instead of Observe.

Currently plans are checked if they return `RequiresReplace` at diff calculation time at `Observe` calls, and the plan is refused if they require resource replace with an error due to XRM compliance. 
However, this prevents late initialization of MR fields with RequiresReplace during initial creation, due to first observe after creation calculates a diff with `RequiresReplace` that gets rejected, since those fields are not late-initialized yet. Then, the diff planning exists early, not letting it reach to late initialization section.

Therefore the enforcement of the `RequiresReplace` plans are moved to Update time, to let fields late initialize first and enforce it after update decision is made.

Also, this change suppresses diffs with only computed attributes. The current diff detection mechanism compares the plan response and the current state. Plans can contain unknown values for Computed attributes, which is shown as a "false-positive" diff in the current diff logic.

This change suppreses diff elements that has unknown plan value. Note that it does NOT modify the plan itself in any way, just influences the logic of whether there is a diff or not.

I have:

Tested with new opensearchserverless resources in PR https://github.com/upbound/provider-aws/pull/1130 and all existing Terraform Plugin Framework resources

### DIFF
--- a/pkg/controller/external_tfpluginfw.go
+++ b/pkg/controller/external_tfpluginfw.go
@@ -99,7 +99,7 @@ type terraformPluginFrameworkExternalClient struct {
 	resource       fwresource.Resource
 	server         tfprotov5.ProviderServer
 	params         map[string]any
-	plannedState   *tfprotov5.DynamicValue
+	planResponse   *tfprotov5.PlanResourceChangeResponse
 	resourceSchema rschema.Schema
 	// the terraform value type associated with the resource schema
 	resourceValueTerraformType tftypes.Type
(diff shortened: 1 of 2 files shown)


pkg/controller/external_tfpluginfw.go
```diff
@@ -254,33 +254,34 @@ func (n *terraformPluginFrameworkExternalClient) getDiffPlan(ctx context.Context
...
 	}
 
-	return planResponse.PlannedState, len(diffso) > 0, nil
+	// filter diffs that has unknown plan value which corresponds to
+	// computed values. These cause unnecessary diff detection when only computed
+	// attribute diffs exist in the raw diff and no actual diff exists in the
+	// parametrizable attributes
+	filteredDiff := make([]tftypes.ValueDiff, 0)
+	for _, diff := range rawDiff {
+		if diff.Value1.IsKnown() {
```
sergenyalcin, line 279: For understanding better, as I understand, there are two `*Value` in `ValueDiff`: `Value1` and `Value2`. We use `Value1` here. Do we have an assumption here? Is it possible that we are missing something by not looking at Value2 at this point?

sergenyalcin (commented):
Thanks @erhancagirici left a comment

pkg/controller/external_tfpluginfw.go
erhancagirici: hi, we are comparing planned state with current state . Therefore in the resulting diff `Value1` corresponds to plan value and `Value2` is current state value (since invoked as `plannedStateValue.Diff(tfStateValue)` , plan is at LHS, current state at RHS).

Current state values are always known by nature, since all is read from cloud API. Plan can contain unknown values, since there can be computed values assigned by the cloud API, which can only be known after applying them. 

For such attributes we are not interested in their diffs when they are the only ones, i.e the calculated diff only consists of computed attributes (observable fields) with unknown values and no parametrizable field has changed.

sergenyalcin (approved):
Thanks @erhancagirici LGTM!

PR #358, opened 2024-03-01
Removing the applying of StateFuncs to parameters

sergenyalcin:
Relevant PR: https://github.com/upbound/provider-aws/pull/1188

This PR removes the applying of `StateFunc`s to parameters. This PR contains a fix for the https://github.com/upbound/provider-aws/issues/1109.

In the Diff function of [terraform-plugin-sdk](https://github.com/hashicorp/terraform-plugin-sdk/blob/d924635d4ed38beb1cc8295e9e9769a7fe2c6be8/helper/schema/schema.go#L1547), the StateFuncs have already applied to parameters. So, when we apply the StateFuncs before the `Observe`, the StateFuncs is called two times.

Big portion of the StateFuncs are idempotents like `NomalizeJson`, `strings.ToUpper` etc. However, some of them are not idempotent and we observed the problem in this type of case.

During the investigations of https://github.com/upbound/provider-aws/issues/1109 issue, we observed that a problematic behavior in upjet side. The `aws.ec2.Instance` resource's field `user_data` has a `StateFunc` that calculates the hash of the provided string and stores this hash value of string in status. For this resource the StateFunc is not idempotent and this causes a problem because we call the StateFunc two times. Result of this behavior, we calculate the wrong input for the resoruce.

This PR proposes removal of applying the StateFuncs to parameters. We will continue to process HCL expressions in parameters.

I have:

Tested against the following resources that have different StateFuncs (in AWS):

- ec2.Instance
- networkmanager.CoreNetwork
- dynamodb.Table

### DIFF
--- a/pkg/controller/external_tfpluginsdk.go
+++ b/pkg/controller/external_tfpluginsdk.go
@@ -158,13 +158,13 @@ func getExtendedParameters(ctx context.Context, tr resource.Terraformed, externa
 	return params, nil
 }
 
-func (c *TerraformPluginSDKConnector) processParamsWithStateFunc(schemaMap map[string]*schema.Schema, params map[string]any) map[string]any {
+func (c *TerraformPluginSDKConnector) processParamsWithHCLParser(schemaMap map[string]*schema.Schema, params map[string]any) map[string]any {
 	if params == nil {
 		return params
 	}
 	for key, param := range params {
 		if sc, ok := schemaMap[key]; ok {
-			params[key] = c.applyStateFuncToParam(sc, param)
+			params[key] = c.applyHCLParserToParam(sc, param)
 		} else {
 			params[key] = param
 		}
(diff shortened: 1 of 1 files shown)


pkg/controller/external_tfpluginsdk.go
```diff
@@ -216,16 +216,8 @@ func (c *TerraformPluginSDKConnector) applyStateFuncToParam(sc *schema.Schema, p
...
-	case schema.TypeBool, schema.TypeInt, schema.TypeFloat:
-		if sc.StateFunc != nil {
-			return sc.StateFunc(param)
-		}
 		return param
-	case schema.TypeInvalid:
+	case schema.TypeBool, schema.TypeInt, schema.TypeFloat, schema.TypeInvalid:
 		return param
 	default:
 		return param
```
ulucinar, line 223: We can merge this case with the `default` case:
```suggestion
	default:
		return param
```

ulucinar (approved):
Thanks @sergenyalcin, lgtm. If we do not observe any issues with the proposed changes here in our tests in https://github.com/upbound/provider-aws/pull/1188, then we can merge this PR.

pkg/controller/external_tfpluginsdk.go
sergenyalcin: Done!

PR #360, opened 2024-03-03
Swap SYNCED and READY columns in output

### LINKED ISSUE #359: Generated resources have "READY SYNCED" additional columns instead of "SYNCED READY"
### What happened?
When Crossplane generates CRDs for Claims and Composites it defines the additional printer columns "SYNCED" and "READY" in that order, which is logically read from left to right in the output.  A resource cannot be READY before it is SYNCED, so it is easier to read the resource status in that order.

upjet generates CRDs for resources with the columns "READY  SYNCED" which is inconsistent with the resources that Crossplane generates.

### How can we reproduce it?
Observe the `kubectl get` output for Claims/Composites and generated upjet resources

bobh66:
Crossplane creates CRDs for Claims and Composites with the additionalPrinterColumns "SYNCED" and "READY" in that order.

This PR modifies the code generation to use the same order of SYNCED and READY instead of the current READY and SYNCED.

Fixes #359 

I have:

- [X] Read and followed Upjet's [contribution process].
- [X] Run `make reviewable` to ensure this PR is ready for review.
- [X] Added `backport release-x.y` labels to auto-backport this PR if necessary.

Generated the Upbound provider-aws using these changes and confirmed that the CRDs were generated with the SYNCED and READY fields in that order

### DIFF
--- a/pkg/pipeline/templates/crd_types.go.tmpl
+++ b/pkg/pipeline/templates/crd_types.go.tmpl
@@ -45,8 +45,8 @@ type {{ .CRD.Kind }}Status struct {
 // +kubebuilder:storageversion
 
 // {{ .CRD.Kind }} is the Schema for the {{ .CRD.Kind }}s API. {{ .CRD.Description }}
-// +kubebuilder:printcolumn:name="READY",type="string",JSONPath=".status.conditions[?(@.type=='Ready')].status"
 // +kubebuilder:printcolumn:name="SYNCED",type="string",JSONPath=".status.conditions[?(@.type=='Synced')].status"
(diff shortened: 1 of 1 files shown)


ulucinar (approved):
Thanks @bobh66 for the change. There's a slight possibility that there might be shell scripts that depend on the current order of the `READY` and `SYNCED` columns but I think it's better for us to align with the core Crossplane resources and I would not consider such scripts as following the best practices. I will mark the PR as breaking, treating the `kubectl` output as part of our contract and it should be fine if we mention the change in the provider release notes.

Issue #136, opened 2022-11-07
One-element arrays could be considered as objects


muvaf:
### What problem are you facing?

There are cases where a field is actually a struct in the API but Terraform has it as single-element array, like here https://github.com/hashicorp/terraform-provider-aws/blob/7ff39c5b11aafe812e3a4b414aa6d345286b95ec/internal/service/efs/access_point.go#L52

### How could Upjet help solve your problem?

We can generate the arrays with `maxLength: 1` as structs so that we don't run into much [patching problems](https://github.com/crossplane/crossplane/issues/3335) and also the structural difference between the cloud API and CRD would be reduced. Though this would mean a breaking change, hence new `apiVersion` for many resources.

I think we should assess how reliable `maxLength: 1` condition is to figure out whether it should be struct or not. If it's close to 100%, it may be worth to bear the burden.

shanproofpoint:
imho, I think the provider should hide the terraform limitations from the idealized API.
having it leak through to the upbound gcp public API defeats abstraction and is the antithesis of crossplane..
perhaps the provider can do some transformation before giving it to terraform.

mbbush:
If we find cases where the `maxLength: 1` condition is not 100% reliable, it would also be nice to have a customization in `config` for enabling or disabling this, along the lines of `config.MoveToStatus`. 

As much as I am frustrated by the need to have all of these single element arrays, (especially when patching to `spec.forProvider.foo[0].bar[0].baz.other[0]`), it is nice to be able to have a consistent set of rules that I can apply in my head when reading terraform and translating it to crossplane.

Issue #10, opened 2022-06-09
Add "+kubebuilder:validation:MaxItems" marker for validating embedded structs in CRD schemas


ulucinar:
### What problem are you facing?

While working on #4, we have realized that Terraform represents embedded blocks always with Sets/Lists of length of most 1 (if optional, length is 0 or 1, if required, it will be always of length 1). An example is the [policy_settings](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/web_application_firewall_policy#policy_settings) block in `azurerm_web_application_firewall_policy`. In fact, this is an optional configuration block nested in `azurerm_web_application_firewall_policy` block. We would normally want to represents it with a CRD schema as follows:

```go
type WebApplicationFirewallPolicyParameters struct {
	// +kubebuilder:validation:Optional
	PolicySettings *PolicySettingsParameters `json:"policySettings,omitempty" tf:"policy_settings,omitempty"`
        ...
```

However, it looks like (probably due to HCL semantics), Terraform provider instead [represents](https://github.com/hashicorp/terraform-provider-azurerm/blob/3391b7fcae5bf6d36f0e8b65e45151089ca407b9/internal/services/network/web_application_firewall_policy_resource.go#L244) this as a list of length at most 1:
```go
			"policy_settings": {
				Type:     pluginsdk.TypeList,
				Optional: true,
				MaxItems: 1,
```

Upjet CRD generation pipeline also [does not allow](https://github.com/upbound/upjet/blob/39f56961709863bc3b739f99577e038fbb10566a/pkg/types/builder.go#L134) directly embedding a `pluginsdk.Resource` directly under another `pluginsdk.Resource` as expected.

Thus, Terrajet currently generates the following instead of the above fragment:

```go
type WebApplicationFirewallPolicyParameters struct {
	// +kubebuilder:validation:Optional
	PolicySettings []PolicySettingsParameters `json:"policySettings,omitempty" tf:"policy_settings,omitempty"`
        ...
```

Since this is a generated API that reflects the underlying Terraform resource API, I believe this is okay. But we have some room for improvement for such nested blocks (structs) by asserting a max length of 1 with kubebuilder markers.

### How could Terrajet help solve your problem?

We can use the Terraform resource schema to observe the `pluginsdk.Schema.MaxItems` to place the appropriate kubebuilder marker on the generated field. This will provide a better UX around the generated Crossplane provider API. The resulting generated code could be something like the following for the above field:
```go
[...]

PR #380, opened 2024-03-15
Use granular merge strategy for singleton lists

JonathanO:

Many TF providers nest blocks using a ListType with MaxItems=1, which upjet converts into a list containing a map. The default server side apply merge strategy for lists is atomic, resulting in conflicts between the provider and composition functions. This can be fixed on a case-by-case basis in the providers (https://github.com/crossplane-contrib/provider-upjet-gcp/pull/457), but it seems likely that all of these should be treated as if they were maps (which default to granular merge.)

While it would be great to fix this by flattening the unnecessary array, that would force an API change on all the providers.

This PR instead adjusts the ServerSideApplyMergeStrategies in the config to use ListTypeMap with a defaulted key for a singleton list. It works by altering the config, rather than during creation of Fields, to avoid duplicating the implementation used for ServerSideApplyMergeStrategies.

It also moves the existing server side apply marker defaulting for scalar sets and maps to work in the same way for consistency. I'm not sure that it's necessary to apply the granular strategy in the TypeMap case, as I think that's the k8s default, but I've left it in place to avoid additional changes to the output.

I have:

Generated updated GCP provider, which was run in Kind.

### DIFF
--- a/pkg/config/provider.go
+++ b/pkg/config/provider.go
@@ -351,6 +351,7 @@ func NewProvider(schema []byte, prefix string, modulePath string, metadata []byt
 		p.Resources[name] = DefaultResource(name, terraformResource, terraformPluginFrameworkResource, providerMetadata.Resources[name], p.DefaultResourceOptions...)
 		p.Resources[name].useTerraformPluginSDKClient = isTerraformPluginSDK
 		p.Resources[name].useTerraformPluginFrameworkClient = isPluginFrameworkResource
+		p.injectServerSideApplyDefaults(p.Resources[name], terraformResource, "")
 	}
 	for i, refInjector := range p.refInjectors {
 		if err := refInjector.InjectReferences(p.Resources); err != nil {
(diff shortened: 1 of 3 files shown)


ulucinar:
Hi @JonathanO,
Thank you for bringing this up. We had previously considered the idea of generating, by the default, an SSA merge strategy of `map` for singleton lists here:
- https://github.com/crossplane/upjet/pull/301: Please see this [comment](https://github.com/crossplane/upjet/pull/301#discussion_r1395061790) and my [reply](https://github.com/crossplane/upjet/pull/301#discussion_r1414370839) to it.
- https://github.com/crossplane/crossplane/issues/3335: I had originally classified this issue as an enhancement and not a bug. My main motivation in doing so was that for certain associative lists, it makes more sense to update the list elements atomically, and for some others, granularly. This depends on the relationships between the list object's fields and I think this is one of the reasons, in the context of SSA, we have a merge strategy of `atomic` in addition to the `map` strategy. 

My concerns with defaulting an SSA merge strategy of `map` for all object lists with the `MaxItems = 1` constraint are as follows:
- As you've also mentioned in the PR description, we would like to convert singleton object lists into embedded objects. A relevant issue is [here](https://github.com/crossplane/upjet/issues/136). We've (relatively) recently introduced support for multi-version CRDs and reusable chained API conversion functions in upjet, which will allow us to hide the breaking changes when we convert singleton lists into embedded objects (and there are some other challenges in implementing this). Converting singleton lists into embedded objects is currently on our roadmap. A relevant roadmap item is generating the CRD validation rules that will enforce the cardinality constraints on the object lists including the singleton ones.
- There might already exist an "index set" (i.e., object fields that constitute the list map keys). Adding an artificial index field when there already exists a natural index set (possibly with a cardinality greater than 1) is a normalization issue for singleton object lists. We had done some [observations](https://github.com/crossplane/upjet/pull/301#discussion_r1414370839) across the providers `crossplane-contrib/provider-upjet-{aws,azure,gcp}`, and there are object lists with already existing natural index sets in these providers. We could implement some heuristics like if a field with name `name` exists in the list element's schema, do not inject an artificial index field but rather declare the index set as `{"name"}`. The available [resource configuration API](https://github.com/crossplane/upjet/blob/50919febc5ab73b58020533d616a3d37227aff92/pkg/config/resource.go#L337) for specifying a `map` strategy for object lists allows this. However, there are also problems associated with such heuristics as for a proper SSA merge implementation, the index set must uniquely identify distinct objects. From a theoretical point of view, a heuristic like this one is flawed as it could potentially violate the index set uniqueness property. Of course, the `MaxLength = 1` constraint on the list size in our context relaxes the index set uniqueness constraint (as there can only be one element in the list) but we currently [don't enforce](https://github.com/crossplane/upjet/issues/10) the list size constraints at the API server. It's currently only enforced at the Terraform layer. As mentioned above, generating the proper CRD validation rules is on our roadmap.
- What happens if a (singleton) object list's `MaxItems = 1` constraint is relaxed and it's no longer a singleton? For certain object lists, this could be semantically viable and for some others, this does not make sense. Currently, when configuring the SSA merge strategy for object lists, we can make judgements in this regard. For example, if the singleton object list we are configuring is a `defaultNodeConfig`, I feel more comfortable because it's _the_ default node configuration, probably it doesn't make much sense to have multiple defaults in the future. This question is relevant in other contexts too, such as converting singleton lists into embedded objects.
- I also don't like the fact that we are injecting fields into our API as it complicates the API and the implementation. This idea had previously been considered as a violation of the high-fidelity principal of the Crossplane resource model [here](https://github.com/crossplane/crossplane/issues/3335#issuecomment-1264455941) but a client-side merge implementation never took off and we had to address crossplane-contrib/provider-upjet-aws/issues/975, and so opted to address it via the SSA merge strategy.

So, my proposal would be:
1. We implement the CRD validation rules that will enforce the size constraints from the underlying Terraform schema. The Terraform layer already enforces them but we had better enforce first at the API server level in alignment with the Terraform schema. 
1. We convert singleton lists into embedded objects so that we no longer need to configure the list map strategy and the associated index sets for them. Having introduced the support for multi-version CRDs has unblocked this.

What do you think @JonathanO?

JonathanO:
> We convert singleton lists into embedded objects so that we no longer need to configure the list map strategy and the associated index sets for them. Having introduced the support for multi-version CRDs has unblocked this.

That would be the ideal solution, but I raised this PR as a possible stop-gap solution until such a time that converting to embedded objects is implemented. If converting to embedded objects is likely to be available in the very near future then I agree that this PR is not necessary.

The atomic merge behaviour creates problems, and today the fix is to add ListTypeMap merge configuration to individual resources in providers. That's quite a frustrating user experience (because you discover the incorrect merge strategy only after running into problems), and will also need to be undone once the conversion to embedded objects is available.

Covering your other concerns:
> There might already exist an "index set"

As you go on to mention, this is not a problem for the case where `MaxLength = 1`, since there should only be one entry.

> What happens if a (singleton) object list's MaxItems = 1 constraint is relaxed

It's an API change and would need a new CRD version (which will be the same as if the singleton list had been converted to an embedded object.) Fortunately this seems unlikely to happen often given how these blocks are used in the Terraform.

> I also don't like the fact that we are injecting fields into our API

I agree, but at the moment we're raising PRs against providers to add these injected fields on a case-by-case basis - they're slowly ending up (inconsistently) in provider APIs already.

jeanduplessis:
In today's Upjet SIG meeting, we discussed not going ahead with this PR with @JonathanO. 

The team from Upbound will implement a solution where singleton lists (lists of only one (required or optional) object) are converted to embedded objects. We hope to roll out this solution in the next few weeks.

PR #381, opened 2024-03-18
Add a new configuration option for required field generation

sergenyalcin:

This PR adds a new configuration option for required field generation. The name of this new configuration API is `RequiredFields` and it is included to the `ExternalName`. The reason of this, we need to perform this operation in the context of external name configuration. If a resource has a value of optional (in TF schema) parameter, and the client do not provide this, we cannot construct the ID of this resource. For these type of situations we generate these fields as `Required` and also by marking `IdentifierField`. Also, the `IdentifierFields` API is also available under `ExternalName` and these two APIs (`RequiredFields` and `IdentifierFields`) are related to each other.

This change aims to generate fields as `Required` when needed, without interfering with the underlying provider schema. Before, for generating a field as `Required`, we used the [MarkAsRequired](https://github.com/crossplane/upjet/blob/2ef7077f6d1664868ead4a34bd1aa36be11e2266/pkg/config/common.go#L131) function. This function directly intervenes in the schema and sets the `Computed` and `Optional` values of the relevant field equal to `false`. Schema interventions may have the power to directly affect the internal operations of the underlying provider. This may lead to unexpected behavioral changes. We recently observed the mentioned situation in this [issue](https://github.com/upbound/provider-gcp/issues/472).

In this issue, it was observed that the behavior of resources that schema was intervened (using `MarkAsRequired`) changed after the underlying provider version bump, directly related to this schema intervention. This behavioral change directly blocks the provisioning of resources and even causes existing resources be `unsynced`. This reveals the extent of the dangers of schema interventions.

Of course, there are such needs for generation. However, such needs can be solved without schema intervention. This change paves the way for a field that is not required in the underlying provider schema to be generated as 'Required', with a high-level configuration API, without any schema intervention.

This PR also proposes the deprecation of `MarkAsRequired` API.

I have:

Tested generation in provider-gcp by removing MarkAsRequired and adding the following configuration:

```go
r.ExternalName.RequiredFields = append(r.ExternalName.RequiredFields, "zone")
```


ulucinar (commented):
Thanks @sergenyalcin, left some comments for you to consider.

pkg/config/resource.go
```diff
@@ -162,6 +162,10 @@ type ExternalName struct {
 	// management policy is including the Observe Only, different from other
 	// (required) fields.
 	IdentifierFields []string
+
+	// RequiredFields are the fields that are marked as required, although
+	// it is not required in the TF schema.
```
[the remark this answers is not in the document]
sergenyalcin: Done!

pkg/config/resource.go
```diff
@@ -162,6 +162,10 @@ type ExternalName struct {
...
+	RequiredFields []string
```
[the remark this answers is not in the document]
sergenyalcin: Thanks @ulucinar for this comment. I updated the PR and applied your suggested course of action:

- Moved the `Resource.ExternalName.RequiredFields` to `Resource.requiredFields`
- Deprecated `config.MarkAsRequired` in favor of a new configuration function on `*config.Resource` that still accepts a slice to mark multiple fields as required.

pkg/config/resource.go
```diff
@@ -489,6 +489,14 @@ type Resource struct {
 	// the value of the generated Kind, for example:
 	// "TagParameters": "ClusterTagParameters"
 	OverrideFieldNames map[string]string
+
+	// requiredFields are the fields that will be marked as required in the
+	// generated CRD schema, although they are not required in the TF schema.
+	requiredFields []string
+}
+
+func (r *Resource) RequiredFields() []string {
```
ulucinar, line 498: Let's add a doc comment here.

pkg/types/builder.go
```diff
@@ -381,6 +381,9 @@ func newTopLevelRequiredParam(path string, includeInit bool) *topLevelRequiredPa
 
 func (r *resource) addParameterField(f *Field, field *types.Var) {
 	requiredBySchema := !f.Schema.Optional
+	if f.Required {
+		requiredBySchema = f.Required
+	}
```
ulucinar, line 386: nit:
```suggestion
	requiredBySchema := !f.Schema.Optional || f.Required
```

ulucinar (approved):
Thanks @sergenyalcin, lgtm.

pkg/config/resource.go
sergenyalcin: Added!

pkg/types/builder.go
sergenyalcin: Done!

sergenyalcin:
The tests in GCP are successful: https://github.com/crossplane-contrib/provider-upjet-gcp/pull/490#issue-2194878925

PR #385, opened 2024-03-26
Add external API calls metric

mergenci:

Add a metric to count the number of external API calls, so that provider users can monitor their API call rate. Doing so is essential in addressing rate limiting concerns.

I have:

In my `provider-upjet-aws` code I called, `metrics.ExternalAPICalls.WithLabelValues("myService", "myOperation").Inc()`. I observed the effect by visiting `localhost:8080/metrics`, which is the provider's Prometheus metrics endpoint.

### DIFF
--- a/pkg/metrics/metrics.go
+++ b/pkg/metrics/metrics.go
@@ -44,6 +44,20 @@ var (
 		Buckets:   []float64{1, 5, 10, 15, 30, 60, 120, 300, 600, 1800, 3600},
 	}, []string{"operation"})
 
+	// ExternalAPICalls is a counter metric of the number of external
+	// API calls. "service" and "operation" labels could be used to
+	// classify calls into a two-level hierarchy, in which calls are
+	// "operations" that belong to a "service". Users should beware of
+	// performance implications of high cardinality that could occur
+	// when there are many services and operations. See:
+	// https://prometheus.io/docs/practices/naming/#labels
+	ExternalAPICalls = prometheus.NewCounterVec(prometheus.CounterOpts{
+		Namespace: promNSUpjet,
+		Subsystem: promSysResource,
+		Name:      "external_api_calls_total",
+		Help:      "The number of external API calls.",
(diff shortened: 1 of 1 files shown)


pkg/metrics/metrics.go
```diff
@@ -44,6 +44,13 @@ var (
 		Buckets:   []float64{1, 5, 10, 15, 30, 60, 120, 300, 600, 1800, 3600},
 	}, []string{"operation"})
 
+	ExternalAPICalls = prometheus.NewCounterVec(prometheus.CounterOpts{
+		Namespace: promNSUpjet,
+		Subsystem: promSysResource,
+		Name:      "external_api_calls",
```
ulucinar, line 50: Because the external API time metric is named as `ext_api_duration`, should we name the new external API call counter metric as `ext_api_calls` instead of `external_api_calls`? Or something like `ext_api_calls_counter`?

pkg/metrics/metrics.go
```diff
@@ -44,6 +44,13 @@ var (
...
+		Help:      "The number of external API calls.",
+	}, []string{"service", "service_operation"})
```
ulucinar, line 52: nit: Are `service` and `service_operation` nomenclature specific to AWS? Could you please check the counterparts in Azure & GCP? These labels will potentially be used by the other providers and the metric name and the label names become part of our API. Would something like `api_group` & `api` a more common name across the cloud service providers? What did they call these labels in the community provider?

ulucinar (commented):
Thank you @mergenci, left some comments to discuss.

Maybe we should also warn the provider developers regarding the dangers of [high cardinality](https://prometheus.io/docs/practices/naming/#labels), which may adversely affect the performance of the monitoring system if the provider has a very large number of observed service,operation pairs. I think such a warning would be suitable in the doc comments of the `metrics.ExternalAPICalls` counter.

ulucinar (approved):
Thank you @mergenci, lgtm.

PR #400, opened 2024-05-10
Allow Specification of the CRD API Version a Controller Watches & Reconciles

ulucinar:

This PR adds `config.Resource.ControllerReconcileVersion` through which the CRD API version which the associated controller will watch for & reconcile can be specified. This will allow us to set the controller's watch API version to the storage version of the CRD to prevent webhook conversions for multi-versioned CRDs.

In order to support this, we also add `config.Resource.TerraformConversions` which accepts a list of `config.TerraformConversion` implementations. These `TerraformConversion`s are responsible for converting data before it's passed down to the Terraform layer from the Crossplane layer and also for converting it back when it's carried up to the Crossplane layer from the Terraform layer. Currently, the only implementation is the `config.singletonListConversion`, which is responsible for converting between the singleton lists and embedded objects for converted APIs. Previously, this bidirectional conversion logic was hard-coded into the external client. This PR allows us to abstract this conversion logic behind the `config.TerraformConversion` interface.

We also address a bug in this PR discovered by @sergenyalcin in the context of https://github.com/crossplane-contrib/provider-upjet-gcp/pull/508. The singleton list CRD API converter (invoked by the conversion webhook) only handled API list API conversions in `spec.forProvider`, ignoring `spec.initProvider` and `status.atProvider`, where such conversions are still relevant. 

This PR also deprecates `config.Reference.Type` in favor of `config.Reference.TerraformName`. We had introduced `config.Reference.TerraformName` in https://github.com/crossplane/upjet/pull/12 and it's a more stable and less error prone API compared to `config.Reference.Type` because it automatically accounts for the configuration changes affecting the resource kind name, group or version. We've already been discouraging the `config.Reference.Type` in favor of `config.Reference.TerraformName` since it's been introduced.

I have:

Manually tested via provider-upjet-{azuread,gcp} resources. Will also be tested in the uptest runs in https://github.com/crossplane-contrib/provider-upjet-gcp/pull/508.


sergenyalcin (approved):
Thanks @ulucinar LGTM!

PR #406, opened 2024-05-22
Generate Secret References for Sensitive Parameters under the spec.initProvider API tree

ulucinar:

While working on adding the `User.mq` resource in `crossplane-contrib/provider-upjet-aws`, we hit the limitation that currently we do not generate secret references for the sensitive parameters under the `spec.initProvider` API. Currently, one can specify an  ActiveMQ broker user with `crossplane-contrib/provider-upjet-aws` as follows:
```yaml
apiVersion: mq.aws.upbound.io/v1beta1
kind: Broker
metadata:
  name: test-broker
spec:
  forProvider:
    ...
    user:
    - passwordSecretRef:
        key: password
        name: mq-secret
        namespace: upbound-system
      username: admin
```

However, we cannot fully specify the broker user using the `spec.initProvider` API tree because there's no `spec.initProvider.user[*].passwordSecretRef` and a partial specification like the following:

```yaml
apiVersion: mq.aws.upbound.io/v1beta1
kind: Broker
metadata:
  name: test-broker
spec:
  forProvider:
    # broker users are defined under spec.initProvider
    ...
  initProvider:
    user:
      username: admin
```
will not function as the required user password is missing in the user specification, which renders the `spec.initProvider.user` API useless.

This PR adds support for the generation of secret references for sensitive parameters under the `spec.initProvider` API tree. As a result, it now becomes possible to fully specify a broker user as follows:
```yaml
apiVersion: mq.aws.upbound.io/v1beta1
kind: Broker
metadata:
  name: test-broker
spec:
  initProvider:
    ...
    user:
    - passwordSecretRef:
        key: password
        name: mq-secret
        namespace: upbound-system
      username: admin
  forProvider:
    ...
```

The PR proposes both the code generation changes & the runtime secret resolution changes:
### Code generation changes:
1. Upjet now generates the secret references under the `spec.initProvider` API tree. As an example, `Broker.mq` now has `spec.initProvider.user[*].passwordSecretRef` as a sensitive parameter.
2. This will result in the marking of certain previously required sensitive fields as optional under the `spec.forProvider` API tree because they have now counterparts under the `spec.initProvider` subtree, i.e., they can now be specified from under `spec.initProvider`.
3. Upjet now generates, in `zz_*_terraformed.go` files, connection details mappings without the `spec.forProvider` prefixes in CRD paths. For example, what upjet used to generate as follows:
```yaml
// GetConnectionDetailsMapping for this SharedDirectory
func (tr *SharedDirectory) GetConnectionDetailsMapping() map[string]string {
  return map[string]string{"notes": "spec.forProvider.notesSecretRef"}
}
```
now becomes:
```yaml
// GetConnectionDetailsMapping for this SharedDirectory
func (tr *SharedDirectory) GetConnectionDetailsMapping() map[string]string {
  return map[string]string{"notes": "notesSecretRef"}
}

```
This is because, the mapping is now not 1:1. The sensitive value for the `notes` Terraform configuration argument can now be resolved from either a `spec.forProvider` reference or a `soec.initProvider` reference.

### Runtime changes:
Upjet runtime now has to be able to resolve secret references from under the `spec.initProvider` tree and the mapping between the Terraform configuration arguments & secret references is no longer one-to-one. If both a `spec.initProvider` and a `spec.forProvider` secret reference target the same Terraform configuration argument, then the `spec.forProvider`'s reference prevails over.

I have:

Manually tested via the `User.users` `crossplane-contrib/provider-upjet-azuread` resource.

### DIFF
--- a/pkg/resource/sensitive.go
+++ b/pkg/resource/sensitive.go
@@ -127,7 +127,7 @@ func GetSensitiveAttributes(from map[string]any, mapping map[string]string) (map
 			// Note(turkenh): k8s secrets uses a strict regex to validate secret
 			// keys which does not allow having brackets inside. So, we need to
 			// do a conversion to be able to store as connection secret keys.
-			// See https://github.com/crossplane/upjet/pull/94 for
+			// See https://github.com/crossplane/terrajet/pull/94 for
 			// more details.
 			k, err := fieldPathToSecretKey(fp)
 			if err != nil {
--- a/pkg/resource/sensitive.go
+++ b/pkg/resource/sensitive.go
@@ -22,13 +22,16 @@ import (
 )
 
 const (
-	errCannotExpandWildcards               = "cannot expand wildcards"
+	errGetAdditionalConnectionDetails = "cannot get additional connection details"
+	errCannotExpandWildcards          = "cannot expand wildcards"
+
 	errFmtCannotGetValueForFieldPath       = "cannot not get a value for fieldpath %q"
 	errFmtCannotGetStringForFieldPath      = "cannot not get a string for fieldpath %q"
 	errFmtCannotGetSecretKeySelector       = "cannot get SecretKeySelector from xp resource for fieldpath %q"
 	errFmtCannotGetSecretKeySelectorAsList = "cannot get SecretKeySelector list from xp resource for fieldpath %q"
 	errFmtCannotGetSecretKeySelectorAsMap  = "cannot get SecretKeySelector map from xp resource for fieldpath %q"
 	errFmtCannotGetSecretValue             = "cannot get secret value for %v"
+	errFmtCannotOverrideExistingKey        = "overriding a reserved connection key (%q) is not allowed"
 )
 
 const (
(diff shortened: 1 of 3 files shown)


sergenyalcin (approved):
Thanks @ulucinar LGTM!

ulucinar:
The CLI-based [external client](https://github.com/crossplane/crossplane-runtime/blob/fc036618ffd82023aeefe97554acdeb3a3db972f/pkg/reconciler/managed/reconciler.go#L309) relies on Terraform's [ignore_changes lifecycle meta-argument](https://developer.hashicorp.com/terraform/language/meta-arguments/lifecycle#ignore_changes) to ignore changes to MR fields that are exclusively set in `spec.initProvider` and not set in `spec.forProvider` for drift detection. This PR only considers ignoring changes in `spec.initProvider` sensitive fields (i.e., secret references) for the no-fork architecture and such changes will be considered as drift in the old architecture. We can work on this problem if there's need for it. The `spec.initProvider` use cases are rather limited and further, not many will involve the newly added sensitive fields. 

A possible implementation could use the mapping from the `spec.initProvider` sensitive field to the corresponding Terraform target sensitive argument and place that sensitive argument to the `ignore_changes` block. But let's tackle with it if there's any real use case for it.

PR #407, opened 2024-05-22
Add a new late-init configuration to skip already filled field in spec.initProvider

sergenyalcin:

This PR adds a new late-init API to skip already filled field in `spec.initProvider`.

Even though a field is specified in initProvider, it is late-init for forProvider. This can cause problems in some cases because `forProvider` is more powerful. With this new configuration API, the late-init operation of the field in `forProvider` can be skipped for fields set in `initProvider`.

I have:

Tested in provider-aws by using this configuration. The example configuration for aws ec2.Subnet resource:

```go
	p.AddResourceConfigurator("aws_subnet", func(r *config.Resource) {
		r.LateInitializer = config.LateInitializer{
			ConditionalIgnoredFields: []string{
				"cidr_block",
			},
		}
	})
```

Generated `LateInitialize` function:

```go
func (tr *Subnet) LateInitialize(attrs []byte) (bool, error) {
	params := &SubnetParameters_2{}
	if err := json.TFParser.Unmarshal(attrs, params); err != nil {
		return false, errors.Wrap(err, "failed to unmarshal Terraform state parameters for late-initialization")
	}
	opts := []resource.GenericLateInitializerOption{resource.WithZeroValueJSONOmitEmptyFilter(resource.CNameWildcard)}
	initParams, err := tr.GetInitParameters()
	if err != nil {
		return false, errors.Wrapf(err, "cannot get init parameters for resource '%q'", tr.GetName())
	}
	opts = append(opts, resource.WithConditionalFilter("CidrBlock", initParams))

	li := resource.NewGenericLateInitializer(opts...)
	return li.LateInitialize(&tr.Spec.ForProvider, params)
}
```

### DIFF
--- a/pkg/resource/lateinit.go
+++ b/pkg/resource/lateinit.go
@@ -10,12 +10,14 @@ import (
 	"runtime/debug"
 	"strings"
 
+	"github.com/crossplane/crossplane-runtime/pkg/fieldpath"
 	xpmeta "github.com/crossplane/crossplane-runtime/pkg/meta"
 	xpresource "github.com/crossplane/crossplane-runtime/pkg/resource"
 	"github.com/pkg/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
 	"github.com/crossplane/upjet/pkg/config"
+	"github.com/crossplane/upjet/pkg/types/name"
 )
 
 const (
@@ -43,8 +45,9 @@ const (
 
 // GenericLateInitializer performs late-initialization of a Terraformed resource.
 type GenericLateInitializer struct {
-	valueFilters []ValueFilter
-	nameFilters  []NameFilter
+	valueFilters       []ValueFilter
+	nameFilters        []NameFilter
+	conditionalFilters []ConditionalFilter
 }
 
 // SetCriticalAnnotations sets the critical annotations of the resource and reports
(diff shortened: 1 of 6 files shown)


pkg/resource/lateinit.go
```diff
@@ -230,6 +262,16 @@ func (li *GenericLateInitializer) handleStruct(parentName string, desiredObject
 			continue
 		}
 
+		for _, f := range li.conditionalFilters {
+			if f(cName) {
```
ulucinar, line 266: If we follow the above suggestions, this would turn into something like:
```suggestion
			if f(cName, desiredPaved, actualPaved) {
```

pkg/types/field.go
```diff
@@ -162,6 +162,12 @@ func NewField(g *Builder, cfg *config.Resource, r *resource, sch *schema.Schema,
 		}
 	}
 
+	for _, ignoreField := range cfg.LateInitializer.ConditionalIgnoredFields {
+		if ignoreField == traverser.FieldPath(f.TerraformPaths) {
+			cfg.LateInitializer.AddConditionalIgnoredCanonicalFields(traverser.FieldPath(f.CanonicalPaths))
+		}
+	}
+
```
ulucinar, line 170: How about accepting the actual CRD canonical path during the configuration? Then we would not need to do the conversion here.

ulucinar (commented):
Thank you @sergenyalcin. Left some suggestions to generalize the implementation to support other types of conditional filters that we can introduce on top of this implementation. There are some other types of conditional filters that we need to prevent certain categorical errors in addition to the race-condition preventing filter that we are adding in this PR. 

We may also decide to accept the approach proposed in this PR and do a generalization in a follow-up PR. Let's discuss.

ulucinar (approved):
We've sync'ed offline and decided to continue with this PR as is. We will consider the generalization of the conditions for skipping the late-initialization in future iterations. We've successfully validated this PR & #406 together in the context of https://github.com/crossplane-contrib/provider-upjet-aws/pull/1311.

PR #409, opened 2024-05-24
add crossplane-external-name tag to external resources

### LINKED ISSUE #408: Ensure a tag containing external-name is present in all external resources supporting tags

Many cloud provider external resources are associated with a non-deterministic ID/external-name. To deal with that, Crossplane providers rely on the `crossplane.io/external-name` annotation for discovering existing external resources that a managed resource controls.

It's not unusual for Crossplane users to set `deletionPolicy` to `Orphan` on production environments to prevent accidental deletions of external-resources in the event of human error or catastrophic failure in Kubernetes clusters requiring re-creation of all resources.

During such events, the MR gets re-created without the annotation, and one of these scenarios follow:

1. The MR breaks and never syncs, because it identifies the resource you're attempting to manage already exists, and it does not get imported automatically. Human intervention is required.
2. The MR results in the creation of a different resource, effectively duplicating everything, and the provider is now managing a resource which might not even be at use. In this scenario, the failure to import the existing resource is not reported, and it's marked as healthy (even though the original resource is now orphaned and unmanaged). Not only requires manual intervention, but might go unnoticed until other things start to break.

These problems are aggravated when compositions are used, as it makes it all more complex. For example, imagine a composition that manages DNS hosted zones. As well as creating the zone itself, it might also create NS records on the parent zone, to delegate authority to it.

Within the dropdown you'll find a concrete example of a real scenario we faced.

<details>
  <summary>Detailed Example</summary>

  To make this example more concrete, let's use AWS' Route53. Assume the zone already exists in AWS, previously created by upjet-aws-provider.
  
  ```mermaid
  flowchart TD
   subgraph subgraph_aws["AWS"]
          hosted_zone(["Hosted Zone zone1 (foo.bar.baz)"])
          ns_record(["NS Record (on zone bar.baz, points to zone1)"])
    end
      mr_zone["Zone Managed Resource (external-name zone1)"]
      mr_record["Record Managed Resource"]
      xr["XR"] --> mr_zone & mr_record
      mr_zone --> hosted_zone
      mr_record --> ns_record
  
  ```
  
  However, for one reason or another, the claim/XR/MRs that manage it had to be re-created. The resources were left in AWS. So far no disruption to DNS resolution is happening.
  
  ```mermaid
  flowchart TD
   subgraph subgraph_aws["AWS"]
          hosted_zone(["Hosted Zone zone1 (foo.bar.baz)"])
          ns_record(["NS Record (on zone bar.baz, points to zone1)"])
    end
  ```
  
  Once the claim, XR and MRs are recreated, a duplicate DNS zone will be created, and it will be empty.
  
  ```mermaid
  flowchart TD
   subgraph subgraph_aws["AWS"]
          hosted_zone(["Hosted Zone zone1 (foo.bar.baz)"])
          duplicated_zone(["Hosted Zone zone2 (foo.bar.baz)"])
          ns_record(["NS Record (on zone bar.baz, points to zone1)"])
    end
      mr_zone["Zone Managed Resource (external-name zone2)"]
      mr_record["Record Managed Resource"]
      xr["XR"] --> mr_zone & mr_record
      mr_zone --> duplicated_zone
      mr_record --> ns_record
  
  ```
  
  
  Because it throws no errors to halt the composition (as far as Crossplane is concerned, everything is as it should be), the NS resource also gets updated on the parent DNS zone, delegating authority to the new, empty DNS zone. 
  
  From this point forward, all DNS resolution for the original zone start failing, until an engineer updates the external-name annotation on the hosted zone MR to point to the old zone.

</details>

Upjet could ensure that a tag with the external-name value is inserted into the external resource, along with `crossplane-kind` and `crossplane-name` tags it already currently inserts.

This would allow for importing mechanisms to filter existing resources using tags, then setting the external-name annotation on the managed resource if it finds the external-name tag set on it.

### Proof of Concept

To deal exactly with the problem I mentioned we wrote a composition function that uses AWS' [Resource Groups Tagging API](https://docs.aws.amazon.com/resourcegroupstagging/latest/APIReference/overview.html) to filter any AWS resource based on the tag Crossplane/Upjet already currently populates (`crossplane-kind`, `crossplane-name`).

https://github.com/gympass/function-aws-importer

The nice thing about this tagging API is that you can make the same API calls regardless of the exact type of the external resource, as long as it supports tags (most resources do). It works for Route53 Hosted Zones, EC2 Security Groups, etc.

It currently requires composition authors to ensure the external-name tag is there themselves, which leads to longer reconciliations, weird corner cases, and overall non-ideal user experience. If Upjet ensured this tag was there itself, we could avoid all of this and make the function itself self-contained.

To have a better idea of how this would work on Upjet, we deployed a fork with the following change: https://github.com/crossplane/upjet/commit/a6c0bc99c89e59363d279b81b1858a43e9fabbff

This is not necessarily the best way to implement this, but it proves the concept. We've been using this fork along with our function in our development environment with no failures to report so far.

### Future

In the future, if this concept proves to be robust, Upjet could even use this tag to assemble the tfstate it produces on observe operations, making the function above useless by incorporating its functionality entirely. One step at a time though :-)

LCaparelli:

Ensures a new tag containing the value set to the external-name annotation exists in external resources.

Fixes #408 

I kinda didn't wait for discussion to pan out on the issue, and I understand if this is not at all the direction we want to take. Just thought it'd be nice to present a problem along with a possible solution. If you have concerns about the direction, please feel free to chime in at the issue or ping me directly on Crossplane's Slack.

I have:

I have pulled provider-aws-upjet locally, checked out at the v1.5.0 tag, pointed the dependency to this branch using go.mod replace, built the EC2 provider and deployed it in our development environments. All EC2 managed resources now have this additional tag.

### DIFF
--- a/pkg/config/resource.go
+++ b/pkg/config/resource.go
@@ -12,6 +12,7 @@ import (
 
 	xpv1 "github.com/crossplane/crossplane-runtime/apis/common/v1"
 	"github.com/crossplane/crossplane-runtime/pkg/fieldpath"
+	"github.com/crossplane/crossplane-runtime/pkg/meta"
 	"github.com/crossplane/crossplane-runtime/pkg/reconciler/managed"
 	xpresource "github.com/crossplane/crossplane-runtime/pkg/resource"
 	fwresource "github.com/hashicorp/terraform-plugin-framework/resource"
@@ -315,7 +316,7 @@ func (t *Tagger) Initialize(ctx context.Context, mg xpresource.Managed) error {
 	if err != nil {
 		return err
 	}
-	pavedByte, err := setExternalTagsWithPaved(xpresource.GetExternalTags(mg), paved, t.fieldName)
+	pavedByte, err := setExternalTagsWithPaved(getExternalTags(mg), paved, t.fieldName)
 	if err != nil {
 		return err
 	}
@@ -328,11 +329,20 @@ func (t *Tagger) Initialize(ctx context.Context, mg xpresource.Managed) error {
 	return nil
 }
 
+const externalResourceTagKeyExternalName = "crossplane-external-name"
+
+func getExternalTags(mg xpresource.Managed) map[string]string {
+	externalTags := xpresource.GetExternalTags(mg)
+	externalTags[externalResourceTagKeyExternalName] = mg.GetAnnotations()[meta.AnnotationKeyExternalName]
+	return externalTags
+}
+
 func setExternalTagsWithPaved(externalTags map[string]string, paved *fieldpath.Paved, fieldName string) ([]byte, error) {
 	tags := map[string]*string{
 		xpresource.ExternalResourceTagKeyKind:     ptr.To(externalTags[xpresource.ExternalResourceTagKeyKind]),
 		xpresource.ExternalResourceTagKeyName:     ptr.To(externalTags[xpresource.ExternalResourceTagKeyName]),
 		xpresource.ExternalResourceTagKeyProvider: ptr.To(externalTags[xpresource.ExternalResourceTagKeyProvider]),
+		externalResourceTagKeyExternalName:        ptr.To(externalTags[externalResourceTagKeyExternalName]),
 	}
 
 	if err := paved.SetValue(fmt.Sprintf("spec.forProvider.%s", fieldName), tags); err != nil {
--- a/pkg/config/resource_test.go
+++ b/pkg/config/resource_test.go
@@ -11,17 +11,20 @@ import (
 
 	"github.com/crossplane/crossplane-runtime/pkg/errors"
 	"github.com/crossplane/crossplane-runtime/pkg/fieldpath"
+	"github.com/crossplane/crossplane-runtime/pkg/meta"
 	xpresource "github.com/crossplane/crossplane-runtime/pkg/resource"
 	"github.com/crossplane/crossplane-runtime/pkg/resource/fake"
 	"github.com/crossplane/crossplane-runtime/pkg/test"
 	"github.com/google/go-cmp/cmp"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
 const (
-	kind     = "ACoolService"
-	name     = "example-service"
-	provider = "ACoolProvider"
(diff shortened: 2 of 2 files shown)


LCaparelli:
In order to make my fork work, I had to manually assign RBAC permissions to the generated ServiceAccount:
```yaml
rules:
- apiGroups:
  - aws.upbound.io
  resources:
  - providerconfigs
  - providerconfigs/status
  - providerconfigusages
  - providerconfigusages/status
  - storeconfigs
  - storeconfigs/status
  verbs:
  - '*'
```

I think it might be because I only built the EC2 provider, and not the entire family. Honestly I'm not sure. I don't see how this change could interfere with that, so I'm guessing this is unrelated, but appreciate any pointers.

pkg/config/resource.go
```diff
@@ -328,11 +329,20 @@ func (t *Tagger) Initialize(ctx context.Context, mg xpresource.Managed) error {
 	return nil
 }
 
+const externalResourceTagKeyExternalName = "crossplane-external-name"
+
+func getExternalTags(mg xpresource.Managed) map[string]string {
+	externalTags := xpresource.GetExternalTags(mg)
```
LCaparelli, line 335: This was the only call to `xpresource.GetExternalTags` I could find in the code, so I'm assuming this is the best and only place to provide this additional tag. If there are other places I'm missing, please let me know and I'll adjust.

ulucinar:
Hi @LCaparelli,
Would you like to discuss this proposal in the next [sig-upjet](https://crossplane.slack.com/archives/C05T19TB729) meeting, which is scheduled for Wednesday, June 5 at 3:00pm GMT?

LCaparelli:
@ulucinar thanks, I would love to, but at that time I'll be on a holiday trip. Maybe o the next one?

I'm able to discuss async at #408 until then.

jbw976:
@ulucinar feel free to checkout the recording of @LCaparelli presenting on this topic: https://github.com/crossplane/upjet/issues/408#issuecomment-2145458535

LCaparelli:
@ulucinar I added/suggested a discussion topic to SIG Upjet's [meeting backlog](https://docs.google.com/document/d/1QJfsAk-pdo_j2cKsJtG6W-BOQh_3ilhiRVTYVIGkDXM/edit)

PR #415, opened 2024-06-02
feat: add AddReference method to References

yordis:

@haarchri was helping at https://github.com/crossplane-contrib/provider-upjet-digitalocean/pull/41 (thank you), and during the code review, I noticed a oversee where he replaced the existing `r.References` as follow:

```diff
diff --git a/config/provider.go b/config/provider.go
index 83eada3..8e69d59 100644
--- a/config/provider.go
+++ b/config/provider.go
@@ -294,13 +294,9 @@ func GetProvider() *ujconfig.Provider {
 			Type:          referenceType(pc, “project”, “v1alpha1", “Project”),
 			TerraformName: “digitalocean_project”,
 		}
-
-		r.References = map[string]ujconfig.Reference{
-			“forwarding_rule.certificate_name”: {
-				TerraformName: “digitalocean_certificate”,
-			},
+		r.References[“forwarding_rule.certificate_name”] = ujconfig.Reference{
+			TerraformName: “digitalocean_certificate”,
 		}
-
 	})
 	pc.AddResourceConfigurator(“digitalocean_domain”, func(r *ujconfig.Resource) {
 		r.ShortGroup = “dns”
```

I am glad that I saw the diff, but honestly, this is a straightforward thing to miss when they are so much code-generated files noise, or people tend to ignore those.

As @haarchri explained, "That's why I prefer this, for example, https://github.com/crossplane-contrib/provider-upjet-aws/blob/main/config/elbv2/config.go#L13," which is a really common scenario.

I never considered using the existing style. I didn't think much about it. He didn’t try to break things. Still, we ran into the issue. Since we are talking about "style" here, it is really difficult to be objective and define who is more proper.

So, the fix here is to expose an SDK API that avoids such a mistake.

I am not proposing adding getters and setters everywhere either, just reacting to the situation and figuring out how to mitigate the issues in the future; as I usually say, "let the code teach you what to do, not the other way around."

I have:

- Unit Testing

### DIFF
--- a/pkg/config/resource.go
+++ b/pkg/config/resource.go
@@ -197,12 +197,28 @@ type TFPluginFrameworkOptions struct {
 	ComputedIdentifierAttributes []string
 }
 
+var ErrReferenceAlreadyExists = errors.New("reference for field already exists")
+
 // References represents reference resolver configurations for the fields of a
 // given resource. Key should be the field path of the field to be referenced.
 // The key is the Terraform field path of the field to be referenced.
 // Example: "vpc_id" or "forwarding_rule.certificate_name" in case of nested
 // in another object.
 type References map[string]Reference
+
+// AddReference adds a reference configuration for the given field path.
+// The fieldPath is the Terraform field path of the field to be referenced.
+//
+// Example: "vpc_id" or "forwarding_rule.certificate_name" in case of nested
+// in another object.
+func (r References) AddReference(fieldPath string, ref Reference) error {
+	if _, ok := r[fieldPath]; ok {
+		return ErrReferenceAlreadyExists
+	}
+
+	r[fieldPath] = ref
+	return nil
+}
 
 // Reference represents the Crossplane options used to generate
 // reference resolvers for fields
--- a/pkg/config/resource_test.go
+++ b/pkg/config/resource_test.go
@@ -296,3 +296,34 @@ func TestRemoveSingletonListConversion(t *testing.T) {
 		})
 	}
 }
+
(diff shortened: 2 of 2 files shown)


yordis:
A few things from this,

1. Should all documentation and code-generated code be updated to this?
2. Should `References` be a private field instead of exposing it? You can still assign `r.References`. Ideally, Provider Authors shouldn't care about replacing the entire `r.References,` and in that case, we can add `r.ReplaceReferences(references)`, but I am not sure when this would be necessary.

pkg/config/resource.go
```diff
@@ -165,10 +165,26 @@ type ExternalName struct {
...
 type References map[string]Reference
 
+// AddReference adds a reference configuration for the given field path.
+// The fieldPath is the Terraform field path of the field to be referenced.
+//
+// Example: "vpc_id" or "forwarding_rule.certificate_name" in case of nested
+// in another object.
+func (r References) AddReference(fieldPath string, ref Reference) error {
+	if _, ok := r[fieldPath]; ok {
+		return ErrReferenceAlreadyExists
```
mbbush, line 181: I don't love returning an error in this situation. My instinct is that the action of setting a reference should be idempotent; as you've implemented it all times it's invoked but the first will fail with an error. 

Maybe it's ok because it would happen at generation time, although I don't know if we could guarantee that.

yordis: Hey @mbbush, what would be the resolution here? I am not sure if you requested some changes or not

PR #417, opened 2024-06-06
Do not prefix JSON fieldpaths starting with status.atProvider in resource.GetSensitiveParameters

### LINKED ISSUE #416: Wrong jsonPath to get sensitive parameters with latest 1.4.0
### What happened?
It seems this PR #406 introduced a bug.
With an AWS IAM AccessKey, I endup with the following error, because of a wrong jsonPath (`spec.forProvider.status.atProvider.secret`)

```
cannot store sensitive parameters into params: cannot expand wildcard for xp resource: cannot expand wildcards for segments: \"spec.forProvider.status.atProvider.secret\": spec.forProvider.status: not an object"
```

### How can we reproduce it?

1. use the latest `provider-aws-iam` `1.5.0` builded with the latest `upjet` `1.4.0`
2. create an AWS IAM AccessKey resource and look for the reconcile error in the object status

ulucinar:

Fixes #416 

If the MR API has a `spec.forProvider.status` field and there are sensitive attributes, then `fieldpath.Paved.ExpandWildcards` complains instead of expanding as an empty slice, which breaks the reconciliation.

This bug has been observed with the [`AccessKey.iam` resource](https://marketplace.upbound.io/providers/upbound/provider-aws/v1.5.0/resources/iam.aws.upbound.io/AccessKey/v1beta1) of `crossplane-contrib/provider-upjet-aws@v1.5.0`, where we have introduced the sensitive parameters under the `spec.initProvider` API tree by consuming the upjet changes from #406, and reported in #416.

This bug only affects those resources with a [`spec.forProvider.status` field](https://github.com/crossplane-contrib/provider-upjet-aws/blob/4b4cac415e215870c7f04328b51a4751c28e706c/apis/iam/v1beta1/zz_accesskey_types.go#L72) and with [sensitive attributes](https://github.com/crossplane-contrib/provider-upjet-aws/blob/4b4cac415e215870c7f04328b51a4751c28e706c/apis/iam/v1beta1/zz_accesskey_terraformed.go#L24), which are removed from `status.atProvider` in the corresponding CRD schema but still represented in the connection details mapping so that the corresponding sensitive information can be published in the connections details secret.

Before #406, such mappings were effectively being ignored by the [`fieldpath.Paved.ExpandWildcards`](https://github.com/crossplane/upjet/blob/577bfa78fba5f8c401021027970b57cfbd7e94ea/pkg/resource/sensitive.go#L108) because such sensitive attributed are removed, during code generation, from `status.atProvider`. However, #406 now prefix such JSON fieldpath expressions with `spec.forProvider` (and `spec.initProvider`) and if there already exists a `spec.forProvider.status` field, then the `fieldpath` library now errors instead of expanding the expressions as an empty slice (as it does in other cases).

Arguably we could address this in the `fieldpath` library but this PR proposes to address this issue by reverting back to the old conditions: If the JSON fieldpath expression starts with a `status.atProvider`, then we do not prefix it with `spec.forProvider` or `spec.initProvider`, which restores the old behavior as there cannot be the corresponding `status.atProvider` fields and the `fieldpath` library now correctly handles the wildcard expansion by returning an empty expanded set instead of erroring.

I have:

Tested with https://github.com/crossplane-contrib/provider-upjet-aws/pull/1344.

### DIFF
--- a/pkg/resource/sensitive.go
+++ b/pkg/resource/sensitive.go
@@ -168,17 +168,25 @@ func GetSensitiveParameters(ctx context.Context, client SecretClient, from runti
 		return err
 	}
 	pavedTF := fieldpath.Pave(into)
+	prefixes := []string{"spec.initProvider.", "spec.forProvider."}
 
 	for tfPath, jsonPath := range mapping {
 		jp := jsonPath
 		groups := reFieldPathSpec.FindStringSubmatch(jsonPath)
 		if len(groups) == 3 {
 			jp = groups[2]
+		} else if strings.HasPrefix(jsonPath, "status.atProvider.") {
+			// we will not be prefixing the JSON fieldpath expression if it starts
+			// with "status.atProvider" in case there is a spec.forProvider.status
+			// field. If there exists a spec.forProvider.status field, then the
+			// fieldpath.ExpandWildcards will complain instead of expanding the
+			// fieldpath expression as an empty slice.
+			prefixes = []string{""}
 		}
 
 		// spec.forProvider secret references override the spec.initProvider
 		// references.
-		for _, p := range []string{"spec.initProvider.", "spec.forProvider."} {
+		for _, p := range prefixes {
 			if err := storeSensitiveData(ctx, client, tfPath, p+jp, pavedTF, pavedJSON, mapping); err != nil {
 				return err
 			}


sergenyalcin (approved):
Thanks @ulucinar LGTM!

PR #461, opened 2025-01-16
Remove diff calculation in observe-only reconciliation

mergenci:

Fixes https://github.com/crossplane-contrib/provider-upjet-aws/issues/1597.

While importing a DynamoDB Table, in the linked issue, Terraform executes [`CustomizeDiff` functions](https://github.com/hashicorp/terraform-provider-aws/blob/v5.82.2/internal/service/dynamodb/table.go#L69) when Upjet calculates the diff in `Observe()`. [One of the `CustomizeDiff` functions](https://github.com/hashicorp/terraform-provider-aws/blob/v5.82.2/internal/service/dynamodb/table.go#L74) is a validation function that checks whether the configuration that is subject to diff is valid. Because the configuration is practically empty — it only contains the resource name in the import path — validation fails and returns an error, which in turn causes our `Observe()` to fail.

I've checked how the behavior of managed reconciler changes for observe-only reconciliation. When Upjet always returns `ResourceUpToDate = true`, [this block](https://github.com/crossplane/crossplane-runtime/blob/v1.17.0/pkg/reconciler/managed/reconciler.go#L1235-L1254) executes. Otherwise, when Upjet calculates the diff and returns `ResourceUpToDate` accordingly, [this block](https://github.com/crossplane/crossplane-runtime/blob/v1.17.0/pkg/reconciler/managed/reconciler.go#L1260-L1266) executes in case `ResourceUpToDate = false`. The difference between those two blocks are (1) debug messages and (2) metric records.

The difference doesn't matter for the import path, in which `ResourceUpToDate` doesn't make much sense. We can consider `ResourceUpToDate = true`, by definition, which is the case in this PR. The difference might matter in observe-only reconciliation that is not an import operation.

I have:

I tested manually, using the resource manifest given in the linked issue above. Here are the steps I followed:
1. Create the resource with `deletionPolicy: Orphan`.
2. Delete the resource.
3. Import the resource with `managementPolicies: ["Observe"]`.
4. Check that the import succeeded without any errors in the debug console.
5. Copy essential fields from `status.atProvider` to `spec.forProvider`.
6. Set `managementPolicies: ["*"]`.
7. Add a tag to the resource and check if the update succeeded without any errors in the console.

In addition to DynamoDB Table, I tested a VPC resource as well.

I also ran Uptests in https://github.com/crossplane-contrib/provider-upjet-aws/pull/1651.

### DIFF
--- a/pkg/controller/external.go
+++ b/pkg/controller/external.go
@@ -330,6 +330,10 @@ func (e *external) Observe(ctx context.Context, mg xpresource.Managed) (managed.
 		if e.eventHandler != nil {
 			e.eventHandler.Forget(rateLimiterStatus, mg.GetName())
 		}
+
+		// TODO(cem): Consider skipping diff calculation (terraform plan) to
+		// avoid potential config validation errors in the import path. See
+		// https://github.com/crossplane/upjet/pull/461
 		plan, err := e.workspace.Plan(ctx)
 		if err != nil {
 			return managed.ExternalObservation{}, errors.Wrap(err, errPlan)
--- a/pkg/controller/external_tfpluginfw.go
+++ b/pkg/controller/external_tfpluginfw.go
@@ -330,6 +330,9 @@ func (n *terraformPluginFrameworkExternalClient) Observe(ctx context.Context, mg
 		}
 	}
 
+	// TODO(cem): Consider skipping diff calculation to avoid potential config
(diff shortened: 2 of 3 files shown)


sergenyalcin (approved):
Thanks @mergenci LGTM!

PR #465, opened 2025-02-05
Expose conversion option to inject key/values in the conversion to list

turkenh:

We have observed some issue with the EKS Cluster resource where when we apply the manifest and let it to resolve references to `spec.forProvider.vpcConfig.securityGroupIds`, we noticed the following error in the status:

```
  - lastTransitionTime: "2025-02-06T07:28:46Z"
    message: 'cannot patch the managed resource via server-side apply: Cluster.eks.aws.upbound.io
      "some-eks-cluster" is invalid: [spec.forProvider.vpcConfig: Invalid
      value: "null": spec.forProvider.vpcConfig in body must be of type array: "null",
      <nil>: Invalid value: "null": some validation rules were not checked because
      the object was invalid; correct the existing errors to complete validation]'
    reason: ReconcileError
    status: "False"
    type: Synced
```

Our assumption (which is validated by observing this changes fixes the problem) is as follows:

Server Side Apply uses [merge strategies](https://kubernetes.io/docs/reference/using-api/server-side-apply/#merge-strategy) to make decisions during merging changes by various owners/managers. In EKS Cluster object, for `vpcConfig` field, we mark `+listType` as map and `//+listMapKey` as `index`. During conversion between `v1beta1` to `v1beta2`, we convert the that field from array to object, losing the index field since it is not in the schema. This is fine in most cases since in `v1beta1` of the object schema index defaults to `"0"` , even though you don’t provide it.

However, with Server Side Apply, apparently some on the fly conversions happening when different managers using different api versions and losing index field causing unexpected merging results and drop of the whole `spec.forProvider.vpcConfig` object.

I have:

- [ ] Read and followed Upjet's [contribution process].

Unit tests and a custom build for provider-aw-eks with [this change](https://github.com/crossplane-contrib/provider-upjet-aws/pull/1669).

### DIFF
--- a/pkg/config/conversion/conversions.go
+++ b/pkg/config/conversion/conversions.go
@@ -178,22 +178,38 @@ func NewCustomConverter(sourceVersion, targetVersion string, converter func(src,
 
 type singletonListConverter struct {
 	baseConversion
-	pathPrefixes []string
-	crdPaths     []string
-	mode         ListConversionMode
+	pathPrefixes   []string
+	crdPaths       []string
+	mode           ListConversionMode
+	convertOptions *ConvertOptions
+}
+
+type SingletonListConversionOption func(*singletonListConverter)
+
+// WithConvertOptions sets the ConvertOptions for the singleton list conversion.
+func WithConvertOptions(opts *ConvertOptions) SingletonListConversionOption {
+	return func(s *singletonListConverter) {
+		s.convertOptions = opts
+	}
 }
 
 // NewSingletonListConversion returns a new Conversion from the specified
 // sourceVersion of an API to the specified targetVersion and uses the
 // CRD field paths given in crdPaths to convert between the singleton
 // lists and embedded objects in the given conversion mode.
-func NewSingletonListConversion(sourceVersion, targetVersion string, pathPrefixes []string, crdPaths []string, mode ListConversionMode) Conversion {
-	return &singletonListConverter{
+func NewSingletonListConversion(sourceVersion, targetVersion string, pathPrefixes []string, crdPaths []string, mode ListConversionMode, opts ...SingletonListConversionOption) Conversion {
+	s := &singletonListConverter{
 		baseConversion: newBaseConversion(sourceVersion, targetVersion),
 		pathPrefixes:   pathPrefixes,
 		crdPaths:       crdPaths,
 		mode:           mode,
 	}
+
+	for _, o := range opts {
+		o(s)
+	}
+
+	return s
 }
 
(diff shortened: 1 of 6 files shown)


sergenyalcin (approved):
@turkenh Firstly, thank you very much for your very valuable effort and handling this. And also thank you for your time that you spent resolving this issue🙏🏼. Your solution addresses the problem effectively and makes a lot of sense to me.

nit: This is not a blocker. I just wanted to note an alternative for future consideration.

From a design perspective, I was wondering about the possibility of making `ConvertOptions` available in a more generic way. Currently, we have a specific definition and are modifying the `Convert` function for this case, which leads to resource/provider-specific business logic being implemented on the upjet side. If we think of `ConvertOptions` as an interface, it might offer a chance to configure it through a broader and more generic API. This could also help decouple resource/provider-specific logic from upjet (although how exactly to achieve this requires further thought—just a high-level comment).

That said, I’m not fully convinced that this level of generalization is necessary, since it’s unclear how often we’ll encounter similar cases or what other custom configurations might be needed in the future.

Once again, thank you for your effort—it solves the problem effectively. LGTM!

turkenh:
@turkenh Firstly, thank you very much for your very valuable effort and handling this. And also thank you for your time that you spent resolving this issue🙏🏼. Your solution addresses the problem effectively and makes a lot of sense to m

❤️ 🙇  

> Currently, we have a specific definition and are modifying the Convert function for this case, which leads to resource/provider-specific business logic being implemented on the upjet side. If we think of ConvertOptions as an interface, it might offer a chance to configure it through a broader and more generic API. This could also help decouple resource/provider-specific logic from upjet (although how exactly to achieve this requires further thought—just a high-level comment).

Not really. Currently, we don't have resource/provider specific logic embedded into upjet. This PR exposes an option to inject in the `SingletonListConverter`, with a non breaking change by introducing an optional argument (`opts`) to the `NewSingletonListConversion` function. So, logic still lives on the provider side per resource, see the example PR on [provider side](https://github.com/crossplane-contrib/provider-upjet-aws/pull/1669/files#diff-d49e9fd43f2b540f81d60e86fa6ba2e4623334ea5d1fbc10711d79ef78db73efR239-R258). At first, I went one step further and introduced this option in the [config.Resource](https://github.com/crossplane/upjet/blob/main/pkg/config/resource.go#L407) struct, enabling overriding this under `config/eks/config.go` just like the other configuration, but later reverted that since this is only needed by a single resource right now and no need to generalize further. Another advantage is this implementation is closer to conversion initialization. Happy to explain my thought process further if this still doesn't makes sense :)

PR #480, opened 2025-03-23
Generate namespaced managed resources

jbw976:

This PR updates the upjet code generation logic to optionally generate namespaced resources in addition to cluster scoped resources. The results can be seen in action in https://github.com/crossplane-contrib/provider-upjet-aws/pull/1751.

Some highlights to be aware of:

* the code generation pipeline can now be run for namespaced resources. This is a second pipeline run in addition to the one for cluster scoped resources. Some hard coded directory paths were made configurable to enable common pipeline code to run for both cluster and namespaced.
* the external client and support files is now namespace aware. In places where just a name was used, a NamespacedName is now used.
* Conversions can now be registered for both cluster and namespaced resources. This code originally assumed it would be singleton instance that only contained cluster scoped objects, but now also maintains a map for namespaced resources as well.

This should not be taken as a final product. This is a practical update to this provider as a proof of concept for what we think a provider could look like in Crossplane v2.  There are many short cuts taken, it is not production ready, and we are very open to feedback to change directions or improve the architecture.

The work in this PR should be merged into the new `crossplane-v2` branch.

I have:

By consuming these changes in https://github.com/crossplane-contrib/provider-upjet-aws/pull/1751 and running `make generate` to generate and build that provider with namespaced resources.


pkg/pipeline/run.go
```diff
@@ -23,8 +23,81 @@ type terraformedInput struct {
...
+			Scope: "Namespaced",
+		}
+
+		// Map of service name (e.g. ec2) to resource controller packages. Should be
+		// the same for cluster and namespaced, so we only save one.
+		groups = cluster.Run(pcCluster)
+		_ = namespaced.Run(pcNamespace)
+	}
+
+	if len(pcCluster.MainTemplate) > 0 {
```
negz, line 79: It's unfortunate that when there's two configs, we'll only use the main template from one of them (and you need to know which). Ideally we'd break `MainTemplate` out to be something separate from either config.

pkg/pipeline/run.go
```diff
@@ -23,8 +23,81 @@ type terraformedInput struct {
...
+			DirHack:        filepath.Join(rootDir, "hack"),
+
+			ModulePathAPIs:        filepath.Join(pcCluster.ModulePath, "apis"),
+			ModulePathControllers: filepath.Join(pcCluster.ModulePath, "internal", "controller"),
+
+			Scope: "Cluster",
+		}
+
+		groups = cluster.Run(pcCluster)
+	} else {
```
negz, line 47: Instead of this pattern where most of this function is inside an `if/else` clause, I'd prefer a flow like:

```go
func Run(...) {
	// Handle the case where there's no namespaced config inside the smaller `if` block
	if pcNamespace == nil {
		cluster := ...
		if pcCluster.MainTemplate != "" { ... }
		return
	}

	// Handle the case where there's both with less nesting
	cluster := ...
	namespaced := ...
	if pcCluster.MainTemplate != "" { ... }
	return
}
```

pkg/controller/conversion/registry.go
```diff
@@ -21,28 +21,37 @@ var instance *registry
...
 // GetConversions returns the conversion.Conversions registered in this
 // registry for the specified Terraformed resource.
 func (r *registry) GetConversions(tr resource.Terraformed) []conversion.Conversion {
 	t := tr.GetTerraformResourceType()
-	if r == nil || r.provider == nil || r.provider.Resources[t] == nil {
+
+	p := r.providerCluster
+	if tr.GetNamespace() != "" {
+		p = r.providerNamespaced
+	}
```
negz, line 48: Is `GetConversions` always called on a real instance resource that exists in the API server? If so, `GetNamespace` will always have a value. If it's just being called on e.g. an empty `&v1beta.Instance{}` it won't.

negz (approved):
Nice work @jbw976! Few non-blocking style comments.

pkg/pipeline/run.go
negz: Not blocking, but maybe open a tracking issue to make sure we find a better approach post-preview.

pkg/controller/conversion/registry.go
jbw976: After tracing through how these conversions are used, they appear to all be on path for a controller-runtime generic conversion webhook handler and it's [Convertible interface](https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.18.2/pkg/conversion#Convertible) --> upjet conversion [RoundTrip()](https://github.com/crossplane-contrib/provider-upjet-aws/blob/main/apis/athena/v1beta1/zz_generated.conversion_spokes.go#L20) --> [GetConversions()](https://github.com/jbw976/upjet/blob/negz-spatial/pkg/controller/conversion/functions.go#L48)
--> [this namespace check](https://github.com/jbw976/upjet/blob/negz-spatial/pkg/controller/conversion/registry.go#L46).

So I do expect this object to be populated fully by the API server as opposed to a user calling it directly with an object only created locally and potentially missing information.

pkg/pipeline/run.go
jbw976: it's on my list of shortcuts and i'll make sure it's tracked in issues soon ✅

pkg/pipeline/run.go
jbw976: pushed https://github.com/crossplane/upjet/pull/480/commits/dc3707c290956fa9e04993c61a30bde00d67055e for this!

jbw976:
After integrating PR feedback, we have all green status checks and I did a quick sanity test of the latest in this PR with a `make generate` on provider-upjet-aws. I think this is good!

PR #515, opened 2025-07-25
Custom state check configuration for TF plugin framework resources

sergenyalcin:

We observed that a new region interceptor for the AWS TF Framework Plugin resources was introduced. This interceptor intervenes between API calls and the upjet fw controller and unconditionally injects the `region` into the state.

In the upjet fw controllers, we rely on the API request will return the uninterpreted state (the original form). If this state does not reflect the actual state, it will affect the decision-making processes of the upjet fw controller.

For example, in the `Observe` call of the upjet fw controller, we decide whether a resource exists by checking whether the returned state is null. However, as pointed out, the interceptor always injects the region field. So, every null check will return false, which means that this check will not work correctly. Also, the TF plugin framework's internal decisions rely on performing checks based on whether the state is null or not.

To handle this problem, we introduced a new configuration mechanism for framework resources. At this resource level, it will provide a generic structure for resolving this issue. The goal is to write custom state checks based on specific conditions at the provider and resource levels to determine whether the state is truly empty.

The main reasons for implementing this generically are to avoid moving provider—and resource-specific logic to upjet, to enable the creation of custom empty checkers at the resource level capable of handling the effects of different interceptors in specific situations, and to keep the scope of this change provider—and resource-specific.

I have:

Tested locally for a few TF plugin framework resources.

Screenshot 2025-07-25 at 15 47 37

Screenshot 2025-07-25 at 15 51 35

Screenshot 2025-07-25 at 15 52 22

Screenshot 2025-07-25 at 16 03 37

### DIFF
--- a/pkg/config/resource.go
+++ b/pkg/config/resource.go
@@ -519,11 +519,11 @@ type Resource struct {
 	// Terraform InstanceDiff is computed during reconciliation.
 	TerraformCustomDiff CustomDiff
 
-	// TerraformPluginFrameworkStateEmptyCheckFn allows customizing the logic
+	// TerraformPluginFrameworkIsStateEmptyFn allows customizing the logic
 	// for determining whether a Terraform Plugin Framework state value should
 	// be considered empty/nil for resource existence checks. If not set, the
 	// default behavior uses tfStateValue.IsNull().
-	TerraformPluginFrameworkStateEmptyCheckFn TerraformPluginFrameworkStateEmptyCheckFn
+	TerraformPluginFrameworkIsStateEmptyFn TerraformPluginFrameworkIsStateEmptyFn
 
 	// ServerSideApplyMergeStrategies configures the server-side apply merge
 	// strategy for the fields at the given map keys. The map key is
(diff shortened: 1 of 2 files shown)


pkg/config/resource.go
```diff
@@ -642,6 +650,13 @@ type CustomDiff func(diff *terraform.InstanceDiff, state *terraform.InstanceStat
 // the JSON tags and tfMap is obtained by using the TF tags.
 type ConfigurationInjector func(jsonMap map[string]any, tfMap map[string]any) error
 
+// TerraformPluginFrameworkStateEmptyCheckFn is a function that determines whether
+// a Terraform Plugin Framework state value should be considered empty/nil for the
+// purpose of determining resource existence. This allows providers to implement
+// custom logic to handle cases where the standard IsNull() check is insufficient,
+// such as when provider interceptors add fields like region to all state values.
+type TerraformPluginFrameworkStateEmptyCheckFn func(ctx context.Context, tfStateValue tftypes.Value, resourceSchema rschema.Schema) (bool, error)
```
erhancagirici: nit: from the function name, it might be a little ambiguous what this function returns as bool (like does it return true when the state is empty or filled). Maybe consider naming it as sth like
`TerraformPluginFrameworkIsStateEmptyFn`. IMHO, this way it is more idiomatic as a bool-returning function name and self-documenting for implementers.

erhancagirici (approved):
Thanks @sergenyalcin , overall looks good to me. I left a nit comment, which is not a blocker.

pkg/config/resource.go
sergenyalcin: Makes sense. Changed to the suggestion.

PR #535, opened 2025-10-09
Add config.Provider.ExampleManifestConfiguration to configure the example manifest generation pipeline

ulucinar:

Add `config.Provider.ExampleManifestConfiguration` to configure the example manifest generation pipeline at provider scope. Provider authors may use it to configure provider-level aspects of the example manifest generation pipeline as follows:

```go
...
pc := ujconfig.NewProvider([]byte(providerSchema), resourcePrefix, modulePath, []byte(providerMetadata),
		ujconfig.WithRootGroup("template.m.crossplane.io"),
		ujconfig.WithIncludeList(ExternalNameConfigured()),
		ujconfig.WithFeaturesPackage("internal/features"),
		ujconfig.WithDefaultResourceOptions(
			ExternalNameConfigurations(),
		),
		ujconfig.WithExampleManifestConfiguration(ujconfig.ExampleManifestConfiguration{
			ManagedResourceNamespace: "crossplane-system",
		}))
...
```
With the above configuration, generated example manifests for any namespace-scoped resources (and the secret references they have) will have the `crossplane-system` namespace.

Some further refactoring is needed in the example generation pipeline. We are planning to do so in follow-up PRs.

This is a backwards-compatible change meaning that if the provider authors do not override the default namespace, they will continue to use `upbound-system`.

I have:

Tested against `crossplane/upjet-provider-template` & `crossplane-contrib/provider-upjet-azure`.

### DIFF
--- a/pkg/pipeline/run.go
+++ b/pkg/pipeline/run.go
@@ -124,9 +124,9 @@ func (r *PipelineRunner) Run(pc *config.Provider) []string { //nolint:gocyclo
 		resourcesGroups[group][resource.Version][name] = resource
 	}
 
-	var exampleGeneratorOpts []examples.GeneratorOption
+	exampleGeneratorOpts := []examples.GeneratorOption{examples.WithCRDScope(r.Scope), examples.WithNamespace(pc.ExampleManifestConfiguration.ManagedResourceNamespace)}
 	if r.Scope == tjtypes.CRDScopeNamespaced {
-		exampleGeneratorOpts = append(exampleGeneratorOpts, examples.WithLocalSecretRefs(), examples.WithNamespacedExamples())
+		exampleGeneratorOpts = append(exampleGeneratorOpts, examples.WithLocalSecretRefs())
 	}
 	exampleGen := examples.NewGenerator(r.DirExamples, r.ModulePathAPIs, pc.ShortName, pc.Resources, exampleGeneratorOpts...)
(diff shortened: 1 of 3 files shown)


pkg/pipeline/run.go
```diff
@@ -124,9 +124,9 @@ func (r *PipelineRunner) Run(pc *config.Provider) []string { //nolint:gocyclo
 		resourcesGroups[group][resource.Version][name] = resource
 	}
 
-	var exampleGeneratorOpts []examples.GeneratorOption
+	exampleGeneratorOpts := []examples.GeneratorOption{examples.WithCRDScope(r.Scope), examples.WithNamespace(pc.ExampleManifestConfiguration.ManagedResourceNamespace)}
 	if r.Scope == tjtypes.CRDScopeNamespaced {
-		exampleGeneratorOpts = append(exampleGeneratorOpts, examples.WithLocalSecretRefs(), examples.WithNamespacedExamples())
+		exampleGeneratorOpts = append(exampleGeneratorOpts, examples.WithLocalSecretRefs())
```
erhancagirici, line 129: nit: just for the sake of completeness, no change needed.
With CRD scope, we can be opinionated and infer whether the secret refs should be local, and we won't need this extra option. On the other hand, having this option lets us control it explicitly.

ulucinar: Thanks @erhancagirici, agreed. Looks like we can do some refactoring in the pipeline. Planning to tackle with these as follow-ups.

PR #539, opened 2025-10-16
Add doc on managing CRD versions

### LINKED ISSUE #529: Resource Version update management
After a terraform provider version update, some terraform resource schemas have been updated (eg. new field is added).
For example, a CRD previously generated with upjet is in v1alpha1 version. I guess the new generation of this resource has to have a version update (eg. v1alpha2) since the new field added to the schema.
What is the procedure ?

jeanduplessis:

Takes the information from the discussion in https://github.com/crossplane/upjet/issues/529#issuecomment-3405840728 and creates a documentation page for guidance on managing CRD versions,.

Fixes #529

I have:

n/a

### DIFF
--- a/docs/managing-crd-versions.md
+++ b/docs/managing-crd-versions.md
@@ -57,8 +57,12 @@ version:
    - The existing version stays backward compatible
    - Old manifests continue to work without modification
 
-2. **Optional Fields with Defaults**: New fields that have default values
-   - Resources will function with or without the new field specified
+2. **Optional Fields with Defaults**: New fields that have default values in the
+   Terraform provider schema
+   - Note: Upjet does not generate default values in the CRD schema itself, even
+     when defaults exist in the Terraform schema
+   - Defaults are applied at runtime by the Terraform provider or cloud provider
+   - Resources will function with or without the new field specified in the manifest
 
 3. **Additional Enum Values**: Adding new allowed values to existing fields
    - Existing values remain valid
@@ -67,12 +71,16 @@ version:
 
 ### Non-Breaking Changes (Backward Compatible)
 
-When new fields are **optional** or have **default values**, no version bump is
-necessary. The existing API version remains backward compatible because:
+When new fields are **optional** or have **default values in the Terraform
+provider schema**, no version bump is necessary. The existing API version
+remains backward compatible because:
 
 - Old manifests that don't specify the new fields continue to work
-- The Terraform provider handles the absence of these fields gracefully
-- Late initialization can populate sensible defaults
+- The Terraform provider handles the absence of these fields gracefully at
+  runtime (applying defaults when needed)
+- Late initialization can populate sensible defaults from the provider state
+- Note: While defaults exist in the Terraform schema, they are not reflected in
+  the generated CRD schema itself
 
 **Example**: Adding an optional `encryption_enabled` field to a database
 resource doesn't break existing manifests that don't specify this field.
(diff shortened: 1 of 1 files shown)


docs/managing-crd-versions.md
```diff
@@ -0,0 +1,296 @@
...
+### Changes That Do NOT Require Version Bumps
+
+The following changes maintain backward compatibility and do not require a new
+version:
+
+1. **Optional New Fields**: Adding new optional fields to the schema
+   - The existing version stays backward compatible
+   - Old manifests continue to work without modification
+
+2. **Optional Fields with Defaults**: New fields that have default values
```
sergenyalcin: In the upjet-generated APIs, we don't generate default values for the fields. I mean, even the TF schema has a default value, which you will not see in the CRD schema. In the runtime, it may still behave with defaulting due to the TF provider's or remote-cloud provider's implementation. However, in the MR manifest in the cluster, you will not see this default value in the corresponding field (before late-init or observe).

Let's give a concrete example. Let's consider there is an X resource in the TF side and there is a f field. This is optional and has a default value.

In the XP/Upjet schema, the X.f field will be generated as optional and without a default value (the reason mentioned). When you try to apply an MR to the cluster, the X.f field will not be filled by k8s-api-server because the CRD schema doesn't have the default value. However, in runtime, during the underlying TF calls, it may be filled before sending the request to the cloud provider because this field has a default in the TF schema.

We may consider this differentiation to clarify the matter. It can be understood as CRD schema defaulting since the doc's context is this.

docs/managing-crd-versions.md
```diff
@@ -0,0 +1,296 @@
...
+   - Resources will function with or without the new field specified
+
+3. **Additional Enum Values**: Adding new allowed values to existing fields
+   - Existing values remain valid
+
+## Breaking vs. Non-Breaking Changes
+
+### Non-Breaking Changes (Backward Compatible)
+
+When new fields are **optional** or have **default values**, no version bump is
```
sergenyalcin: Please see the above comment about defaulting.

docs/managing-crd-versions.md
```diff
@@ -0,0 +1,296 @@
...
+        // Previous versions to maintain
+        r.PreviousVersions = []string{"v1alpha2", "v1alpha1"}
+
+        // Additional configuration for conversions (see below)
+    })
+}
+```
+
+This configuration will:
+1. Generate CRDs for all three versions: `v1alpha1`, `v1alpha2`, and `v1beta1`
```
sergenyalcin: This does not appear entirely accurate. Upjet only generates the latest (current) version. In there r.Version's value is `v1beta1`.

docs/managing-crd-versions.md
```diff
@@ -0,0 +1,296 @@
...
+2. Mark `v1beta1` as the storage version (hub)
```
sergenyalcin: This does not appear entirely accurate. We can clarify the following:
- Hub-Spoke version terminology is different from the storage version. I mean, the hub version and storage version can be different.
- We may consider adding a statement that may be related to storage version, and hub version by default set to r.Version. But they are configurable with the following configuration options:
  - `SetCRDHubVersion`
  - `SetCRDStorageVersion`

docs/managing-crd-versions.md
```diff
@@ -0,0 +1,296 @@
...
+
+### Hub-and-Spoke Pattern
+
+Kubernetes uses a "hub-and-spoke" conversion model:
+
+```
+v1alpha1 ←→ v1beta1 (hub) ←→ v1alpha2
+```
+
+- **Hub**: The storage version (usually the latest stable version)
```
sergenyalcin: As mentioned, the hub version may not be the same as the storage version. In practice and in the most reasonable context, this is the most efficient way. However, they are distinct concepts and may differ.

sergenyalcin (commented):
Thanks, @jeanduplessis, for this doc. I left a few comments to clarify some points.

sergenyalcin (approved):
Thanks @jeanduplessis LGTM!

PR #544, opened 2025-10-24
Add per-field InitProvider tag & kubebuilder marker overrides

ulucinar:

This PR proposes the following changes and refactorings:
- Introduce `pkg/types/structtag` with a typed struct tag model (Value, parsing, override, no-omit)
- Add per-field InitProvider overrides:
  - `config.InitProviderOverrides` with `kubebuilder.Options` and `config.TagOverrides`
    - Tag overrides for `json` and `tf` via `structtag.Value`
    - Kubebuilder marker overrides via `pkg/types/markers/kubebuilder.Options`
- Move `KubebuilderOptions` into `pkg/types/markers/kubebuilder` to prevent import cycles with the `config` package
- Rename Terrajet markers file to Upjet markers file (terrajet.go → upjet.go)
- Refactor tag handling to use `structtag.Value` consistently for the ForProvider, InitProvider and Observation API generation
- Strengthen validation of JSON tag names and tag combinations, previously there was no validation for these.
  - `json` struct tag values are validated using the regex `^[a-zA-Z_][a-zA-Z0-9_]*$` to prevent upjet from generating invalid CRD field names.
  - `tf` tag values are currently not validated. This is future work (we need to investigate their syntax first)
- Treat `nil` values in `initProvider` as absent when computing ignore list; avoids spurious plan diffs

We are planning to follow up this PR by deprecating the `markers.UpboundOptions` API in favor of a `config.SchemaElementOption` based one.

**Breaking changes:**

*Behavior*:
- `nil` initProvider fields no longer produce ignore-change markers. Although this is a behavior change, because the Go struct tags are generated using the `omitempty` specifier, this was not observable previously.
- `json` struct tag values are now validated and invalid configurations will now cause panic early.

*API*:
- `markers.KubebuilderOptions` has been moved to `kubebuilder.Options` to prevent import cycles with the `config` package. Not expected to cause huge migration efforts as validated in the official provider repositories.
- `config.SchemaElementOptions.AddToObservation()` has been removed due to its confusing naming. Again, not expected to cause migration issues for provider authors.

I have:

Validated across `crossplane-contrib/provider-upjet-{azure/ad, gcp, aws}` and observed no diffs in the generated code artifacts. Also validated with `crossplane-contrib/provider-upjet-azure` by overriding the `network.Subnet`'s `delegation` and `serviceEndpoints` Kubebuilder markers and the generated `InitProvider` struct tags.

https://github.com/crossplane-contrib/provider-upjet-azure/pull/1096 also consumes these changes and applies the `network.Subnet` configuration below to enable the new behaviour:
```go
		// delegation schema element options.
		r.SchemaElementOptions.SetInitProviderOverrides("delegation", &config.InitProviderOverrides{
			// remove `,omitempty` from spec.initProvider.delegation struct tags.
			TagOverrides: config.TagOverrides{
				TFTag:   structtag.MustParseTF(``),
				JSONTag: structtag.MustParseTF(``),
			},
			// mark spec.initProvider.delegation as optional.
			KubebuilderOptions: &kubebuilder.Options{
				Required: ptr.To(false),
			},
		})
```

### DIFF
--- a/pkg/config/resource.go
+++ b/pkg/config/resource.go
@@ -692,6 +692,13 @@ func (m SchemaElementOptions) SetAddToObservation(el string) {
 	m[el].AddToObservation = true
 }
 
+// AddToObservation returns true if the schema element at the specified path
+// should be added to the CRD type's Observation type.
+// Deprecated: Use SchemaElementOptions.AddToObservation instead.
+func (m SchemaElementOptions) AddToObservation(el string) bool {
+	return m[el] != nil && m[el].AddToObservation
+}
+
 // TFListConversionPaths returns the Resource's runtime Terraform list
 // conversion paths in fieldpath syntax.
 func (r *Resource) TFListConversionPaths() []string {
--- a/pkg/types/structtag/value.go
+++ b/pkg/types/structtag/value.go
@@ -16,8 +16,8 @@ const (
 	errFmtInvalidJSONTagName = "invalid JSON struct tag name: %q (must match %s)"
 	errFmtInlineWithName     = "invalid struct tag: cannot set the name %q with inline"
 	errFmtInlineWithOmit     = "invalid struct tag: cannot set inline with omit option %q"
-	errFmtInvalidOptions     = "invalid struct tag format: %q (cannot combine %q with other options)"
 	errFmtInvalidCombination = "invalid struct tag format: %q (cannot combine %s with %s)"
+	errFmtInvalidOption      = "invalid struct tag option: %q"
 )
 
 // reJSONName matches valid Kubernetes CRD field names.
@@ -142,7 +142,7 @@ func parse(key Key, value string) (*Value, error) { //nolint:gocyclo // easier t
 	name := strings.TrimSpace(parts[0])
 	if name == string(OmitAlways) {
 		if len(parts) > 1 {
-			return nil, errors.Errorf(errFmtInvalidOptions, value, OmitAlways)
+			return nil, errors.Errorf(errFmtInvalidCombination, value, OmitAlways, "other options")
 		}
 		// no name expected for "-" (OmitAlways).
 		v.omit = OmitAlways
@@ -164,7 +164,7 @@ func parse(key Key, value string) (*Value, error) { //nolint:gocyclo // easier t
 		case "inline":
 			v.inline = true
 		default:
-			// ignore unknown options for forward compatibility.
+			return nil, errors.Errorf(errFmtInvalidOption, option)
 		}
 	}
 
--- a/pkg/types/structtag/value_test.go
+++ b/pkg/types/structtag/value_test.go
@@ -111,16 +111,12 @@ func TestParseJSON(t *testing.T) {
 			},
 		},
 		"UnknownOption": {
-			reason: "Unknown options in a JSON tag should be silently ignored while known options are parsed.",
+			reason: "Unknown options in a JSON tag should return an error",
 			args: args{
 				value: "fieldName,omitempty,unknownopt",
 			},
 			want: want{
-				v: &Value{
-					key:  KeyJSON,
-					name: "fieldName",
-					omit: OmitEmpty,
-				},
+				err: errors.New(`invalid struct tag option: "unknownopt"`),
 			},
 		},
 		"ErrorOmitAlwaysWithOptions": {
@@ -129,7 +125,7 @@ func TestParseJSON(t *testing.T) {
 				value: "-,omitempty",
 			},
 			want: want{
-				err: errors.New(`invalid struct tag format: "-,omitempty" (cannot combine "-" with other options)`),
+				err: errors.New(`invalid struct tag format: "-,omitempty" (cannot combine - with other options)`),
 			},
 		},
 		"ErrorNameWithInline": {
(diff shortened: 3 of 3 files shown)


pkg/types/structtag/value.go
```diff
@@ -0,0 +1,284 @@
...
+		opt(v)
+	}
+
+	// validate name if set for JSON tags.
+	if key == KeyJSON && len(v.name) > 0 && !reJSONName.MatchString(v.name) {
+		panic(fmt.Sprintf(errFmtInvalidJSONTagName, v.name, reJSONName.String()))
+	}
+	// make sure name and inline are not both set.
+	if v.inline && v.name != "" {
+		panic(fmt.Sprintf("invalid struct tag: cannot set the name %q with inline", v.name))
```
sergenyalcin: nit: Similar to the constant error message (`errFmtInvalidJSONTagName`) usage in line 104, what do you think about using constants for error messages for this?

pkg/types/structtag/value.go
```diff
@@ -0,0 +1,284 @@
...
+	}
+	// make sure omit and inline are not both set.
+	if v.inline && v.omit != NotOmitted {
+		panic(fmt.Sprintf("invalid struct tag: cannot set inline with omit option %q", v.omit))
```
sergenyalcin: nit: Similar to the constant error message (`errFmtInvalidJSONTagName`) usage in line 104, what do you think about using constants for error messages for this?

pkg/types/structtag/value.go
```diff
@@ -0,0 +1,284 @@
...
+
+	for i := 1; i < len(parts); i++ {
+		option := strings.TrimSpace(parts[i])
+		switch option {
+		case string(OmitEmpty):
+			v.omit = OmitEmpty
+		case "inline":
+			v.inline = true
+		default:
+			// ignore unknown options for forward compatibility.
```
sergenyalcin: A question: Do you think silent ignoring of unknown options might mask typos? Do you consider logging unknown options? I mean, if a user unintentionally writes `inine` (without using `l`), this will be silently ignored. As you pointed out in the code comment, I think this is appropriate in terms of compatibility. I believe not taking an action is fair. However, logging may be worth considering.

pkg/config/resource.go
```diff
@@ -690,12 +692,6 @@ func (m SchemaElementOptions) SetAddToObservation(el string) {
 	m[el].AddToObservation = true
 }
 
-// AddToObservation returns true if the schema element at the specified path
-// should be added to the CRD type's Observation type.
-func (m SchemaElementOptions) AddToObservation(el string) bool {
```
sergenyalcin, line 695: I agree with this statement in the PR description about this code change:

> config.SchemaElementOptions.AddToObservation() has been removed due to its confusing naming. Again, not expected to cause migration issues for provider authors.

However, since this is part of our public configuration API (although I didn't observe the usage of the big three providers, and it will not have a significant effect), we may consider deprecating it first and removing it in the next releases.

pkg/types/builder.go
```diff
@@ -402,32 +401,30 @@ func (r *resource) addParameterField(f *Field, field *types.Var) {
...
 	// For removing omitempty tag from json tag, we are just checking if the field is required by the schema.
 	if requiredBySchema {
 		// Required fields should not have omitempty tag in json tag.
-		// TODO(muvaf): This overrides user intent if they provided custom
-		// JSON tag.
-		r.paramTags = append(r.paramTags, fmt.Sprintf(`json:"%s" tf:"%s"`, strings.TrimSuffix(f.JSONTag, ",omitempty"), f.TFTag))
+		r.paramTags = append(r.paramTags, fmt.Sprintf("%s %s", f.JSONTag.NoOmit(), f.TFTag))
 	} else {
-		r.paramTags = append(r.paramTags, fmt.Sprintf(`json:"%s" tf:"%s"`, f.JSONTag, f.TFTag))
+		r.paramTags = append(r.paramTags, fmt.Sprintf("%s %s", f.JSONTag, f.TFTag))
```
sergenyalcin, line 415: nit: As far as I understand, the TagOverrides mechanism is generic, but currently, they are usable only for `InitProvider`. As a follow-up improvement, do you think adding this to `ForProvider` and `AtProvider` is a suitable approach?

sergenyalcin (commented):
Thank you, @ulucinar, for this PR. I have left a few comments for your consideration. I didn't identify any significant issues or obstacles. I will happily approve it if you think it is ready to proceed in its current state.

pkg/types/structtag/value.go
ulucinar: Done. Now we are using error string constants consistently in the file.

pkg/types/structtag/value.go
ulucinar: Done.

pkg/types/structtag/value.go
ulucinar: Done. Thank you, I think this is a very valid point. Unknown options will now result in an error. I don't think we really need to be tolerant to unknown options as we control the JSON library and we are opinionated about the generated Go structs.

pkg/config/resource.go
ulucinar: Done. Added the function back with a deprecation notice. This is an internal utility probably not used elsewhere but it makes sense to obey the API lifecycle management hygiene :) Maybe we should also consider a larger refactoring for future to move similar things to internal packages.

pkg/types/builder.go
ulucinar: Yes, that's exactly the idea. This refactoring in the code generation pipeline will allow us to be able to individually control the generated struct tags at different parts of the API. We had also better start using different field objects for initProvider, forProvider and atProvider. Currently the generation code modifies the same field object in place and passes it down to the subpipeline of these APIs. It's becoming more and more difficult to maintain the code generation pipelines.

sergenyalcin (approved):
Thanks @ulucinar LGTM!

PR #545, opened 2025-10-24
Fix the race on the registered conversion paths in conversion.Convert

ulucinar:

Multiple Go routines can race on the registered conversion paths when in-place sorts are performed in `conversion.Convert`. Although we could not reproduce these issues like the one reported in https://github.com/crossplane-contrib/provider-upjet-aws/issues/1714, observations done by @jonathan-innis [here](https://github.com/crossplane-contrib/provider-upjet-aws/issues/1714#issuecomment-3408035762) using a provider package with the fix proposed in this PR suggest that this data race could be the underlying reason behind similar failures.

This PR fixes the data race by making a copy of the list of the registered conversion paths before they are sorted in-place.

I have:

We have tested the data race fix using the provider package `xpkg.upbound.io/upbound/provider-aws-lambda:v1.20.5-1.gf4470285b`. There are logging differences between that implementation and the one proposed here. The test image has more verbose logging, which is not suggested with this PR due to concerns with PII.

### DIFF
--- a/pkg/config/conversion/list_conversion.go
+++ b/pkg/config/conversion/list_conversion.go
@@ -90,7 +90,9 @@ type ConvertOptions struct {
 // an embedded object will be converted into a singleton list or a singleton
 // list will be converted into an embedded object) is determined by the mode
 // parameter.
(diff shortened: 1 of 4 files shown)


sergenyalcin (approved):
Thank you very much @ulucinar for catching and fixing this. LGTM!

PR #549, opened 2025-11-01
Add config.FrameworkResourceWithComputedIdentifier external-name configuration

ulucinar:
Relevant PR: https://github.com/crossplane-contrib/provider-upjet-aws/pull/1900

This PR adds a new external-name configuration `config.FrameworkResourceWithComputedIdentifier` to be used with the Terraform plugin framework resources with a computed identifier field. It also introduces `config.ExternalName.TFPluginFrameworkOptions` struct type for logically grouping framework resource related external-name configuration options. `TFPluginFrameworkOptions.ComputedIdentifierAttributes` can be used to exclude the configured computed identifier fields while calculating drifts between the desired and actual states.

This PR also fixes an existing bug in `config.ExternalNameFrom` so that it now properly exhibits the parent configuration's behavior.

I have:

Manually tested against `crossplane-contrib/provider-upjet-aws`.

### DIFF
--- a/pkg/config/resource.go
+++ b/pkg/config/resource.go
@@ -181,6 +181,20 @@ type ExternalName struct {
 	// "ignore" the supplied diagnostics, i.e. they
 	// correspond to "resource not found".
 	IsNotFoundDiagnosticFn func(diags []*tfprotov6.Diagnostic) bool
+
+	// TFPluginFrameworkOptions represents options related to Terraform plugin
+	// framework resources.
+	TFPluginFrameworkOptions TFPluginFrameworkOptions
+}
+
+// TFPluginFrameworkOptions are external-name configuration options that
+// are specific to Terraform plugin framework resources.
+type TFPluginFrameworkOptions struct {
+	// ComputedIdentifierAttributes is the list of computed Terraform identifier
+	// attribute names for a framework resource. When set,
+	// these computed identifier attributes will be ignored from the desired
+	// state when calculating the drifts between the desired and actual states.
+	ComputedIdentifierAttributes []string
 }
 
 // References represents reference resolver configurations for the fields of a
--- a/pkg/controller/external_tfpluginfw.go
+++ b/pkg/controller/external_tfpluginfw.go
@@ -8,6 +8,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"maps"
 	"math"
 	"math/big"
 	"strings"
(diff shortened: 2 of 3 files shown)


pkg/config/resource.go
```diff
@@ -181,6 +181,20 @@ type ExternalName struct {
...
+}
+
+// TFPluginFrameworkOptions are external-name configuration options that
+// are specific to Terraform plugin framework resources.
+type TFPluginFrameworkOptions struct {
+	// ComputedIdentifierName is the computed Terraform identifier attribute
+	// name for a framework resource.
+	// When set, ExternalName.SetIdentifierArgumentFn will
+	// conditionally set this attribute from the external-name annotation value.
+	ComputedIdentifierName *string
```
erhancagirici: nit: I cannot find a concrete resource example but in theory there could be multiple computed identifiers. We can maybe consider representing this with a slice.

pkg/config/resource.go
```diff
@@ -181,6 +181,20 @@ type ExternalName struct {
 	// "ignore" the supplied diagnostics, i.e. they
 	// correspond to "resource not found".
 	IsNotFoundDiagnosticFn func(diags []*tfprotov6.Diagnostic) bool
```
erhancagirici, line 183: nit: This option is also framework-specific, we can consider moving it to `TFPluginFrameworkOptions` too.

pkg/config/resource.go
```diff
@@ -181,6 +181,20 @@ type ExternalName struct {
...
+
+	// TFPluginFrameworkOptions represents options related to Terraform plugin
+	// framework resources.
+	TFPluginFrameworkOptions TFPluginFrameworkOptions
```
erhancagirici, line 187: nit: WDYT about inlining this?

pkg/controller/external_tfpluginfw.go
```diff
@@ -313,9 +314,13 @@ func (n *terraformPluginFrameworkExternalClient) filteredDiffExists(rawDiff []tf
...
 // to be recreated) an error is returned as Crossplane Resource Model (XRM)
 // prohibits resource re-creations and rejects this plan.
-func (n *terraformPluginFrameworkExternalClient) getDiffPlanResponse(ctx context.Context,
-	tfStateValue tftypes.Value) (*tfprotov6.PlanResourceChangeResponse, bool, error) {
-	tfConfigDynamicVal, err := protov6DynamicValueFromMap(n.params, n.resourceValueTerraformType)
+func (n *terraformPluginFrameworkExternalClient) getDiffPlanResponse(ctx context.Context, tfStateValue tftypes.Value) (*tfprotov6.PlanResourceChangeResponse, bool, error) {
+	params := maps.Clone(n.params)
+	// if a computed identifier has been configured, remove it from config.
+	if n.config.ExternalName.TFPluginFrameworkOptions.ComputedIdentifierName != nil {
+		delete(params, *n.config.ExternalName.TFPluginFrameworkOptions.ComputedIdentifierName)
```
erhancagirici: note: `ComputedIdentifierName` should always be a purely computed field name. `Optional+Computed` fields can still be provided from config. I am assuming that such identifiers are very rare, but still possible in theory.

pkg/config/resource.go
ulucinar: Agreed. It's possible that there are multiple identifier attributes for a (framework) resource. For example for an AWS VPC (although currently not a framework resource), both the region and the provider-assigned (computed) ID could be identifier data.

The region for the VPC would also be a required or optional parameter for the resource and this configuration API does not target such optional parameters. We had better not filter out optional parameters when calculating the drift. 

Although I've not seen any concrete example of this, in theory, there could also be multiple computed identifier attributes, like `id1` and `id2`, which are both assigned by the provider probably because the CSP API assigns these two distinct attributes during creation instead of returning a single identifier back. If we observe something like this in the wild, then we will also need to consider how to format those multiple computed attributes into an XRM external-name back and forth. 

What I propose for now (before adding that complexity until we observe that it's actually needed) is:
- Change the API to a slice as you propose so that the API will not change if we encounter framework resources with multiple computed identifier attributes
- Not introduce the extra configuration options needed to handle multiple computed identifier attributes for now
- Document it as an error if multiple computed identifiers are configured. In the future, we can consider making this a build-time error.

pkg/config/resource.go
ulucinar: I did not want to move it to the newly introduced `config.TFPluginFrameworkOptions` struct because:
- Not to introduce breaking changes to this PR (currently, it's backward compatible), although I would still be okay with this...
- More importantly, I'm not sure if this configuration actually belongs to external-name configuration.

pkg/config/resource.go
ulucinar: I would prefer keeping this not-inlined because these options are meant to be used by only a subset of Terraform plugin framework resources. This organization makes it more clear to those who are writing resource configurations that options introduced with this struct are specific to framework resources.

pkg/controller/external_tfpluginfw.go
ulucinar: Yes, the target configuration for these changes are only computed identifier fields for framework resources. If it's optional or required config, then we had better not filter them out from the diffs.

pkg/config/resource.go
erhancagirici: Agreed with the explanation and the approach 👍

PR #555, opened 2025-11-10
framework-external-client: Flush reconstructed states for recalculation after Observation failures

erhancagirici:

Marks reconstructed states while saving them and flushes reconstructed states after Observation errors.

Background:
In the initial reconciliation of the MRs, just after being created in the k8s api, the very first `Connect()` constructs a (partial) TF prior state using the current configuration values, and saves it to the opTracker state cache. This reconstructed state contains enough information to identify the external resource, and facilitates the TF resource read operation in the subsequent `Observe()` call. 
- After the successful initial observation, this state is immediately updated with the latest full state data from the resource read (empty state for a non-existent external resource). Subsequent reconciles do not reconstruct the state anymore.
- If the resource read fails, an error is returned without touching the constructed prior state and the reconcile is requeued.

This causes an issue for the following scenario:
- An MR manifest is applied with some invalid parameter values
- in the initial `Connect()` a prior state is constructed with those values and saved to state cache.
- the initial observation fails and returns an error. The state cache is not modified.
- State cache has still the "invalid" value.
- MR manifest is updated with the correct value.
- In the next reconciliation, the state is not reconstructed, because a prior state already exists.
- The observation/resource read fails again because the invalid prior state.
- Resource is stuck, until a provider pod restart.

In this change:
- while recording the TF states into the cache, attach a meta-information regarding whether this state is a "reconstructed" state or a regular one (that comes from a TF resource CRUD operation).
- After observation failures, flush reconstructed states, so that they are rebuilt, with new configuration values if any.
This change also refactors the logic for TF resource config's corresponding`tftypes.Value` calculation into a central place during `Connect()`. Also, clears any read-only attribute from the config if any. Those should not exist in the configuration.

Fixes #

I have:

tested manually with provider-upjet-aws `PodIdentityAssociation` MR

Create a `PodIdentityAssociation` with an invalid role arn
```yaml
apiVersion: eks.aws.upbound.io/v1beta1
kind: PodIdentityAssociation
metadata:
  name: example-pia
spec:
  forProvider:
    clusterName: eks-cluster-test
    namespace: example
    region: us-west-1
    roleArn: foo-invalid-arn
    serviceAccount: default
```

The observation fails, stating that the `roleArn` is invalid.

Edit the resource and set a valid role ARN

```yaml
apiVersion: eks.aws.upbound.io/v1beta1
kind: PodIdentityAssociation
metadata:
  annotations:
    meta.upbound.io/example-id: eks/v1beta1/podidentityassociation
  labels:
    testing.upbound.io/example-name: example-pia
  name: example-pia
spec:
  forProvider:
    clusterName: eks-cluster-test
    namespace: example
    region: us-west-1
    roleArn: arn:aws:iam::123456789012:role/example-pia
    serviceAccount: default
```

The change is picked up and the resource gets ready

### DIFF
--- a/pkg/controller/external_tfpluginfw.go
+++ b/pkg/controller/external_tfpluginfw.go
@@ -939,7 +939,7 @@ func getFatalDiagnostics(diags []*tfprotov6.Diagnostic) error {
 func protov6DynamicValueFromMap(data map[string]any, terraformType tftypes.Type) (*tfprotov6.DynamicValue, error) {
 	tfValue, err := tfValueFromMap(data, terraformType)
 	if err != nil {
-		return nil, err
+		return nil, errors.Wrap(err, "cannot construct TF value from raw map")
 	}
 	dynamicValue, err := tfprotov6.NewDynamicValue(terraformType, tfValue)
 	if err != nil {


pkg/controller/external_tfpluginfw.go
```diff
@@ -140,17 +142,6 @@ func getFrameworkExtendedParameters(ctx context.Context, tr resource.Terraformed
...
 
-	// we need to parameterize the following for a provider
-	// not all providers may have this attribute
-	// TODO: tags-tags_all implementation is AWS specific.
-	// Consider making this logic independent of provider.
-	// TODO(erhan): this might be already resolved with proposed state impl
-	_, hasTags := fwResSchema.Attributes["tags"]
-	_, hasTagsAll := fwResSchema.Attributes["tags_all"]
-	if hasTags && hasTagsAll {
-		params["tags_all"] = params["tags"]
```
sergenyalcin, line 151: For confirmation, are we removing this part because it has been resolved based on the proposed state implementation?

pkg/controller/external_tfpluginfw.go
```diff
@@ -906,24 +937,31 @@ func getFatalDiagnostics(diags []*tfprotov6.Diagnostic) error {
 // protov6DynamicValueFromMap constructs a protov6 DynamicValue given the
 // map[string]any using the terraform type as reference.
 func protov6DynamicValueFromMap(data map[string]any, terraformType tftypes.Type) (*tfprotov6.DynamicValue, error) {
-	jsonBytes, err := json.Marshal(data)
+	tfValue, err := tfValueFromMap(data, terraformType)
 	if err != nil {
-		return nil, errors.Wrap(err, "cannot marshal json")
+		return nil, err
```
sergenyalcin: I think you may want to consider wrapping this error, as it is a convention in Crossplane repositories.

sergenyalcin (approved):
Thanks @erhancagirici LGTM! Just requested one error wrapping. I believe that after addressing this, we can merge the PR. Thank you for handling this problem.

pkg/controller/external_tfpluginfw.go
erhancagirici: thanks! done ✅

pkg/controller/external_tfpluginfw.go
erhancagirici: yes, I've tested adding/removing/modifying existing tags/importing for FW resources and all succeeded. Also step-debugged the TF's tag interceptor code to validate it behaves properly.
Few additional motivations:
- As a matter of fact, IIRC this was borrowed from the TF SDK external client during FW client impl. For TF FW external client, this already might have worked properly without this hack. For TF SDK this might still be needed.
- `tags_all` is always a read-only (purely computed) attribute.  It should not be a part of the TF config value anyways. This can potentially cause superfluous diff behavior. 
- FW client's plan phase correctly handles computed value behavior.

PR #563, opened 2025-12-23
Adding new conversion functions and handling their runtime aspects

sergenyalcin:

This PR introduces new conversion functions for handling API version changes and implements runtime mechanisms to prevent data loss when working with controllers running older API versions.

### 1. New Conversion Functions (`pkg/config/conversion/conversions.go`)

**Newly Introduced Field Converter** (`newlyIntroducedFieldConverter`)
- Handles fields that exist in some API versions but not others
- Two modes:
  - `ToAnnotation`: Stores field value in annotation when converting to an older version without the field
  - `FromAnnotation`: Restores field value from annotation when converting to a newer version with the field
- Usage: `NewNewlyIntroducedFieldConversion(sourceVersion, targetVersion, fieldPath, mode)`

**Field Type Converter** (`fieldTypeConverter`)
- Handles primitive type changes between API versions
- Supported conversions:
  - `IntToString`/`StringToInt`: int64 ↔ string
  - `BoolToString`/`StringToBool`: bool ↔ string
  - `FloatToString`/`StringToFloat`: float64 ↔ string
- Usage: `NewFieldTypeConversion(sourceVersion, targetVersion, fieldPath, mode)`

Both converters use the `internal.upjet.crossplane.io/field-conversions` annotation prefix to store temporary values.

### 2. Runtime Spec Annotation Merge

**`MergeAnnotationFieldsWithSpec` function**
- Called during parameter preparation before passing to Terraform
- Merges annotation values (prefixed with `internal.upjet.crossplane.io/field-conversions`) back into `spec.forProvider` and `spec.initProvider`
- Handles both camelCase and snake_case field names
- Prevents data loss when controller version is old and newly added fields are stored in annotations

### 3. Runtime Status Annotation Move

**`MoveTFStateValuesToAnnotation` function**
- Called after receiving state from Terraform during Observe, Create, and Update operations
- Compares TF state fields with CRD status.atProvider fields
- If a field exists in TF state but not in status.atProvider (old API version), stores it in an annotation
- Prevents status data loss during TF state → CRD status conversion

**Configuration:**
- New `TfStatusConversionPaths []string` field in `config.Resource`
- Allows providers to specify which status fields should be moved to annotations for backwards compatibility
- Paths should be in the format `status.atProvider.fieldName`

## Use Cases

**Scenario 1: New field added to API v1beta2**
- User creates resource using v1beta2 with the new field
- Controller running v1beta1 receives the resource
- Field is stored in annotation during v1beta2→v1beta1 conversion
- `MergeAnnotationFieldsWithSpec` restores the field value before passing to Terraform
- User's configuration is preserved

**Scenario 2: New status field returned by Terraform**
- Terraform returns state with a new field
- Controller running old API version doesn't have this field in status schema
- `MoveTFStateValuesToAnnotation` stores the field in annotation

**Scenario 3: Field type change between versions**
- API v1beta1 has field as int64, v1beta2 changes it to string
- Type converter handles the conversion during API version transitions
- No manual intervention needed

```text
NOTE: config.Resource.ControllerReconcileVersion field is deprecated here. Please see the discussions below about the details.
```

I have:

Tested locally by using the provider-upjet-azuread Applications.Application resource. For testing, this resource created a "fictional" difference between v1beta1 and v1beta2 versions. Validated:
- Resource's lifecycle: create, update, import, and delete
- Conversions
- Annotation-based strategy
- New runtime aspects like merging the annotation values with spec and moving status the annotation fields.

Here are examples for new field conversion registrations:

```go
// Field conversions
r.Conversions = append(r.Conversions, conversion.NewOptionalFieldConversion("v1beta1", "v1beta2", "spec.forProvider.displayName", conversion.FromAnnotation))
r.Conversions = append(r.Conversions, conversion.NewOptionalFieldConversion("v1beta2", "v1beta1", "spec.forProvider.displayName", conversion.ToAnnotation))
r.Conversions = append(r.Conversions, conversion.NewOptionalFieldConversion("v1beta1", "v1beta2", "spec.initProvider.displayName", conversion.FromAnnotation))
r.Conversions = append(r.Conversions, conversion.NewOptionalFieldConversion("v1beta2", "v1beta1", "spec.initProvider.displayName", conversion.ToAnnotation))
r.Conversions = append(r.Conversions, conversion.NewOptionalFieldConversion("v1beta1", "v1beta2", "status.atProvider.displayName", conversion.FromAnnotation))
r.Conversions = append(r.Conversions, conversion.NewOptionalFieldConversion("v1beta2", "v1beta1", "status.atProvider.displayName", conversion.ToAnnotation))

// This is for not losing the status data. It is explained in the PR description.
r.TfStatusConversionPaths = []string{"status.atProvider.displayName"}

// The better solution is to make the API version and the controller version compatible with the following configuration. This will automatically resolve the status data loss since the controller versioned API and the TF API are compatible.
// r.ControllerReconcileVersion = "v1beta2"
```

### DIFF
--- a/pkg/controller/external_tfpluginfw.go
+++ b/pkg/controller/external_tfpluginfw.go
@@ -118,7 +118,7 @@ func getFrameworkExtendedParameters(ctx context.Context, tr resource.Terraformed
 			return nil, errors.Wrap(err, "cannot get merged parameters")
 		}
 	} else {
-		params, err = MergeAnnotationFieldsWithSpec(tr, initParamsMerged, tr.GetAnnotations())
+		params, err = mergeAnnotationFieldsWithSpec(tr, initParamsMerged, tr.GetAnnotations())
 		if err != nil {
 			return nil, errors.Wrap(err, "cannot merge annotation fields")
 		}
@@ -200,7 +200,7 @@ func (c *TerraformPluginFrameworkConnector) Connect(ctx context.Context, mg xpre
 		if err != nil {
 			return nil, errors.Wrap(err, "failed to get the observation")
 		}
-		if err := MergeAnnotationFieldsWithStatus(tfState, mg.GetAnnotations(), c.config); err != nil {
+		if err := mergeAnnotationFieldsWithStatus(tfState, mg.GetAnnotations(), c.config); err != nil {
 			return nil, errors.Wrapf(err, "failed to merge annotations on resource %q", client.ObjectKeyFromObject(mg))
 		}
 		tfState, err = c.config.ApplyTFConversions(tfState, config.ToTerraform)
@@ -628,7 +628,7 @@ func (n *terraformPluginFrameworkExternalClient) Observe(ctx context.Context, mg
 			return managed.ExternalObservation{}, errors.Wrap(err, "could not get observation")
 		}
 		annotations := mg.GetAnnotations()
-		annotationUpdate, err := MoveTFStateValuesToAnnotation(stateValueMap, obs, annotations, n.config)
+		annotationUpdate, err := moveTFStateValuesToAnnotation(stateValueMap, obs, annotations, n.config)
 		if err != nil {
 			return managed.ExternalObservation{}, errors.Wrap(err, "cannot move status values to annotation")
 		}
@@ -732,7 +732,7 @@ func (n *terraformPluginFrameworkExternalClient) Create(ctx context.Context, mg 
 		return managed.ExternalCreation{}, errors.Wrap(err, "could not get observation")
 	}
 	annotations := mg.GetAnnotations()
-	if annotationUpdate, err := MoveTFStateValuesToAnnotation(stateValueMap, obs, annotations, n.config); err != nil {
+	if annotationUpdate, err := moveTFStateValuesToAnnotation(stateValueMap, obs, annotations, n.config); err != nil {
 		return managed.ExternalCreation{}, errors.Wrap(err, "cannot move status values to annotation")
 	} else if annotationUpdate {
 		mg.SetAnnotations(annotations)
@@ -816,7 +816,7 @@ func (n *terraformPluginFrameworkExternalClient) Update(ctx context.Context, mg 
 		return managed.ExternalUpdate{}, errors.Wrap(err, "could not get observation")
 	}
 	annotations := mg.GetAnnotations()
-	if annotationUpdate, err := MoveTFStateValuesToAnnotation(stateValueMap, obs, annotations, n.config); err != nil {
+	if annotationUpdate, err := moveTFStateValuesToAnnotation(stateValueMap, obs, annotations, n.config); err != nil {
 		return managed.ExternalUpdate{}, errors.Wrap(err, "cannot move status values to annotation")
 	} else if annotationUpdate {
 		mg.SetAnnotations(annotations)
--- a/pkg/controller/external_tfpluginsdk.go
+++ b/pkg/controller/external_tfpluginsdk.go
@@ -132,7 +132,7 @@ func getExtendedParameters(ctx context.Context, tr resource.Terraformed, externa
 			return nil, errors.Wrap(err, "cannot get merged parameters")
 		}
 	} else {
-		params, err = MergeAnnotationFieldsWithSpec(tr, initParamsMerged, tr.GetAnnotations())
+		params, err = mergeAnnotationFieldsWithSpec(tr, initParamsMerged, tr.GetAnnotations())
 		if err != nil {
 			return nil, errors.Wrap(err, "cannot merge annotation fields")
 		}
@@ -269,7 +269,7 @@ func (c *TerraformPluginSDKConnector) Connect(ctx context.Context, mg xpresource
 		if err != nil {
 			return nil, errors.Wrap(err, "failed to get the observation")
 		}
-		if err := MergeAnnotationFieldsWithStatus(tfState, mg.GetAnnotations(), c.config); err != nil {
+		if err := mergeAnnotationFieldsWithStatus(tfState, mg.GetAnnotations(), c.config); err != nil {
 			return nil, errors.Wrapf(err, "failed to merge annotations on resource %q", client.ObjectKeyFromObject(mg))
 		}
 		tfState, err = c.config.ApplyTFConversions(tfState, config.ToTerraform)
@@ -582,7 +582,7 @@ func (n *terraformPluginSDKExternal) Observe(ctx context.Context, mg xpresource.
 			return managed.ExternalObservation{}, errors.Wrap(err, "could not get observation")
 		}
 		annotations := mg.GetAnnotations()
-		annotationUpdate, err := MoveTFStateValuesToAnnotation(stateValueMap, obs, annotations, n.config)
+		annotationUpdate, err := moveTFStateValuesToAnnotation(stateValueMap, obs, annotations, n.config)
 		if err != nil {
 			return managed.ExternalObservation{}, errors.Wrap(err, "cannot move status values to annotation")
 		}
@@ -701,7 +701,7 @@ func (n *terraformPluginSDKExternal) Create(ctx context.Context, mg xpresource.M
 		return managed.ExternalCreation{}, errors.Wrap(err, "could not get observation")
 	}
 	annotations := mg.GetAnnotations()
-	if annotationUpdate, err := MoveTFStateValuesToAnnotation(stateValueMap, obs, annotations, n.config); err != nil {
+	if annotationUpdate, err := moveTFStateValuesToAnnotation(stateValueMap, obs, annotations, n.config); err != nil {
 		return managed.ExternalCreation{}, errors.Wrap(err, "cannot move status values to annotation")
 	} else if annotationUpdate {
 		mg.SetAnnotations(annotations)
@@ -773,7 +773,7 @@ func (n *terraformPluginSDKExternal) Update(ctx context.Context, mg xpresource.M
 		return managed.ExternalUpdate{}, errors.Wrap(err, "could not get observation")
 	}
 	annotations := mg.GetAnnotations()
-	if annotationUpdate, err := MoveTFStateValuesToAnnotation(stateValueMap, obs, annotations, n.config); err != nil {
+	if annotationUpdate, err := moveTFStateValuesToAnnotation(stateValueMap, obs, annotations, n.config); err != nil {
 		return managed.ExternalUpdate{}, errors.Wrap(err, "cannot move status values to annotation")
 	} else if annotationUpdate {
 		mg.SetAnnotations(annotations)
--- a/pkg/controller/annotation_conversions.go
+++ b/pkg/controller/annotation_conversions.go
@@ -19,7 +19,7 @@ import (
 	"github.com/crossplane/upjet/v2/pkg/types/name"
 )
 
-// MergeAnnotationFieldsWithSpec merges field values stored in annotations back into
+// mergeAnnotationFieldsWithSpec merges field values stored in annotations back into
 // the spec.forProvider and spec.initProvider parameter maps. This function is critical
 // for handling API version compatibility when controllers run older API versions.
 //
@@ -49,7 +49,7 @@ import (
 //   - annotations: The resource's annotations map
 //
 // Returns the merged parameters map or an error if field operations fail.
-func MergeAnnotationFieldsWithSpec(tr resource.Terraformed, shouldMergeInitProvider bool, annotations map[string]string) (map[string]any, error) { //nolint:gocyclo
+func mergeAnnotationFieldsWithSpec(tr resource.Terraformed, shouldMergeInitProvider bool, annotations map[string]string) (map[string]any, error) { //nolint:gocyclo
 	parameters, err := tr.GetParameters()
 	if err != nil {
 		return nil, errors.Wrap(err, "cannot get parameters")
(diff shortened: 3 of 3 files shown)


pkg/config/resource.go
```diff
@@ -538,6 +538,31 @@ type Resource struct {
...
+	//
+	// Example:
+	//   TfStatusConversionPaths: []string{
+	//       "status.atProvider.newStatusField",
+	//       "status.atProvider.anotherNewField",
+	//   }
+	//
+	// When the field doesn't exist in the old API version's status.atProvider,
+	// it will be stored as annotation["internal-upjet/status.atProvider.newStatusField"].
+	TfStatusConversionPaths []string
```
erhancagirici, line 576: Just for clarification, these paths are CRD fieldpaths. Therefore, they actually describe a particular version's fieldset. This configuration is a resource-level config and does not distinguish versions.

pkg/config/conversion/conversions.go
```diff
@@ -335,3 +406,264 @@ func ExpandParameters(prefixes []string, excludePaths ...string) []string {
 func DefaultPathPrefixes() []string {
 	return []string{pathForProvider, pathInitProvider, pathAtProvider}
 }
+
+// generateAnnotationKey generates the annotation key from a field path.
+// The caller must ensure fieldPath produces a valid Kubernetes annotation key
+// (max 63 characters, matching regex: [a-z0-9]([-a-z0-9]*[a-z0-9])?).
+func generateAnnotationKey(fieldPath string) string {
```
erhancagirici: Is there a way out if the fieldpath is long or produce an invalid annotation name?

Also, just thinking out loud: 
Instead of having annotation per fieldpath, Maybe we can consider having a single annotation, that stores all of the fields to inject. A JSON-encoded string that is map of fieldpaths to field values. I think we'd be modifying this annotation atomically and with full intent all the time. Something like:
```yaml
metadata:
  annotations:
    internal.upjet.crossplane.io/new-fields: |
      {"spec.forProvider.foo":{"fizz":"fizzval", "buzz":"buzzval"}}"
```

pkg/config/conversion/conversions.go
```diff
@@ -335,3 +406,264 @@ func ExpandParameters(prefixes []string, excludePaths ...string) []string {
...
+	}
+
+	return result, nil
+}
+
+// NewFieldTypeConversion returns a new Conversion that handles type changes
+// of fields between API versions. It converts primitive types according to the
+// given conversion mode.
+// Type assumptions for upjet:
+// - Integers are represented as int64 (though JSON unmarshaling may produce float64)
```
erhancagirici, line 685: nit: might need to double check on the json library we use for this to get a consistent behavior.

pkg/controller/external_tfpluginfw.go
```diff
@@ -699,6 +708,14 @@ func (n *terraformPluginFrameworkExternalClient) Create(ctx context.Context, mg
 		return managed.ExternalCreation{}, errors.Errorf("could not set observation: %v", err)
 	}
 
+	obs, err := mg.(resource.Terraformed).GetObservation()
+	if err != nil {
+		return managed.ExternalCreation{}, errors.Wrap(err, "could not get observation")
+	}
+	if err := MoveTFStateValuesToAnnotation(stateValueMap, obs, mg.GetAnnotations(), n.config.TfStatusConversionPaths); err != nil {
```
erhancagirici: Here, we possibly update annotations, and it is on the runtime object only at this stage.. We need to persist this. Might need to double check if the managed reconciler persists this.

pkg/controller/external_tfpluginsdk.go
```diff
@@ -228,6 +231,140 @@ func (c *TerraformPluginSDKConnector) applyHCLParserToParam(sc *schema.Schema, p
...
+						if err := parametersPaved.SetValue(snakeKey, v); err != nil {
+							return errors.Wrapf(err, "cannot set value for %s", snakeKey)
+						}
+					} else {
+						return errors.Wrapf(err, "cannot get the current value for %s", snakeKey)
+					}
+				}
+			case strings.HasPrefix(t, "spec.initProvider."):
+				key := strings.TrimPrefix(t, "spec.initProvider.")
+				snakeKey := name.NewFromCamel(key).Snake
```
erhancagirici: loud thinking:
- check if there are exceptions for this
- how does it play out with fooRefs and fooSelectors?

pkg/controller/external_tfpluginsdk.go
```diff
@@ -228,6 +231,140 @@ func (c *TerraformPluginSDKConnector) applyHCLParserToParam(sc *schema.Schema, p
 	return param
 }
```
erhancagirici, line 239: During `Connect()`, we restore prior state from observation and populate the opTracker state cache.
A variant of this function is also needed for merging annotation-sourced values to reconstructed state,
both in `external_tfpluginfw.go` ([here](https://github.com/crossplane/upjet/blob/cf84e7188f601c45ebd9bf87b008c7f11238925e/pkg/controller/external_tfpluginfw.go#L192)) and `external_tfpluginsdk.go`([here](https://github.com/crossplane/upjet/blob/cf84e7188f601c45ebd9bf87b008c7f11238925e/pkg/controller/external_tfpluginsdk.go#L262)) at `Connect()`

pkg/controller/external_tfpluginsdk.go
```diff
@@ -228,6 +231,140 @@ func (c *TerraformPluginSDKConnector) applyHCLParserToParam(sc *schema.Schema, p
...
+// If annotation "internal-upjet/spec.forProvider.newField" contains a value,
+// and "new_field" doesn't exist in parameters, it will be set from the annotation.
+//
+// Parameters:
+//   - parameters: The spec.forProvider parameters as a map
+//   - initParameters: The spec.initProvider parameters as a map
+//   - annotations: The resource's annotations map
+//
+// Returns an error if field operations fail.
+func MergeAnnotationFieldsWithSpec(parameters, initParameters map[string]any, annotations map[string]string) error { //nolint:gocyclo
```
erhancagirici: AFAIU, when the controller reconcile version is the latest available version, i.e. at runtime, we use the most up-to-date Go Structs of the MR,these runtime injections should be no-op, or not called at all. Maybe we can build stronger guardrails around this and exit early?

pkg/controller/external_tfpluginfw.go
```diff
@@ -609,6 +609,15 @@ func (n *terraformPluginFrameworkExternalClient) Observe(ctx context.Context, mg
 		if err != nil {
 			return managed.ExternalObservation{}, errors.Errorf("could not set observation: %v", err)
 		}
+
+		obs, err := mg.(resource.Terraformed).GetObservation()
+		if err != nil {
+			return managed.ExternalObservation{}, errors.Wrap(err, "could not get observation")
+		}
+		if err := MoveTFStateValuesToAnnotation(stateValueMap, obs, mg.GetAnnotations(), n.config.TfStatusConversionPaths); err != nil {
```
erhancagirici: Here, we possibly update annotations, and it is on the runtime object only at this stage. We need to persist this.
This could possibly need us to determine if any work was done here, and adjust `specUpdateRequired` accordingly.
note that this will require `LateInitialize` to be enabled, otherwise we would probably need an extra mechanism at the managed reconciler, so that we ensure updating annotations here.

erhancagirici (commented):
@sergenyalcin thanks for handling this! It is very clear that this is something tricky to handle. I've left some comments to consider

erhancagirici:
Reflecting our offline conversation with @sergenyalcin, a general comment about the runtime aspect:
Thinking about the configurability of `ControllerReconcileVersion`. Having it always set to latest available CRD version almost always makes sense. When we do that, actually all the runtime-related conversion logic *needed to be introduced in this PR*, becomes no-op, simplifying the flow and one less knob.
Otherwise, as more MR versions are introduced, the runtime logic needs to adapt to different storage version-reconcile version combinations, increasing the cognitive load, especially for resource configurations that involve path specifications and chain other conversions etc.
Therefore, we can consider deprecating the `ControllerReconcileVersion` eventually, and always use the latest by default.

pkg/config/resource.go
sergenyalcin: You're absolutely right that these paths describe a particular version's fieldset. I've updated the documentation for `TfStatusConversionPaths` to clarify this.

The reason this is at the resource level is that it describes the **complete set** of fields that exist in the `Version` (the latest/target version) but might not exist in older controller versions. When a controller running an older API version reconciles a resource created with a newer version, these paths tell the controller which fields to preserve in annotations. This is symmetric to how API conversions work - the conversion logic uses the same paths to decide which fields need to be stored/restored.

pkg/config/conversion/conversions.go
sergenyalcin: Excellent suggestion! I've completely refactored the implementation to use your proposed consolidated annotation approach. Changes made:

1. **Removed per-field annotations**: Eliminated `generateAnnotationKey()` and individual annotations

2. **Implemented consolidated JSON annotation**: Now using a single annotation key:
   ```yaml
   internal.upjet.crossplane.io/field-conversions:
     {"spec.forProvider.foo": "value", "status.atProvider.bar": {...}}
   ```

3. **No annotation key length issues**: Since we use a fixed annotation key, there's no risk of exceeding Kubernetes annotation key limits (63 chars)

4. **Atomic updates**: The entire field map is read, modified, and written atomically

This approach is much cleaner and addresses your concern about invalid annotation names!

pkg/config/conversion/conversions.go
sergenyalcin: I've verified we're using `encoding/json` from the standard library consistently across all conversion functions:

- `pkg/config/conversion/conversions.go` - imports `"encoding/json"`
- `pkg/controller/annotation_conversions.go` - imports `encodingjson "encoding/json"` (aliased to avoid conflicts with other json packages)

The behavior is consistent:
- All numeric values are handled uniformly through `any` type
- JSON marshal/unmarshal preserves types correctly (int64, float64, etc.)
- We use the same marshaling approach in both directions (ToAnnotation and FromAnnotation)
- During conversion, we marshal then unmarshal to prevent double-encoding: `json.Marshal(value)` → `json.Unmarshal(bytes, &value)` → store in map

Additionally, the `areValuesEqual()` helper function (annotation_conversions.go:320-342) compares values by marshaling both to JSON and comparing bytes, ensuring consistent equality checks regardless of numeric type representations.

pkg/controller/external_tfpluginsdk.go
sergenyalcin: I've implemented this in both SDK and FW variants. Created a new utility function `MergeAnnotationFieldsWithStatus()` in `pkg/controller/annotation_conversions.go`

**Implementation in Connect() methods:**

In `external_tfpluginsdk.go` (~line 200):
```go
if err := MergeAnnotationFieldsWithStatus(stateValueMap, mg.GetAnnotations(), n.config); err != nil {
    return nil, errors.Wrap(err, "cannot merge annotation fields with status")
}
```

In `external_tfpluginfw.go` (~line 215):
```go
if err := MergeAnnotationFieldsWithStatus(stateValueMap, mg.GetAnnotations(), n.config); err != nil {
    return nil, errors.Wrap(err, "cannot merge annotation fields with status")
}

**What it does:**
1. Reads the consolidated annotation containing field conversions
2. Extracts fields with `status.atProvider.*` prefix from the JSON map
3. Merges those values into the `atProvider` map (converting camelCase to snake_case)
4. Only updates fields if they don't exist or have different values

This ensures that when we reconstruct Terraform state from CRD status, we include fields from annotations, giving Terraform the complete state picture and preventing unnecessary diffs for fields that exist in annotations but not in the current schema.

File: pkg/controller/external_tfpluginsdk.go (hunk 7 above)
Comment (sergenyalcin) [reply pkg/controller/external_tfpluginsdk.go]: I've implemented early exit guards in all three runtime conversion functions:

**In `MergeAnnotationFieldsWithSpec()`**
```go
// Early exit if no conversion needed
parameters, err := tr.GetParameters()
if err != nil {
    return nil, errors.Wrap(err, "cannot get parameters")
}
// ... (continues only if ControllerReconcileVersion != Version)
```

**In `MoveTFStateValuesToAnnotation()`** (annotation_conversions.go:182-244):
```go
//nolint:staticcheck // still handling deprecated field behavior
if c.ControllerReconcileVersion == c.Version {
    return false, nil
}
```

**In `MergeAnnotationFieldsWithStatus()`** (annotation_conversions.go:271-318):
```go
//nolint:staticcheck // still handling deprecated field behavior
if c.ControllerReconcileVersion == c.Version {
    return nil
}
```

When the controller reconcile version matches the latest version, all runtime conversion logic is bypassed immediately, avoiding any unnecessary processing. This provides strong guardrails ensuring conversion overhead only occurs when actually needed (i.e., when a controller is running an older API version but reconciling resources created with newer versions).

The early exits happen before any annotation parsing or field operations, making them true no-ops with minimal overhead.

pkg/controller/external_tfpluginfw.go
sergenyalcin: I handled this by adding a returned bool from `MoveTFStateValuesToAnnotation`. According to its value, I manipulated the value of the `specUpdateRequired`.

As you pointed out, this will just work for `LateInitialize` enabled configurations. However, as we discussed offline, since we will update the controller versions to the latest in the providers, I believe that the new runtime aspect functions will be no-ops for many cases. Additionally, we deprecate the `ControllerReconcileVersion` and, in the medium term, these runtime aspects will be removed.

pkg/pipeline/templates/terraformed.go.tmpl
```diff
@@ -6,6 +6,7 @@ package {{ .APIVersion }}
 
 import (
 	"dario.cat/mergo"
+	"github.com/crossplane/upjet/v2/pkg/controller"
```
erhancagirici: nit: (I am aware that goimports will handle this at generation)
just to double-check, was this intentionally added or a leftover?

sergenyalcin: Thanks for pointing this out. As you are aware, I initially implemented the annotation-spec merge process in the template file. Now I have moved it to the `getExtendedParameters` functions. So this is a leftover. Removed it.

PR #564, opened 2026-01-05
tfsdk external client: Flush reconstructed states for recalculation after Observation failures

erhancagirici:

Continuation of https://github.com/crossplane/upjet/pull/555 for TF SDKv2 external client. #555 was covering TF plugin framework external client.

Quoting the description of #555

Background:
In the initial reconciliation of the MRs just after a provider pod restart OR they get created in the k8s apiserver, the very first `Connect()` constructs a (partial) TF prior state using the current status values (`status.atProvider`), and saves it to the opTracker state cache. This reconstructed state contains enough information to identify the external resource, and facilitates the TF resource read operation in the subsequent `Observe()` call. 
- After the successful initial observation, this state is immediately updated with the latest full state data from the resource read (empty state for a non-existent external resource). Subsequent reconciles do not reconstruct the state anymore.
- If the resource read fails, an error is returned without touching the constructed prior state and the reconcile is requeued.

This causes an issue for the following scenario:
- An MR manifest is applied with some invalid parameter values
- in the initial `Connect()` a prior state is constructed with those values and saved to state cache.
- the initial observation fails and returns an error. The state cache is not modified.
- State cache has still the "invalid" value.
- MR manifest is updated with the correct value.
- In the next reconciliation, the state is not reconstructed, because a prior state already exists.
- The observation/resource read fails again because the invalid prior state.
- Resource is stuck, until a provider pod restart.

In this change:
- while recording the TF states into the cache, attach a meta-information regarding whether this state is a "reconstructed" state or a regular one (that comes from a TF resource CRUD operation).
- After observation failures, flush reconstructed states, so that they are rebuilt, with new configuration values if any.

Addresses https://github.com/crossplane-contrib/provider-upjet-azuread/issues/287
I have:

Tested manually,
thanks to @jonasz-lasut for the clear test steps and testing, see https://github.com/crossplane/upjet/pull/564#issuecomment-3710554788

I conducted the same test with slightly different order and it was successful too.
- install provider-azuread v1.8.0
- apply MRs, observe synced and ready ✅ 
- upgrade provider-azuread to modified version: v2.2.0+this fix consumed
- observe that `Principal` MR is failing now, Synced: false
- update the external name to the new format
- `Principal` MR recovers without a pod restart

### DIFF
--- a/pkg/controller/external_tfpluginsdk.go
+++ b/pkg/controller/external_tfpluginsdk.go
@@ -307,7 +307,7 @@ func (c *TerraformPluginSDKConnector) Connect(ctx context.Context, mg xpresource
 			}
 			s.Meta[schema.TimeoutKey] = timeouts
 		}
-		opTracker.SetTfState(s)
+		opTracker.SetReconstructedTfState(s)
 	}
 
 	return &terraformPluginSDKExternal{
@@ -496,6 +496,7 @@ func (n *terraformPluginSDKExternal) Observe(ctx context.Context, mg xpresource.
 	newState, diag := n.resourceSchema.RefreshWithoutUpgrade(ctx, n.opTracker.GetTfState(), n.ts.Meta)
 	metrics.ExternalAPITime.WithLabelValues("read").Observe(time.Since(start).Seconds())
 	if diag != nil && diag.HasError() {
+		n.opTracker.ResetReconstructedTfState()
 		return managed.ExternalObservation{}, errors.Errorf("failed to observe the resource: %v", diag)
 	}
 	diffState := n.opTracker.GetTfState()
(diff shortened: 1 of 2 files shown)


erhancagirici:
@jonasz-lasut thanks for the clear test steps and conducting the test 🎉 

I have conducted the same test with slightly different order (the external name correction applied last) and it was successful too. 
- install provider-azuread `v1.8.0`
- apply MRs, observe synced and ready ✅ 
- upgrade provider-azuread to version: `v2.2.0`
- observe that `Principal` MR is failing now, Synced: false
- upgrade provider-azuread to modified version: `v2.2.0 + custom upjet with this PR consumed`
- observe that `Principal` MR is still failing,  Synced: false
- update the external name to the new format as the very last step.
- Observe that `Principal` MR recovers without a pod restart ✅ 

This ordering was to ensure that the fix works also without a provider/pod restart (since updating external name then upgrading the provider implicitly involves a restart , then we cannot be sure whether the restart or the code was the actual fix :) )

sergenyalcin (approved):
Thanks @erhancagirici LGTM!

PR #567, opened 2026-01-08
Integrate schema change detection and add some helpers for registering the conversion functions automatically

sergenyalcin:
## Overview

This PR introduces an automated system for detecting and handling schema changes in CRD API versions across upjet-based providers. When Terraform provider schemas evolve (field additions, deletions, or type changes), CRD API versions must adapt, creating potential data loss during upgrades/downgrades. This system automatically generates and registers conversion functions to preserve data integrity across API version transitions.

### Problem Statement

When a Terraform provider introduces schema changes (e.g., `string` field becomes `number`, or a field is added/removed), upjet-based Crossplane providers face several challenges:

1. **Data Loss During Upgrades**: Fields added in v1beta2 are silently dropped when controllers run v1beta1
2. **Type Conversion Failures**: Direct field copying fails when types change between versions
3. **Manual Conversion Burden**: Developers must manually write conversion functions for every schema change
4. **Singleton List Conversions**: Existing singleton list conversion system breaks when fields have type changes

Previously, handling these scenarios required:
- Manually detecting schema changes by comparing CRD versions
- Writing custom conversion functions for each schema change
- Manually excluding type-changed fields from identity converters
- Testing all version transition paths

This manual process is error-prone, time-consuming, and doesn't scale across 100+ resources in a provider.

### Controller Version Architecture

**Important Context**: This system includes two types of conversion logic:

1. **CRD Conversion Webhook Logic (Permanent)**
   - Type converters (string↔number, string↔bool)
   - Field addition/deletion handlers via CRD conversion webhooks
   - Registered by `RegisterAutoConversions()`
   - Handles conversions between API versions (e.g., v1beta1 ↔ v1beta2)
   - **This is the core permanent architecture**

2. **Runtime Annotation Handling (Temporary - Being Deprecated)**
   - Handles version mismatches between controller reconcile version and API storage version
   - Functions: `mergeAnnotationFieldsWithSpec()`, `moveTFStateValuesToAnnotation()`, `mergeAnnotationFieldsWithStatus()`
   - Only needed when controller reconciles on an older API version than the CRD storage version
   - **With this PR, providers update all controller versions to match the latest API version**
   - `ControllerReconcileVersion` configuration is **deprecated** and will be removed in future upjet releases
   - Once removed, the runtime annotation handling will also be removed

**Migration Strategy**: As part of adopting this system, providers update controller reconcile versions to match the latest CRD API version (see provider-upjet-azuread example). This eliminates the need for runtime annotation handling going forward.

### High-Level Flow

```
┌─────────────────────────────────────────────────────────────────────┐
│ Build Time (Code Generation)                                        │
├─────────────────────────────────────────────────────────────────────┤
│                                                                     │
│  1. Generate Go APIs (upjet)                                        │
│         ↓                                                           │
│  2. Generate CRDs (controller-gen)                                  │
│         ↓                                                           │
│  3. Analyze CRD Schema Changes (schemadiff tool)                    │
│         ↓                                                           │
│  4. Generate crd-schema-changes.json                                │
│         ↓                                                           │
│  5. Embed JSON in provider binary                                   │
│                                                                     │
└─────────────────────────────────────────────────────────────────────┘
                              ↓
┌─────────────────────────────────────────────────────────────────────┐
│ Runtime (Provider Startup)                                          │
├─────────────────────────────────────────────────────────────────────────┐
│                                                                         │
│  1. pc.ConfigureResources()                                             │
│         ↓                                                               │
│  2. ExcludeTypeChangesFromIdentity(pc, crdSchemaChanges)                │
│         → Populates IdentityConversionExcludePaths                      │
│         → Prevents type-changed fields from failing identity conversion │
│         ↓                                                               │
│  3. Provider-specific transformations (e.g., bumpVersionsWithLists)     │
│         → Uses IdentityConversionExcludePaths                           │
│         ↓                                                               │
│  4. RegisterAutoConversions(pc, crdSchemaChanges)                       │
│         → Registers field addition/deletion handlers                    │
│         → Registers type conversion functions                           │
│         → Skipped during code generation (generationProvider flag)      │
│                                                                         │
└─────────────────────────────────────────────────────────────────────────┘
                              ↓
┌─────────────────────────────────────────────────────────────────────┐
│ Runtime (CRD Conversion Webhook) - PERMANENT                        │
├─────────────────────────────────────────────────────────────────────┤
│                                                                     │
│  User applies v1beta2 resource                                      │
│         ↓                                                           │
│  Webhook converts to v1beta1 (if necessary/requested)               │
│         ↓                                                           │
│  Registered conversions execute in order:                           │
│    1. Identity conversion (skips excluded paths)                    │
│    2. Type conversions (string↔number, string↔bool)                 │
│    3. Field preservation (store/restore via annotations)            │
│         ↓                                                           │
│  Controller reconciles on v1beta2                                   │
│                                                                     │
└─────────────────────────────────────────────────────────────────────┘
                              ↓
┌─────────────────────────────────────────────────────────────────────┐
│ Runtime (Version Mismatch Handling) - TEMPORARY/DEPRECATED          │
├─────────────────────────────────────────────────────────────────────┤
│                                                                     │
│  ONLY ACTIVE IF: ControllerReconcileVersion ≠ API Version           │
│  (This scenario is being phased out)                                │
│         ↓                                                           │
│  Terraform parameters merged with annotation values                 │
│         ↓                                                           │
│  After Terraform operation, status fields moved to annotations      │
│  if not present in older API version                                │
│                                                                     │
│  [ControllerReconcileVersion is deprecated, will be removed]        │
│                                                                     │
└─────────────────────────────────────────────────────────────────────┘
```

#### 1. **schemadiff Tool** (`cmd/schemadiff/main.go`)
- CLI tool that analyzes CRD files to detect schema changes between API versions
- Uses `upbound/uptest/pkg/crdschema` library (now official-providers-ci)
- Outputs JSON mapping `{group}/{kind}` → list of changes per version transition
- Runs during `go generate` after CRD generation

**Example Output:**
```json
{
  "ec2.aws.upbound.io/VPC": {
    "versions": {
      "v1beta2": {
        "oldVersion": "v1beta1",
        "newVersion": "v1beta2",
        "changes": [
          {
            "path": "spec.forProvider.maxSize",
            "changeType": "type_changed",
            "oldValue": "string",
            "newValue": "number"
          },
          {
            "path": "spec.forProvider.newField",
            "changeType": "field_added"
          }
        ]
      }
    }
  }
}
```

#### 2. **Auto-Registration Functions** (`pkg/config/conversion.go`)

**`ExcludeTypeChangesFromIdentity()`**
- Scans JSON for `type_changed` entries
- Adds affected field paths to `r.AutoConversionRegistrationOptions.IdentityConversionExcludePaths`
- **Why this matters**: Identity conversion copies fields as-is. Type-changed fields would fail (can't assign number to string). Must be excluded BEFORE singleton list conversion runs.
- **Timing**: Must run BEFORE `bumpVersionsWithEmbeddedLists()` (or similar version bumping)

**`RegisterAutoConversions()`**
- Scans JSON for all three change types and registers appropriate converters:
  - **Field additions**: Store/restore via `internal.upjet.crossplane.io/field-conversions` annotation
  - **Field deletions**: Store/restore via same annotation (symmetric with additions)
  - **Type changes**: Register bidirectional type converters (string↔int, string↔float, string↔bool)
- **Timing**: Must run AFTER version bumping (uses the finalized conversion list)
- **Skipped during generation**: Checks `generationProvider` flag (no need for conversions during code gen)

#### 3. **Conversion Implementations** (`pkg/config/conversion/conversions.go`)

**`newlyIntroducedFieldConverter`**
- Handles fields that exist in some versions but not others
- **ToAnnotation mode**: Copies field value from spec/status into annotation (when converting to older version)
- **FromAnnotation mode**: Restores field value from annotation to spec/status (when converting to newer version)

**`fieldTypeConverter`**
- Handles primitive type conversions:
  - `IntToString` / `StringToInt`: int64 ↔ string
  - `FloatToString` / `StringToFloat`: float64 ↔ string
  - `BoolToString` / `StringToBool`: bool ↔ "true"/"false"
- Uses Go's `strconv` package for conversion
- Bidirectional: registers both directions for each type change

#### 4. **Runtime Annotation Handling (Temporary - Being Deprecated)** (`pkg/controller/annotation_conversions.go`)

**Purpose**: These functions handle version mismatches between `ControllerReconcileVersion` and the API version. **With this PR, providers update controller versions to match API versions, making these functions inactive.**

**`mergeAnnotationFieldsWithSpec()`**
- **When active**: Controller reconcile version < API version (e.g., controller on v1beta1, CRD on v1beta2)
- Called during Create/Update operations (in `getExtendedParameters()`)
- Reads `internal.upjet.crossplane.io/field-conversions` annotation
- Merges `spec.forProvider.*` and `spec.initProvider.*` values back into parameter maps
- Ensures Terraform sees complete parameter set even if controller runs older API version
- **Checks `ControllerReconcileVersion == Version` and returns early if equal (most common case)**

**`moveTFStateValuesToAnnotation()`**
- **When active**: Controller reconcile version < API version
- Called after Terraform operations return state
- Compares TF state fields against CRD status.atProvider schema
- For fields in TF state but NOT in status (controller running old version), stores values in annotation
- Only updates annotation if values changed (prevents unnecessary writes)
- Requires manual configuration via `r.TfStatusConversionPaths` (only needed during version mismatch)
- **Checks `ControllerReconcileVersion == Version` and returns early if equal**

**`mergeAnnotationFieldsWithStatus()`**
- **When active**: Controller reconcile version < API version
- Called during Connect() to reconstruct full Terraform state
- Reads annotation and merges `status.atProvider.*` values back into status map
- Ensures Terraform operation tracker sees complete state
- **Checks `ControllerReconcileVersion == Version` and returns early if equal**

**Deprecation Timeline**:
- `ControllerReconcileVersion` field marked as deprecated
- Will be removed in future major upjet release
- Runtime annotation handling functions will be removed at the same time
- CRD conversion webhook logic (Component #3) remains permanent

---

## Design Decision: Why schemadiff Runs in Provider Generation Pipeline

Decision: The schemadiff tool is invoked in the provider's generate/generate.go file, not within upjet's code generation logic.

Architectural Constraint:

Upjet's architecture has a clear separation of concerns:

1. Upjet's responsibility: Generate Go API types (structs, interfaces, controller scaffolding)
2. Provider's responsibility: Run controller-gen to convert Go types into CRDs
3. schemadiff's requirement: Needs CRD YAML files to exist before it can analyze them

This creates a dependency chain that prevents integration into upjet:

Upjet generates Go types
         ↓
Provider runs controller-gen (generates CRDs from Go types)
         ↓
Provider runs schemadiff (analyzes generated CRDs)
         ↓
Provider embeds JSON (makes analysis available at runtime)

Why schemadiff can't run earlier:
- Upjet doesn't generate CRDs—it generates Go types
- CRD generation happens via controller-gen in the provider's pipeline
- schemadiff needs actual CRD YAML files (with OpenAPI schemas) to compare versions
- These files only exist after controller-gen runs

Trade-off accepted: Providers must add two lines to their generate/generate.go file (go:generate directive + embed statement). This manual step is necessary due to the architectural boundary between upjet and CRD generation tooling.

The current design respects existing architectural boundaries while providing maximum flexibility to providers.

---

### New Files

| File | Lines | Purpose |
|------|-------|---------|
| `cmd/schemadiff/main.go` | 128 | CLI tool for CRD schema analysis |
| `pkg/config/conversion.go` | 260 | Auto-registration orchestration |
| `pkg/controller/annotation_conversions.go` | 338 | Runtime annotation merge logic |
| `tests/conversion/integration_test.go` | 1071 | Integration test suite |
| `pkg/config/conversion_test.go` | 1252 | Unit test suite |

### Modified Files

| File | Changes | Purpose |
|------|---------|---------|
| `pkg/config/conversion/conversions.go` | +363 lines | Type/field conversion implementations |
| `pkg/controller/external_tfpluginsdk.go` | Modified | Integration with annotation merge |
| `pkg/controller/external_tfpluginfw.go` | Modified | Integration with annotation merge |
| `pkg/config/resource.go` | +116 lines | New config options |

### Configuration API Additions

**`Resource.AutoConversionRegistrationOptions`**
```go
type AutoConversionRegistrationOptions struct {
    // Skip automatic conversion registration entirely
    SkipAutoRegistration bool

    // Exclude specific paths from auto-registration
    AutoRegisterExcludePaths []string

    // Internal: populated by ExcludeTypeChangesFromIdentity
    IdentityConversionExcludePaths []string
}
```

**`Resource.TfStatusConversionPaths`**
```go
// List of status.atProvider.* paths that should be moved to/from annotations
// when controller runs older API version than CRD storage version
TfStatusConversionPaths []string
```

---

### 1. **Number→String Defaults to Integer Conversion**

**Decision**: When auto-detecting `number→string` type changes, default to int64 conversion.

**Rationale**:
- Cannot access old Terraform schema to determine if field was int or float
- Majority of number fields in Terraform are integers
- Providers have escape hatch: `AutoRegisterExcludePaths` for manual registration

**Limitation**: Float fields (e.g., `3.14159`) will be truncated to `"3"` if not manually handled.

**Mitigation**: Will be documented prominently in README and inline comments. Developers can exclude and manually register float conversions.

### 2. **Function Call Ordering is Manual**

**Decision**: Keep `ExcludeTypeChangesFromIdentity()` and `RegisterAutoConversions()` as separate functions.

**Required Order**:
```go
pc.ConfigureResources()
ExcludeTypeChangesFromIdentity(pc, crdSchemaChanges)  // Must be first
bumpVersionsWithEmbeddedLists(pc)                      // Provider-specific
RegisterAutoConversions(pc, crdSchemaChanges)          // Must be last
```

**Rationale**:
- `bumpVersionsWithEmbeddedLists()` is provider-specific logic (not in upjet)
- Must run between the two upjet functions
- Encapsulating would require upjet to know about provider version bumping strategies

**Limitation**: Order is not enforced by type system.

**Mitigation**:
- Comprehensive documentation in PR and README
- Code comments at call sites explaining ordering requirement
- Integration tests verify correct ordering

### 3. **TfStatusConversionPaths Requires Manual Configuration (Temporary Feature)**

**Decision**: Auto-detect and handle `spec.*` fields automatically, but require manual configuration for `status.atProvider.*` fields.

**Rationale**:
- Status field handling only applies when `ControllerReconcileVersion ≠ Version` (controller on older API version)
- **With this PR, all providers update controllers to latest API version, making this feature inactive**
- `ControllerReconcileVersion` is **deprecated** and will be removed in future upjet releases
- Status field conversions will be removed along with `ControllerReconcileVersion`
- The complexity of auto-detecting status field needs doesn't justify the temporary use case
- Manual configuration makes the temporary nature explicit

**Current Status**:
- Feature implemented but inactive in providers following the new pattern
- Existing for backwards compatibility during migration period
- No need to configure for new provider integrations

**Legacy Usage** (if needed during migration):
```go
r.TfStatusConversionPaths = []string{
    "status.atProvider.newField",
}
```

**Deprecation Path**:
1. Providers update controller versions to match API versions (this PR)
2. `ControllerReconcileVersion` field marked deprecated
3. Future major upjet release removes `ControllerReconcileVersion` and `TfStatusConversionPaths`
4. Runtime annotation handling code removed

---

### Step 1: Update Controller Versions to Match API Versions

**Suggested First Step**: Ensure all resources reconcile on their latest API version. This will decrease runtime complexity.

In your provider-specific version bumping logic (e.g., `bumpVersionsWithEmbeddedLists()`):

```go
func bumpVersionsWithEmbeddedLists(pc *ujconfig.Provider) {
    for name, r := range pc.Resources {
        if slices.Contains(strings.Split(strings.TrimSpace(oldSingletonListAPIs), "\n"), name) {
            r.Version = "v1beta2"
            r.SetCRDStorageVersion("v1beta1")

            // IMPORTANT: Update controller version to match API version
            r.ControllerReconcileVersion = r.Version  // ← Set to latest version

            // Configure conversions...
            excludePaths := append(r.CRDListConversionPaths(), r.AutoConversionRegistrationOptions.IdentityConversionExcludePaths...)
            r.Conversions = []conversion.Conversion{
                conversion.NewIdentityConversionExpandPaths(conversion.AllVersions, conversion.AllVersions, conversion.DefaultPathPrefixes(), excludePaths...),
                conversion.NewSingletonListConversion("v1beta1", "v1beta2", conversion.DefaultPathPrefixes(), r.CRDListConversionPaths(), conversion.ToEmbeddedObject),
                conversion.NewSingletonListConversion("v1beta2", "v1beta1", conversion.DefaultPathPrefixes(), r.CRDListConversionPaths(), conversion.ToSingletonList),
            }
        }
        // Add TerraformConversions for all resources (not just bumped ones)
        r.TerraformConversions = []ujconfig.TerraformConversion{
            ujconfig.NewTFSingletonConversion(),
        }
    }
}
```

**Why this matters**:
- Eliminates version mismatch between controller and API
- Disables runtime annotation handling (temporary deprecated feature)
- Ensures controllers see the latest field definitions
- See provider-upjet-azuread example for reference

### Step 2: Add go:generate Directive

In `generate/generate.go`, after CRD generation:

```go
//go:generate go run -tags generate sigs.k8s.io/controller-tools/cmd/controller-gen object:headerFile=../hack/boilerplate.go.txt paths=../apis/... crd:allowDangerousTypes=true,crdVersions=v1 output:artifacts:config=../package/crds

// Generate CRD schema diff JSON file
//go:generate go run github.com/crossplane/upjet/v2/cmd/schemadiff -i ../package/crds -o ../config/crd-schema-changes.json
```

### Step 3: Embed JSON in Provider Config

In `config/provider.go`:

```go
//go:embed crd-schema-changes.json
var crdSchemaChanges []byte
```

### Step 4: Integrate During Provider Construction

In `GetProvider()`:

```go
func GetProvider(ctx context.Context, generationProvider bool) (*ujconfig.Provider, error) {
    // ... existing provider setup

    pc.ConfigureResources()

    // Step 1: Exclude type-changed fields from identity conversion
    if err := ujconfig.ExcludeTypeChangesFromIdentity(pc, crdSchemaChanges); err != nil {
        return nil, errors.Wrap(err, "cannot exclude type changes from identity conversion")
    }

    // Step 2: Provider-specific version bumping (includes controller version updates from Step 1)
    bumpVersionsWithEmbeddedLists(pc)

    // Step 3: Register automatic conversions (skip during generation)
    if !generationProvider {
        if err := ujconfig.RegisterAutoConversions(pc, crdSchemaChanges); err != nil {
            return nil, errors.Wrap(err, "cannot register auto conversions")
        }
    }

    return pc, nil
}
```

### Step 5: (Optional) Exclude Specific Paths for Custom Conversions

Only needed for float fields or custom conversion logic:

```go
// In resource-specific config (e.g., config/ec2/config.go)
pc.Resources["aws_instance"].AutoConversionRegistrationOptions.AutoRegisterExcludePaths = []string{
    "spec.forProvider.customField",  // Will write custom conversion
    "spec.forProvider.floatField",   // Number→string but field is actually float
}
```

### ~~Step 6: (Legacy) Configure Status Field Preservation~~

**Not needed for new integrations.** This step only applies if maintaining backwards compatibility with `ControllerReconcileVersion ≠ Version`.

<details>
<summary>Click to expand legacy configuration (deprecated)</summary>

```go
// LEGACY: Only if controller runs older API version than CRD storage version
// With Step 1 completed, this is inactive and will be removed in future
pc.Resources["aws_vpc"].TfStatusConversionPaths = []string{
    "status.atProvider.newField",
}
```
</details>

---

### 1. Field Additions
**Scenario**: Field exists in v1beta2 but not in v1beta1

**Behavior**:
- v1beta2 → v1beta1: Store field value in annotation
- v1beta1 → v1beta2: Restore field value from annotation

**Annotation Key**: `internal.upjet.crossplane.io/field-conversions`

### 2. Field Deletions
**Scenario**: Field exists in v1beta1 but not in v1beta2

**Behavior**: Same as field additions (symmetric)

### 3. Type Changes: String ↔ Number
**Scenario**: Field changes from string to number (or vice versa)

**String → Number**:
- Consults Terraform schema to determine int64 vs float64
- Registers appropriate converter (StringToInt or StringToFloat)

**Number → String**:
- Defaults to int64 → string
- **Limitation**: If field was actually float, precision loss occurs
- **Solution**: Use `AutoRegisterExcludePaths` and register manual FloatToString converter

### 4. Type Changes: String ↔ Boolean
**Scenario**: Field changes from string to boolean (or vice versa)

**Behavior**:
- String → Boolean: Parses "true"/"false" using `strconv.ParseBool`
- Boolean → String: Formats to "true"/"false"

---

### Unit Tests (`pkg/config/conversion_test.go`)

**Test Coverage**:
- Valid JSON parsing with multiple resources and version transitions
- Invalid JSON handling (malformed, unknown change types, missing fields)
- Type change detection and exclusion path population
- All type conversion modes (int, float, bool)
- Edge cases (missing Terraform schema, empty paths, etc.)

**Example Test Cases**:
- `TestRegisterAutoConversions_ValidJSON_MultipleResources`
- `TestRegisterAutoConversions_TypeChange_MissingTerraformSchema`
- `TestExcludeTypeChangesFromIdentity_ValidJSON`
- `TestRegisterAutoConversions_InvalidJSON_Malformed`

### Integration Tests (`tests/conversion/integration_test.go`)

**Test Coverage**:
- End-to-end type conversions (string↔int, string↔float, string↔bool)
- Field addition/deletion across version transitions
- Annotation storage and retrieval
- Multiple version hops (v1beta1 → v1beta2 → v1beta1)
- Nested field conversions
- List and map field handling

**Test Resources**:
- Custom test CRD with configurable fields
- Mock Terraform resources
- Test JSON files in `tests/testdata/`

**Example Test Cases**:
- `TestConversionIntegration_StringToInt`
- `TestConversionIntegration_BoolToString`
- `TestConversionIntegration_FieldAddition`
- `TestConversionIntegration_MultipleChanges`

---

### 1. Float Precision in Number→String Conversions
**Impact**: Float fields converting to string will lose precision if not manually configured.

**Workaround**: Use `AutoRegisterExcludePaths` and register manual conversion:
```go
r.AutoConversionRegistrationOptions.AutoRegisterExcludePaths = []string{
    "spec.forProvider.floatField",
}
r.Conversions = append(r.Conversions,
    conversion.NewFieldTypeConversion("v1beta1", "v1beta2",
        "spec.forProvider.floatField", conversion.FloatToString))
```

### 2. Function Call Ordering
**Impact**: Incorrect order causes runtime conversion failures.

**Workaround**: Follow documented order strictly. Integration tests will catch violations.

### 3. Annotation Size Limits
**Impact**: Kubernetes annotations limited to 256KB total per resource.

**Unlikely Scenario**: Would require many large fields to be added/removed simultaneously.

**Monitoring**: System doesn't enforce limit currently (future enhancement).

### 4. Status Field Conversions Require Manual Config (Deprecated Feature)
**Impact**: Status fields not automatically preserved during version transitions when `ControllerReconcileVersion ≠ Version`.

**Current Status**:
- Feature is inactive in providers following the new pattern (controller version = API version)
- Only relevant for legacy configurations
- Will be removed when `ControllerReconcileVersion` is removed in future upjet release

**Legacy Workaround** (if needed): Configure `TfStatusConversionPaths` explicitly.

**Recommended**: Update controllers to latest API version (Integration Guide Step 1) to avoid this entirely.

---

Note: Depends on https://github.com/upbound/official-providers-ci/pull/245

I have:

Tested in the provider-upjet-azuread: https://github.com/sergenyalcin/provider-upjet-azuread/tree/conversion-functions-demo. This branch consumes these changes and creates a toy problem for `applications.Application` resource for testing. This contains field_addition and field_deletion scenario.


pkg/config/testdata/valid/type-change-bool-to-string.json.license
```diff
@@ -0,0 +1,3 @@
+SPDX-FileCopyrightText: 2026 The Crossplane Authors <https://crossplane.io>
+
+SPDX-License-Identifier: Apache-2.0
```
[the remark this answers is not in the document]
sergenyalcin: It is the year 2026.

cmd/schemadiff/main.go
```diff
@@ -0,0 +1,128 @@
...
+
+		// Skip CRDs with no changes (all versions are identical)
+		if changeReport.Empty() {
+			continue
+		}
+
+		// Add this CRD's change report to the output map
+		// Key format: "{group}/{kind}" matches what conversion.go expects
+		crdSpec := sd.GetCRD().Spec
+		jsonData[fmt.Sprintf("%s/%s", crdSpec.Group, crdSpec.Names.Kind)] = changeReport
```
erhancagirici, line 96: for each `changeReport` we might need some stable ordering for the list of changes. Otherwise, those might end up shuffled at each generate. That would cause unnecessary diff for the semantically equivalent JSON. For runtime, this is not an issue but might be annoying for checking out the generated file to git.
Maybe something like below that sorts changes based on path.
```suggestion
    for _, versionChange := range changeReport.Versions {
			slices.SortStableFunc(versionChange.Changes, func(a, b crdschema.SchemaChange) int {
				return strings.Compare(a.Path, b.Path)
			})
		}
		jsonData[fmt.Sprintf("%s/%s", crdSpec.Group, crdSpec.Names.Kind)] = changeReport
```

go.mod
```diff
@@ -27,8 +27,10 @@ require (
 	github.com/pkg/errors v0.9.1
 	github.com/prometheus/client_golang v1.22.0
 	github.com/spf13/afero v1.12.0
+	github.com/stretchr/testify v1.10.0
 	github.com/tmccombs/hcl2json v0.3.3
-	github.com/yuin/goldmark v1.4.13
+	github.com/upbound/uptest v0.12.1-0.20260115130402-161407324a6d
```
erhancagirici: let's bump this to `8965579a213ba2b249225fa311178ac66f107daf` to transitively consume kin-`openapi@v0.133.0`

pkg/config/testdata/invalid/malformed.json
```diff
@@ -0,0 +1,19 @@
...
+{
+  "test.example.io/TestResource": {
+    "versions": {
+      "v1beta1": {
+        "oldVersion": "v1alpha1",
+        "newVersion": "v1beta1",
+        "changes": [
+          {
+            "path": "spec.forProvider.field"
+            // Missing comma - malformed JSON
```
erhancagirici: nit: Standard JSON (RFC 8259) does not support comments

pkg/config/conversion.go
```diff
@@ -0,0 +1,249 @@
...
+					case crdschema.ChangeTypeFieldDeleted:
+						// Field was removed in the new version
+						// old→new: Store field value in annotation before removal (ToAnnotation)
+						// new→old: Restore field value from annotation (FromAnnotation)
+						r.Conversions = append(r.Conversions, conversion.NewNewlyIntroducedFieldConversion(cv.OldVersion, cv.NewVersion, change.Path, conversion.ToAnnotation),
+							conversion.NewNewlyIntroducedFieldConversion(cv.NewVersion, cv.OldVersion, change.Path, conversion.FromAnnotation))
+					case crdschema.ChangeTypeTypeChanged:
+						// Field type changed between versions (e.g., string → number)
+						switch {
+						case change.OldValue == "string" && change.NewValue == "number":
```
erhancagirici: we should handle schemadiff tool changes per https://github.com/crossplane/upjet/pull/567/changes#r2993610996 and use the updated struct to check `TypeChangeDetails`
See an example impl at: https://github.com/erhancagirici/upjet/commit/78be519b97283af64072aadfe3bdff23e9689e2a#diff-b4c4d599bae1cf517143971bf26d497c7b891901b62031f224d84ef6bd0510caR123

pkg/config/conversion.go
```diff
@@ -0,0 +1,249 @@
+// SPDX-FileCopyrightText: 2025 The Crossplane Authors <https://crossplane.io>
+//
+// SPDX-License-Identifier: Apache-2.0
+
+package config
```
erhancagirici, line 5: There are some issues not directly related to this PR, but affecting the general integration. Using here as the placeholder for commenting on them.

erhancagirici: in `pkg/config/conversion/conversions.go`, the `newlyIntroducedFieldConversion` is unable to handle paths with wildcards.

The `schemadiff` tool document generates paths with wildcards and the conversion is registered with those. But at runtime, when working with annotations we need to expand these paths first to obtain the data to be written to annotation.

See this example commit regarding handling this: https://github.com/sergenyalcin/upjet/commit/56dcdcb3bc6c75ae284c507c51550bb429dded09

cmd/schemadiff/main.go
```diff
@@ -0,0 +1,128 @@
...
+	// jsonData will hold all change reports, keyed by "{group}/{kind}"
+	// Example key: "ec2.aws.upbound.io/VPC"
+	jsonData := map[string]*crdschema.ChangeReport{}
+
+	// Process each CRD file
+	for _, cfp := range crdFilePaths {
+		// Create a self-diff analyzer for this CRD
+		// "Self-diff" means comparing different versions within the same CRD file
+		// (e.g., v1beta1 vs v1beta2 in the same CRD)
+		sd, err := crdschema.NewSelfDiff(cfp, crdschema.WithSelfDiffCommonOptions(opts))
```
erhancagirici, line 62: Caveat: today, SelfDiff ignores and does not report diffs for a singleton list conversion is involved in the path. This was intended as we were registering singleton list conversions manually

say in `v1beta1`, we have the following field under `spec.forProvider.foo` which is a list of objects. something like:
```yaml
foo:
- bar: "bar-val"
  baz: "baz-val"
```
in v1beta2, it has a singleton list -> embedded object conversion:
```yaml
foo:
  bar: "bar-val"
  baz: "baz-val"
```
say we are bumping the TF provider and a new field `fizz` is introduced in the `foo`'s element schema.
This field addition is not captured in the generated schema diff JSON documentation, so the changes cannot be auto-registered.
These changes are only visible if you do a revision diff for `v1beta2` between pre and post generated v1beta2 crd schemas.

pkg/config/conversion.go
```diff
@@ -0,0 +1,249 @@
...
+			for _, cv := range changeReport.Versions {
+				for _, change := range cv.Changes {
+					// Allow resources to opt-out of automatic exclusion for specific paths
+					if isExcludedPath(change.Path, r.AutoConversionRegistrationOptions.AutoRegisterExcludePaths) {
+						continue
+					}
+					// Only type changes need to be excluded from identity conversion
+					// Field additions/deletions are handled fine by identity conversion
+					if change.ChangeType == crdschema.ChangeTypeTypeChanged {
+						setIdentityConversionExcludePath(r, change.Path, m)
```
erhancagirici, line 187: we might potentially need a versioning here when we have >2 versions:
say there has been a type-change for field `foo` while transitioning from `v1` to `v2`. Here, we indeed exclude it from identity conversion. But, say for an unrelated field , we have the version `v3`. 

From `v2` to `v3`, the field `foo` did not change. In that case, it should be included in the identity conversion.

erhancagirici (commented):
@sergenyalcin thanks for handling this important feature. I've left some comments for us to consider

cmd/schemadiff/main.go
sergenyalcin: Done!

go.mod
sergenyalcin: Done!

pkg/config/testdata/invalid/malformed.json
sergenyalcin: Done!

pkg/config/conversion.go
sergenyalcin: Done!

pkg/config/conversion.go
sergenyalcin: Done!

cmd/schemadiff/main.go
sergenyalcin: Good catch. This is a real gap in the current design. Since `SelfDiff` intentionally ignores singleton-list paths (to avoid noise from the embedded object conversion itself), any field additions/removals nested inside those paths are invisible to the generated change report. As a result, `RegisterAutoConversions` will never register conversions for those fields.

The right fix is to complement the self-diff with a revision-based diff on the newer version (e.g., comparing the pre-bump and post-bump v1beta2 CRD schemas) that would surface changes within singleton-list converted paths. However, that requires the tool to have access to the previously generated CRDs at run time, which is a bigger change in how the tool is invoked.

For now, I'd like to clearly document this as a known limitation in the tool's output and guide, and track the revision-diff support as a follow-up. Does that sound reasonable to you, or would you prefer we address it in this PR?

pkg/config/conversion.go
sergenyalcin: You're right, the current implementation is missing for providers with more than two versions. `IdentityConversionExcludePaths` is a flat `[]string` applied globally to all identity conversions, so a type change from v1beta1 -> v1beta2 will exclude that field from the v1beta2→v1beta3 identity conversion as well, even if the field's type is stable there.

The proper fix is to make `IdentityConversionExcludePaths` version-pair-aware, e.g., change it to map[string []string keyed by "oldVersion/newVersion". `ExcludeTypeChangesFromIdentity` would then populate the correct key per version pair, and the identity converter would look up only the exclusions relevant to the source/target versions being converted.

However, we can consider it as a future improvement and an iterative progress point. As you know, we don't think this PR is a full implementation. Since this is a library and there are APIs that allow excluding fields for auto-registered conversion functions, we can consider this for the next iteration.

erhancagirici: Let's consider this for the next iteration 👍

cmd/schemadiff/main.go
erhancagirici: I think it is OK to follow-up 👍

erhancagirici (approved):
@sergenyalcin thanks for the tremendous effort here 🙌 LGTM with follow-up items on:
- versioned identity conversion exclusions
-  improved detection for schema changes on singleton-lists

PR #585, opened 2026-01-12
Add new configuration APIs/mechanisms for starting to introduce new MR API version policy

sergenyalcin:

This PR adds infrastructure for managing the lifecycle of Managed Resource API versions, in accordance with Kubernetes deprecation policies.

1. New configurations:
  - `ServedVersions`: Controls which API versions the API server exposes
  - `DeprecatedVersions`: Maps versions to deprecation metadata (warning messages, release info)
  - `GetServedVersions()`: Provides safe access with proper defaulting behavior
2. CRD generation enhancements:
  - Generates `+kubebuilder:deprecatedversion:warning` markers for deprecated versions
  - Generates `+kubebuilder:unservedversion` markers for removed versions
  - Adds deprecation notices to CRD descriptions
3. Post-generation hooks:
  - `NewVersionMarkerUpdateHook()`: Updates lifecycle markers in previous version files
  - `NewStorageVersionMarkerUpdateHook()`: Ensures only the configured storage version has the marker
  - Required because previous version files aren't regenerated during normal pipeline execution

I have:

Tested in the provider-upjet azuread and provider-upjet-gcp. Observed successful generations. Deprecation and serving version markers were observed. There are no unintended changes.

Please see the provider branches that consume this change:
- https://github.com/sergenyalcin/provider-upjet-azuread/tree/marker-update-demo
- https://github.com/sergenyalcin/official-provider-gcp/tree/marker-update-demo

## Summary by CodeRabbit

* **New Features**
  * Added API version deprecation support with customizable warnings and planned removal dates.
  * Introduced explicit control over served API versions with validation.
  * Enhanced CRD generation with automatic deprecation notices, warnings, and unserved version markers for clearer API lifecycle management.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>


pkg/pipeline/crd.go
```diff
@@ -124,3 +148,44 @@ func deleteOmittedFields(sch map[string]*schema.Schema, omittedFields []string)
 		}
 	}
 }
+
+// buildEnhancedDeprecationWarning creates an enhanced deprecation warning message
+// by appending release information to the base warning message if available.
+func buildEnhancedDeprecationWarning(deprecation config.VersionDeprecation) string {
+	warning := deprecation.Warning
```
ulucinar:
```suggestion
	warning := strings.TrimSpace(deprecation.Warning)
```

pkg/config/resource.go
```diff
@@ -435,6 +435,27 @@ type MergeStrategy struct {
...
+
+	// DeprecatedInRelease indicates which provider release deprecated this version.
+	// Format: "v1.2.3"
+	DeprecatedInRelease string
+
+	// RemovalPlannedRelease indicates the provider release in which this version
+	// is planned to be removed. According to the versioning policy, a version must
+	// remain deprecated for at least one release before removal.
+	// Format: "v1.3.0"
+	RemovalPlannedRelease string
```
ulucinar: Because we've a preposition in the name `DeprecatedInRelease`, for consistency:
```suggestion
	RemovalPlannedForRelease string
```

Or something like `DeprecationRelease` & `PlannedRemovalRelease` (or just `RemovalRelease` concisely). I think these forms are more suitable for a declarative API btw.

ulucinar: We are coupling the specific release versions with the resource configurations I believe for the first time, optionally. I think this should be fine because:
- It's optional. Provider authors can opt out if they don't want such coupling.
- We are planning to drive the deprecation policy enforcement as autonomously as possible, i.e., there will be automatic configurators which utilize the information from `VersionDeprecation` to drive the policy as we make new releases.
- If the provider code lives in multiple repositories (has forks), most probably they will be following the same versioning policy and have the same versions. But if there are forks, this may also not be true and if the provider author does not want to maintain this configuration (even via some sort of automatic configuration), she can still opt out using this optional configuration.

So, I think this is perfectly fine. But because we are introducing a coupling with the resource configurations and the release versions of the providers, I believe it's worth surfacing this.

We could hold this information outside of the resource configuration and specific for a provider's implementation. The current approach makes it universally available for all upjet-based providers rather than being part of a repo specific API versioning policy driver. So I believe it's worth the coupling.

pkg/config/resource.go
```diff
@@ -435,6 +435,27 @@ type MergeStrategy struct {
...
+type VersionDeprecation struct {
+	// Warning is the deprecation warning message shown to users when they
+	// interact with the deprecated API version. This message should explain
+	// which version to migrate to and any important migration considerations.
+	// Maps to the +kubebuilder:deprecatedversion:warning marker in generated CRDs.
+	Warning string
+
+	// DeprecatedInRelease indicates which provider release deprecated this version.
+	// Format: "v1.2.3"
+	DeprecatedInRelease string
```
ulucinar:
```suggestion
	DeprecationRelease string
```
Please see the comment below also.

pkg/config/resource.go
```diff
@@ -465,6 +486,48 @@ type Resource struct {
...
+	// Example after deprecation:
+	//   Version: "v1beta2"
+	//   PreviousVersions: []string{"v1beta1"}
+	//   ServedVersions: []string{"v1beta2", "v1beta1"}  // Still served but v1beta1 marked deprecated
+	//
+	// Example after removal:
+	//   Version: "v1beta2"
+	//   PreviousVersions: []string{}  // v1beta1 removed
+	//   ServedVersions: []string{"v1beta2"}  // Only v1beta2 served
+	ServedVersions []string
```
ulucinar: We must ensure that `ServedVersions` contains `Version`. It may not be a superset of the set `PreviousVersions`, i.e., there might be some previous versions that are no longer served. Another constraint could be if a version is not the `Version` or is not in `PreviousVersions`, then it cannot be in `ServedVersions`, i.e., a removed version should not exist in `ServedVersions`. We will also likely be serving some deprecated versions but not all of them.

We could enforce these rules with a validation, or maybe unexporting `ServedVersions` and implement the enforcement in accessors. That would also enforce the use of `{Set,Get}ServedVersions` at the API level.

pkg/config/resource.go
```diff
@@ -465,6 +486,48 @@ type Resource struct {
...
+
+	// DeprecatedVersions maps API versions to their deprecation configuration.
+	// The key is the version string (e.g., "v1beta1"), and the value contains
+	// deprecation metadata including warning messages and release information.
+	// Deprecated versions should still be listed in ServedVersions until they are
+	// completely removed from the API.
```
ulucinar: Is this comment correct? I think we can have some deprecated versions that are no longer served, right?

pkg/config/resource.go
```diff
@@ -812,6 +875,30 @@ func (r *Resource) SetCRDHubVersion(v string) {
...
+// GetServedVersions returns the list of API versions that should be served
+// by the API server. If ServedVersions is explicitly configured, it returns
+// that list. Otherwise, it defaults to serving all versions: the current
+// Version plus all PreviousVersions, maintaining backward compatibility.
+//
+// This method ensures that:
+// - If ServedVersions is explicitly set, it's used as-is
+// - If ServedVersions is empty/nil, all versions are served by default
+// - The returned slice is a new slice to prevent external modifications
+func (r *Resource) GetServedVersions() []string {
```
ulucinar, line 970: The default implementations looks good for backward compatibility so that if a provider author does not start using the new APIs, we assume all previously configured API versions via the existing `Version` and `PreviousVersions` APIs are considered as being served.

One case we need to be careful for backward compatibility is what would happen if a provider author had multiple versions but did not bother to set the `PreviousVersions`? Is this possible? How do we use `PreviousVersions` before these changes? Can you check how we previously make use of the `PreviousVersions` API and leave a comment here for documentation purposes?

If some scenario like this is actually possible, i.e., the `PreviousVersions` API was _optional_ even with multi-version CRDs, then we may also consider not using `ServedVersions` if it's not explicitly configured or making this visible in the release notes.

pkg/pipeline/crd.go
```diff
@@ -124,3 +148,44 @@ func deleteOmittedFields(sch map[string]*schema.Schema, omittedFields []string)
...
+	notice.WriteString(") has been deprecated")
+
+	if deprecation.DeprecatedInRelease != "" {
+		notice.WriteString(fmt.Sprintf(" in release %s", deprecation.DeprecatedInRelease))
+	}
+
+	if deprecation.RemovalPlannedRelease != "" {
+		notice.WriteString(fmt.Sprintf(" and is planned for removal in release %s.", deprecation.RemovalPlannedRelease))
+	} else {
+		notice.WriteString(".")
```
ulucinar: nit: If both deprecation and removal releases are empty strings, then the comment is not terminated with a ".".

pkg/pipeline/crd.go
```diff
@@ -83,20 +83,44 @@ func (cg *CRDGenerator) Generate(cfg *config.Resource) (string, error) {
...
+		"ForProviderType":    gen.ForProviderType.Obj().Name(),
+		"InitProviderType":   gen.InitProviderType.Obj().Name(),
+		"AtProviderType":     gen.AtProviderType.Obj().Name(),
+		"ValidationRules":    gen.ValidationRules,
+		"Path":               cfg.Path,
+		"Scope":              string(cg.Scope),
+	}
+
+	// Add deprecation information if this version is deprecated
+	if deprecation, isDeprecated := cfg.DeprecatedVersions[cfg.Version]; isDeprecated {
```
ulucinar: I think we don't expect the actively being generated API version to be deprecated, right? This code path still generates the types of only the current API version, right?

pkg/pipeline/crd.go
```diff
@@ -83,20 +83,44 @@ func (cg *CRDGenerator) Generate(cfg *config.Resource) (string, error) {
...
+		crdMap["DeprecationWarning"] = buildEnhancedDeprecationWarning(deprecation)
+		crdMap["DeprecationNotice"] = buildDeprecationNotice(cfg.Version, deprecation)
+	}
+
+	// Check if this version should be marked as unserved
+	// Only mark as unserved if ServedVersions is explicitly configured
+	// and the current version is NOT in that list
+	if len(cfg.ServedVersions) > 0 {
```
ulucinar: Misconfiguration like forgetting to put the current API version into the served versions would result in the current version not being served. Please also see the discussions above. I believe we had better prevent such invalid configuration cases.

pkg/pipeline/hooks.go
```diff
@@ -0,0 +1,677 @@
...
+func NewStorageVersionMarkerUpdateHook() PostGenerationHook {
+	return PostGenerationHookFn(func(runner *PipelineRunner, provider *config.Provider, resourcesGroups map[string]map[string]map[string]*config.Resource) error {
+		updater := &storageVersionMarkerUpdater{
+			runner: runner,
+		}
+		return updater.updateMarkers(resourcesGroups, provider)
+	})
+}
+
+type storageVersionMarkerUpdater struct {
```
ulucinar: Let's break the implementations of these two hooks into their own files.

pkg/pipeline/hooks.go
```diff
@@ -0,0 +1,677 @@
...
+func NewVersionMarkerUpdateHook() PostGenerationHook {
+	return PostGenerationHookFn(func(runner *PipelineRunner, provider *config.Provider, resourcesGroups map[string]map[string]map[string]*config.Resource) error {
+		updater := &markerUpdater{
+			runner: runner,
+		}
+		return updater.updateMarkers(resourcesGroups, provider)
+	})
+}
+
+type markerUpdater struct {
```
ulucinar: Let's break the implementations of these two hooks into their own files.

ulucinar: Because we already have a struct type declaration (`markerUpdater`), the type can directly implement the `PostGenerationHook` interface. Or, you can convert `updateMarkers` into a `PostGenerationHookFn` but I think having the type is better for breaking its implementation into multiple methods (we can make both work I think).

Please also consider doing this for `storageVersionMarkerUpdater`.

ulucinar: Would it be possible to use higher-level string functions (search/replace/regex, etc.) to simplify this processor instead of using `go/{ast,parser,token}` maybe with some assumptions on the shape of the generated Go files? Can you give this a try?

We can also consider this for `storageVersionMarkerUpdater`.

pkg/pipeline/run.go
```diff
@@ -64,6 +69,11 @@ func Run(pcCluster, pcNamespace *config.Provider, rootDir string) {
 		ModulePathControllers: filepath.Join(pcCluster.ModulePath, "internal", "controller", "cluster"),
 
 		Scope: tjtypes.CRDScopeCluster,
+		// Register built-in post-generation hooks
+		PostGenerationHooks: []PostGenerationHook{
+			NewStorageVersionMarkerUpdateHook(),
```
ulucinar: A high-level comment: (That we also discussed offline) We would like to break the current monolithic pipeline into smaller, modular steps and be able to selectively generate a single version of a single resource, for example. Once we achieve this, it will become much more feasible for us to generate the selected older API versions of selected resources when we need to do so, like there's an API change in the "preferred" version and we need to update the old versions.

This is currently not practical and we are relying on complex post-processors to work around this. And this also means we are building technical debt when we need updates to the older versions (or simply we rely on manual code updates on previously generated code).

I'm a little bit hesitant with the hooks API we are introducing here, although hooks can be utilized for other purposes than what we are doing here with them (updating the already generated versions). 

Furthermore, we also setup a higher level code generation pipeline using the `//go:generate` markers on Go source, like [here](https://github.com/crossplane-contrib/provider-upjet-aws/blob/main/generate/generate.go#L11). So, the hook API here is a lower-level alternative to that with some convenience of itself being able to programmatically access the upjet resource configuration API.

Because of these reasons, I feel like it's better for us at least not to export this API, even if you'd like to utilize the convenience of accessing the resource configuration (which the new API-related configuration becomes part of).

We could also choose to move the new configuration (ServedVersions, etc.) out of upjet's resource configuration so that the implementations of these two post processors can completely be moved outside of upjet's code generation pipeline, and possibly added to the higher-level `go:generate`-based pipeline above. 

But I still think the current resource configuration is better because:
- Resource configuration determines/affects the shapes of the CRDs generated. Other aspects already live in upjet resource configurations. Why would we then move served version or deprecated version configurations out of upjet's resource configuration?
- Once we achieve a modular upjet code generation pipeline, we will need the resource configuration when re-generating the old versions according to the new version. In my opinion, this is a much cleaner, easier to maintain approach, then trying to post-process the already generated (but no longer auto-generated) code.

Another approach could be the move the new configuration and the post processors both out of upjet's code generation pipeline. But I also think why should we not enjoy the convenience as long as long we don't make this part of the API contract, till we do the refactoring on the code generation pipelines?

TLDR; how about unexporting the hooks APIs? If some unrelated use case pops up (some other use case than modifying the auto-generated code), then we can reconsider exporting them. We already know the drawbacks of updating auto-generated code in post processors so maybe we would like to prioritize the pipeline refactoring over adding new features that will require us to update the older versions. So we _may_ have a change to block further technical debt. 

When we export this, with a small probability, other provider maintainers may start relying on these APIs to update the older API versions like we do here. And according to our discussions, I believe we've consensus on preventing this as we this as technical debt. So let's be proactive and block the way to technical debt. Maybe some other use case pops up and then we reconsider.

pkg/pipeline/crd.go
sergenyalcin: Done!

pkg/config/resource.go
sergenyalcin: Done!

pkg/config/resource.go
sergenyalcin: Done!

pkg/config/resource.go
sergenyalcin: ServedVersions unexported, added accessors and some validations.

pkg/config/resource.go
sergenyalcin: Thanks for catching this. Yes this comment isn't accurate. Changed it.

pkg/config/resource.go
sergenyalcin: About `PreviousVersions` usage, I added comments to a few places. Also, I will add some statements to the release notes.

In the current situation, the `PreviousVersions` API is optional and managed by providers. However, if the provider wants to work everything in place (I mean, without the new configuration), they must be set. Because in the current code generation pipeline, we rely on this field being filled.

So at the API level, we don't enforce this, even if the provider has multiversion CRDs. However, in the current code-generation pipeline, we assume this is properly configured.

pkg/pipeline/crd.go
sergenyalcin: Done!

pkg/pipeline/crd.go
sergenyalcin: One use case must be this for deprecating the last (generated) API version:

The underlying TF provider deprecates some resources in minor versions and removes them in the next major version.

Until today, because we didn't have the resource-level deprecation configuration, we couldn't deprecate them on our side and directly remove resources when we bump the providers in major versions.

As you pointed, in the context of this PR, I mean in the context of multiversion lifecyle of CRDs of resources, this is not expected. However, deprecating a generated version (the latest version) is a legitimate scenario for this type of situation. So I preferred to add this.

pkg/pipeline/crd.go
sergenyalcin: Added validation.

pkg/pipeline/hooks.go
sergenyalcin: Done!

pkg/pipeline/hooks.go
sergenyalcin: Done!

sergenyalcin: Done!

sergenyalcin: Now we don't use the `ast`, `parser`, etc. libraries. Because we don't change any Go code, even if these libraries are just used to ensure more robustness, as you pointed out, these usages are a bit complex.

Now I just did some string manipulations by using the `strings` library. So, the current one is also working on the same logic.

For removal, I used the `Replace` function directly. However, for additions, some processing is necessary and applied.

pkg/pipeline/run.go
sergenyalcin: As we talked about offline, I agree on the unexporting them. Unexported this hook mechanism.

pkg/config/resource.go
ulucinar: Release notes is a good place to discuss these. At a minimum, we had better talk about the new APIs and link to a public discussion of MR API version lifecycle, and what happens if the provider authors opt out of lifecycle management via the new APIs. It's also probably important to underlines these are relevant if their MRs have multiple versions (most providers do not care for breaking APIs and overwrite their existing API versions instead).

pkg/config/resource.go
```diff
@@ -812,6 +884,180 @@ func (r *Resource) SetCRDHubVersion(v string) {
...
+	hasCurrentVersion := false
+	for _, v := range versions {
+		if v == r.Version {
+			hasCurrentVersion = true
+			break
+		}
+	}
+	if !hasCurrentVersion {
+		return errors.Errorf("served versions must include the current version %q", r.Version)
+	}
```
ulucinar, line 925: Let's simplify by merging this loop with the one that checks for known versions below or we could also check simply for `!slices.Contains(versions, r.Version)`.

pkg/config/resource.go
ulucinar: > So at the API level, we don't enforce this, even if the provider has multiversion CRDs. However, in the current code-generation pipeline, we assume this is properly configured.

What I understand from this statement that in the current shape of the code generation pipeline (without the changes from this PR), if a provider has multiple versions for a CRD, the corresponding `PreviousVersions` must already be configured properly, otherwise the generation already produces erroneous code, right (as its assumptions are violated)?

pkg/pipeline/crd.go
ulucinar: So this is more a case for removal of the resource type all together, not a version being deprecated in favor of a newer (preferred) version. We would then deprecate all of its versions and the API migration framework could be used for those scenarios, assuming there's some other resource (or set of resources) to replace the being removed resource.

pkg/config/resource.go
```diff
@@ -812,6 +884,180 @@ func (r *Resource) SetCRDHubVersion(v string) {
...
+// provider maintainers. If you need to serve versions not listed in
+// PreviousVersions, you must first add them to PreviousVersions before
+// calling SetServedVersions().
+//
+// Example usage:
+//
+//	resource.Version = "v1beta2"
+//	resource.PreviousVersions = []string{"v1beta1"}
+//	err := resource.SetServedVersions([]string{"v1beta2", "v1beta1"})
+func (r *Resource) SetServedVersions(versions []string) error {
```
ulucinar, line 908: Let's add some tests for these methods in resource_test.go.

pkg/pipeline/hooks_deprecated_version.go
```diff
@@ -0,0 +1,349 @@
...
+	return &lifecycleMarkerUpdater{}
+}
+
+type lifecycleMarkerUpdater struct {
+	runner *PipelineRunner
+}
+
+// Run implements postGenerationHook for lifecycleMarkerUpdater.
+func (mu *lifecycleMarkerUpdater) Run(runner *PipelineRunner, provider *config.Provider, resourcesGroups map[string]map[string]map[string]*config.Resource) error { //nolint:gocyclo // easier to follow as a unit
+	mu.runner = runner
```
ulucinar, line 35: We had better do this during the construction (in the `new...` functions). We also had better remove the pipeline parameter itself from `postGenerationHook.Run`. We could also just receive the value of the `DirAPIs` instead of the whole pipeline object for the construction of these post hooks (for information hiding).

pkg/pipeline/hooks_deprecated_version.go
```diff
@@ -0,0 +1,349 @@
...
+}
+
+func (mu *lifecycleMarkerUpdater) updateVersionFile(group, version string, resource *config.Resource, isServed bool, isDeprecated bool, deprecation config.VersionDeprecation) error {
+	// Construct file path: <apiDir>/<shortGroup>/<version>/zz_<kind>_types.go
+	shortGroup := strings.ToLower(strings.Split(group, ".")[0])
+	fileName := fmt.Sprintf("zz_%s_types.go", strings.ToLower(resource.Kind))
+	filePath := filepath.Join(mu.runner.DirAPIs, shortGroup, version, fileName)
+
+	// Validate the file path is within the expected directory
+	if err := validateFilePath(filePath, mu.runner.DirAPIs); err != nil {
```
ulucinar, line 80: Why do we need a validation on file path here? We compute it directly from the resource configuration, i.e., pipeline's `DirAPIs`, resource's short group, version and kind.

pkg/pipeline/hooks_deprecated_version.go
ulucinar, line 1: I believe we could simplify all these by:
- Finding on the first occurrence of the `+kubebuilder:object:root=true` marker, then
  - Checking whether the `+kubebuilder:deprecatedversion:warning` marker exists or not
  - Checking whether the `+kubebuilder:storageversion` exists or not
  - Checking whether the `+kubebuilder:unservedversion` marker exists or not

until the first line that does not start with a comment, e.g., `^//.*` (another strong assumption).

The `+kubebuilder:object:root=true` can appear at two places in one of our generated files: Before the root object type for the MR, and before the corresponding list type. The MR type occurrence always comes before the list type's marker (strong assumption).

I think it's okay to make these strong assumptions for our generated files. It would also be good to add some comments telling that we prefer to add/remove these markers during the code generation itself once we support it (hence, the post hooks are a workaround).

PR #588, opened 2026-01-14
Add storage version migration support

sergenyalcin:

This PR adds storage version migration support to upjet, enabling automatic migration of CRD resources from old storage versions to new storage versions when CRD schemas are updated.

When a CRD's storage version changes (e.g., from `v1beta1` to `v1beta2`), existing resources stored in etcd remain in the old version until they are explicitly migrated. This can cause issues during upgrades, especially when introducing breaking changes.

The core logic was borrowed from https://github.com/crossplane/crossplane/blob/main/internal/initializer/crds_migrator.go.

The storage version migrator automates this migration process by:
1. Listing all resources of a CRD
2. Applying an empty patch to each resource (triggering the API server to convert and rewrite them in the new storage version)
3. Updating the CRD's `status.storedVersions` to reflect only the new storage version

### Core Implementation (`pkg/config/crds_migrator.go`)
- **`CRDMigrator`**: Handles storage version migration for a list of GroupVersionKinds
- **`Run()`**: Executes the migration process with pagination support (500 resources per batch)
- **`GetCRDNameFromGVK()`**: Resolves CRD names from GroupVersionKind using discovery client
- **`PrepareCRDMigrator()`**: Scans provider resources and creates a migrator for those with previous versions

### Provider Integration (`pkg/config/provider.go`)
- Added `StorageVersionMigrator` field to the `Provider` struct
- Added `WithStorageVersionMigrator()` provider option

## Example Usage

In provider configuration (`config/provider.go`):
```go
func GetProvider(ctx context.Context, sdkProvider *schema.Provider) (*ujconfig.Provider, error) {
    pc := ujconfig.NewProvider(...)

    // Configure resources
    pc.ConfigureResources()

    // Prepare storage version migrator for resources with previous versions
    ujconfig.PrepareCRDMigrator(pc)

    return pc, nil
}
```

In provider main (`cmd/provider/main.go`):
```go
// After setting up controllers, run storage version migration
discoveryClient, err := discovery.NewDiscoveryClientForConfig(cfg)
if err != nil {
	logr.Info("Failed to create discovery client, skipping the storage version migration", "err", err)
} else {
	// Create a non-cached client for the migration since the manager cache hasn't started yet
	directClient, err := client.New(cfg, client.Options{Scheme: mgr.GetScheme()})
	if err != nil {
		logr.Info("Failed to create direct client for storage version migration", "err", err)
	} else {
		if err = clusterProvider.StorageVersionMigrator.Run(ctx, logr, discoveryClient, directClient); err != nil {
			logr.Info("Failed to run storage version migrator", "err", err)
		}
	}
}
kingpin.FatalIfError(mgr.Start(ctrl.SetupSignalHandler()), "Cannot start controller manager")
```

I have:

Tested in this branch: https://github.com/sergenyalcin/provider-upjet-azuread/tree/sv-migrator. Validated the migration.


pkg/config/crds_migrator.go
```diff
@@ -0,0 +1,171 @@
...
+		}
+
+		// Find the current storage version (the version marked as storage in the spec)
+		var storageVersion string
+		for _, v := range crd.Spec.Versions {
+			if v.Storage {
+				storageVersion = v.Name
+				break
+			}
+		}
```
erhancagirici: very nit: probably impossible in practice, but just for the sake of being defensive, we can check that we actually found a storageVersion. something like
```suggestion
		}
		if storageVersion == "" {
		  errors.New("crd has no storage version")
		}
```

pkg/config/crd_migrator.go
```diff
@@ -0,0 +1,171 @@
...
+	return mapping.Resource.Resource + "." + mapping.Resource.Group, nil
+}
+
+// PrepareCRDsMigrator scans the provider's resources for any that have previous versions
+// and creates a CRDsMigrator to handle storage version migration for those resources.
+// It sets the StorageVersionMigrator field on the Provider with the configured migrator.
+func PrepareCRDsMigrator(pc *Provider) {
+	var gvkList []schema.GroupVersionKind
+	for _, r := range pc.Resources {
+		if len(r.PreviousVersions) != 0 {
```
erhancagirici, line 269: note: just for heads-up, not because there is an immediate issue. This is also not actually specific to the changes in this PR.

I checked the usage of `r.PreviousVersions` in general. We solely rely on the provider (therefore the developer) to correctly configure it. There are no validations/cross-check around this configuration (especially with `r.Version`). e.g:
- it is possible to forget configuring this: `r.Version = "v1beta2"` and `r.PreviousVersions = []` (this might be still a valid scenario though after v1beta1 is actually removed)
- no check for Version and PreviousVersions are disjoint sets. `r.Version = "v1beta2"` and `r.PreviousVersions = ["v1beta1", "v1beta2"]` possible
- no check that `PreviousVersions` are behind the current Version `r.Version = "v1beta2"` and `r.PreviousVersions = ["v1beta3", "v1beta4"]`

Just noting these since it is getting a new feature around it.

pkg/config/crds_migrator.go
```diff
@@ -0,0 +1,171 @@
...
+
+// PrepareCRDsMigrator scans the provider's resources for any that have previous versions
+// and creates a CRDsMigrator to handle storage version migration for those resources.
+// It sets the StorageVersionMigrator field on the Provider with the configured migrator.
+func PrepareCRDsMigrator(pc *Provider) {
+	var gvkList []schema.GroupVersionKind
+	for _, r := range pc.Resources {
+		if len(r.PreviousVersions) != 0 {
+			gvkList = append(gvkList, schema.GroupVersionKind{
+				Group:   r.ShortGroup + "." + pc.RootGroup,
```
erhancagirici: nit: just for being defensive
```suggestion
				Group:   strings.ToLower(r.ShortGroup + "." + pc.RootGroup),
```

pkg/config/crds_migrator.go
```diff
@@ -0,0 +1,171 @@
...
+
+// GetCRDNameFromGVK returns the CRD name (e.g., "resources.group.example.com") for a given GroupVersionKind
+// by using the discovery client to find the REST mapping.
+func GetCRDNameFromGVK(discoveryClient discovery.DiscoveryInterface, gvk schema.GroupVersionKind) (string, error) {
+	groupResources, err := restmapper.GetAPIGroupResources(discoveryClient)
+	if err != nil {
+		return "", err
+	}
+
+	mapper := restmapper.NewDiscoveryRESTMapper(groupResources)
```
erhancagirici: IIRC, these operations are quite heavy on API server discovering all APIs, and this function is called per GVK.

I think calling this portion outside the loop once would suffice for constructing a mapper first.
Then this function can have the `mapper` as an parameter and avoids re-discovery per gvk

pkg/config/crds_migrator.go
```diff
@@ -0,0 +1,171 @@
...
+			if continueToken == "" {
+				break
+			}
+		}
+
+		// Update CRD status to reflect that only the new storage version is stored
+		origCrd := crd.DeepCopy()
+
+		crd.Status.StoredVersions = []string{storageVersion}
+		if err := kube.Status().Patch(ctx, &crd, client.MergeFrom(origCrd)); err != nil {
```
erhancagirici: unfortunately, provider might need an extra RBAC here for patching CRDs, which is not ideal.

sergenyalcin: Yes, that's right. I manually provided these roles to the provider ClusterRole and tested. Therefore, if we don't take this step, I believe it will be updated at some point. However, we can't validate the migration.

From an RBAC manager's perspective, granting this permission to all providers may not be a sensible approach. This is because many providers are unwilling to perform this operation. An option is to proceed without validating this part, i.e., patching the CRDs. This sounds better than granting this right to all providers.

pkg/config/crd_migrator.go
sergenyalcin: Yes, I also considered this, and I believe that having a complete version list, etc., as a new feature, may make sense. And this must be stored automatically without manual configuration.

pkg/config/crds_migrator.go
sergenyalcin: Done!

pkg/config/crds_migrator.go
sergenyalcin: Done!

pkg/config/crds_migrator.go
sergenyalcin: Done!

sergenyalcin:
After the latest changes, the implementation consists of two main components:

## 1. Library: `CRDMigrator` (`pkg/config/crds_migrator.go`)

**Automatic Migration During Provider Startup**:
- The provider automatically migrates custom resources (CRs) to the new storage version during startup
- Resources are patched with an empty patch to trigger conversion by the API server
- Includes configurable retry logic with exponential backoff for resilience

**Default Behavior & RBAC Limitation**:
- ✅ **CR Migration**: Successfully patches all CRs to trigger storage version conversion
- ❌ **CRD Status Update**: By default, **providers lack permission** to update CRD status in the Crossplane ecosystem

The CRD status update is the final step that marks the old storage version as removed. Without this, while all resources are successfully migrated to the new storage version, the CRD status will still list old versions in `status.storedVersions`.

## 2. CLI Tool: `crdmigrator` (`cmd/crdmigrator/`)

A standalone tool for cluster administrators to complete the CRD status update when providers don't have sufficient permissions.

```bash
# Using comma-separated list
crds-migrator update \
  --crd-names "buckets.s3.aws.upbound.io:v1beta2,users.iam.aws.upbound.io:v1beta1"

# Using YAML file
crds-migrator update --crd-file crds-mapping.yaml
```

### YAML format:
```yaml
buckets.s3.aws.upbound.io: v1beta2
users.iam.aws.upbound.io: v1beta1
```

## **Provider Release Integration**

### For Provider Maintainers:

When releasing a provider version that includes storage version changes, include this information in release notes:

1. **List of CRDs requiring migration** with their target storage versions
2. **Two migration options for users:**

#### Option 1: Grant CRD Status Permissions to Provider (Recommended for automated environments)

- Add this ClusterRole to your provider deployment
- Then bind it to your provider's ServiceAccount. The provider will automatically complete the full migration on startup.

#### Option 2: Manual Migration Using CLI Tool (Recommended for controlled environments)

Download the `crdmigrator` tool and run it with cluster admin credentials:

```bash
# Example for provider-aws v2.5.0
crds-migrator update --crd-file migration-v2.5.0.yaml
```

Provide the `migration-v2.5.0.yaml` file in the release assets with the CRD mappings.

cmd/crdsmigrator/main.go
```diff
@@ -0,0 +1,209 @@
...
+		return errors.Wrap(err, "failed to create kube client")
+	}
+
+	// Configure retry backoff
+	retryBackoff := wait.Backoff{
+		Duration: time.Duration(*updateDuration) * time.Second,
+		Factor:   *updateFactor,
+		Jitter:   *updateJitter,
+		Steps:    *updateRetries,
+	}
```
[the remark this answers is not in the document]
sergenyalcin: Added validation and also default for `Steps` field.

pkg/config/crds_migrator.go
```diff
@@ -0,0 +1,267 @@
...
+
+			for i := range resources.Items {
+				// apply empty patch for storage version upgrade with retry
+				res := resources.Items[i]
+				patchErr := retry.OnError(c.retryBackoff, func(err error) bool { return true }, func() error {
+					return kube.Patch(ctx, &res, client.RawPatch(types.MergePatchType, []byte(`{}`)))
+				})
+				if patchErr != nil {
+					return errors.Wrapf(patchErr, "cannot patch %s %q", crd.Spec.Names.Kind, res.GetName())
+				}
```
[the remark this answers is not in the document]
sergenyalcin: Done!

pkg/config/crd_migrator.go
```diff
@@ -0,0 +1,267 @@
...
+				Subresource: "status",
+				Name:        crdName,
+				Verb:        "patch",
+			},
+		},
+	}
+
+	if err := kube.Create(ctx, ssar); err != nil {
+		return false, errors.Wrap(err, "failed to create SelfSubjectAccessReview for verb patch")
+	}
```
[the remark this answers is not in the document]
sergenyalcin: Done!

pkg/config/crds_migrator.go
```diff
@@ -0,0 +1,275 @@
...
+		for _, storedVersion := range storedVersions {
+			if storedVersion != storageVersion {
+				needMigration = true
+				break
+			}
+		}
+
+		if !needMigration {
+			logr.Debug("Skipping CRD migration for CRD because it has already been migrated", crdName)
+			continue
```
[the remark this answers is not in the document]
sergenyalcin: Done!

pkg/config/crds_migrator.go
```diff
@@ -0,0 +1,275 @@
...
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/client-go/discovery"
+	"k8s.io/client-go/restmapper"
+	"k8s.io/client-go/util/retry"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// The all logic was borrowed from crossplane-runtime. Because the relevant
```
ulucinar: nit: Can you mention the specific package we are referring to here? Also, it would be good to mention & perma-link in the PR description.

pkg/config/crds_migrator.go
```diff
@@ -0,0 +1,275 @@
...
+		var needMigration bool
+		for _, storedVersion := range storedVersions {
+			if storedVersion != storageVersion {
+				needMigration = true
+				break
+			}
+		}
+
+		if !needMigration {
+			logr.Debug("Skipping CRD migration for CRD because it has already been migrated", "crd-name", crdName)
```
ulucinar: nit for consistency:

```suggestion
			logr.Debug("Skipping CRD migration for CRD because it has already been migrated", "crd", crdName)
```
We may also use the key `CRD` (all capital).

pkg/config/crds_migrator.go
```diff
@@ -0,0 +1,275 @@
...
+			Kind:    crd.Spec.Names.ListKind,
+		})
+
+		// List all resources in batches and patch each one to trigger storage version migration.
+		// The empty patch causes the API server to read the resource in its stored version
+		// and write it back in the current storage version.
+		var continueToken string
+		for {
+			// Retry resource listing with exponential backoff
+			listErr := retry.OnError(c.retryBackoff, func(err error) bool { return true }, func() error {
```
ulucinar: We may consider checking whether an error is retryable. With the existing implementation, we will retry even on exceeded context deadlines. So, something like:

```suggestion
			listErr := retry.OnError(c.retryBackoff,
			  func(err error) bool {
                 kerrors.IsInternalError(err) ||
                 kerrors.IsServerTimeout(err) ||
                 kerrors.IsTimeout(err) ||
                 kerrors.IsTooManyRequests(err) ||
                 kerrors.IsServiceUnavailable(err) || ...
			  }, func() error {
```

I don't know if there's an existing predicate that would select retryable or non-retryable errors so that we can use it. How does the upstream (crossplane-runtime) implementation handle this?

pkg/config/crds_migrator.go
```diff
@@ -0,0 +1,275 @@
...
+			})
+			if listErr != nil {
+				return errors.Wrapf(listErr, "cannot list %s", resources.GroupVersionKind().String())
+			}
+
+			for i := range resources.Items {
+				// apply empty patch for storage version upgrade with retry
+				res := resources.Items[i]
+				patchErr := retry.OnError(c.retryBackoff, func(err error) bool {
+					return !kerrors.IsNotFound(err)
```
ulucinar: Please see the comment above on the retriable function.

pkg/config/crds_migrator.go
```diff
@@ -0,0 +1,275 @@
...
+				}, func() error {
+					return kube.Patch(ctx, &res, client.RawPatch(types.MergePatchType, []byte(`{}`)))
+				})
+				if patchErr != nil {
+					if kerrors.IsNotFound(patchErr) {
+						continue
+					}
+					return errors.Wrapf(patchErr, "cannot patch %s %q", crd.Spec.Names.Kind, res.GetName())
```
ulucinar: We had better qualify at least using the group name, or the error message could be ambiguous.

pkg/config/crds_migrator.go
```diff
@@ -0,0 +1,275 @@
...
+		}
+
+		// Check if the client has permission to update/patch CRD status before attempting the update
+		hasPermission, err := CheckCRDStatusUpdatePermission(ctx, kube, crdName)
+		if err != nil {
+			return errors.Wrapf(err, "permission check failed for CRD %s", crdName)
+		}
+
+		if !hasPermission {
+			logr.Info(fmt.Sprintf("This client does not have permission to execute %s operation for patch", crdName))
```
ulucinar: nit:

```suggestion
			logr.Info(fmt.Sprintf("This client does not have permission to patch the status of the CRD %s", crdName))
```

pkg/config/crds_migrator.go
```diff
@@ -0,0 +1,275 @@
...
+			continue
+		}
+
+		// Update CRD status to reflect that only the new storage version is stored
+		if err := UpdateCRDStorageVersion(ctx, kube, c.retryBackoff, crdName, storageVersion); err != nil {
+			return err
```
ulucinar: We do not wrap each error in `UpdateCRDStorageVersion`.

pkg/config/crds_migrator.go
```diff
@@ -0,0 +1,275 @@
...
+	return mapping.Resource.Resource + "." + mapping.Resource.Group, nil
+}
+
+// UpdateCRDStorageVersion updates the CRD status to reflect only the specified storage version.
+// It retries the update with exponential backoff and verifies the update was successful.
+func UpdateCRDStorageVersion(ctx context.Context, kube client.Client, retryBackoff wait.Backoff, crdName, storageVersion string) error {
+	var crd extv1.CustomResourceDefinition
+	// Update CRD status to reflect that only the new storage version is stored
+	// Use retry for status updates as they can fail due to conflicts
+	statusUpdateErr := retry.OnError(retryBackoff, func(err error) bool { return true }, func() error {
```
ulucinar: Please see the above comment on the retriable function.

pkg/config/crds_migrator.go
```diff
@@ -0,0 +1,275 @@
...
+		origCrd := crd.DeepCopy()
+		crd.Status.StoredVersions = []string{storageVersion}
+		return kube.Status().Patch(ctx, &crd, client.MergeFrom(origCrd))
+	})
+	if statusUpdateErr != nil {
+		return errors.Wrapf(statusUpdateErr, "couldn't update %s crd", crd.Name)
+	}
+
+	// One more check just to be sure we actually updated the crd
+	if err := kube.Get(ctx, client.ObjectKey{Name: crd.Name}, &crd); err != nil {
```
ulucinar: I think this may potentially result in buggy behavior if we are using a cached client here. Why do we need to assert this? If the above JSON merge patch has succeeded, then we should be fine (assuming no other client modifies the `.status.storedVersions` array, but even in that case, this is a one-shot migrator).

pkg/config/crds_migrator.go
```diff
@@ -0,0 +1,275 @@
...
+			return false, nil
+		}
+		return false, errors.Wrap(err, "failed to create SelfSubjectAccessReview for verb patch")
+	}
+
+	if !ssar.Status.Allowed {
+		return false, nil
+	}
+
+	return true, nil
```
ulucinar: nit:
```suggestion
	return ssar.Status.Allowed, nil
```

pkg/config/provider.go
```diff
@@ -370,6 +374,14 @@ func WithExampleManifestConfiguration(emc ExampleManifestConfiguration) Provider
 	}
 }
 
+// WithStorageVersionMigrator configures a CRDsMigrator for handling storage version
+// migrations of CRD resources when their schemas are updated.
+func WithStorageVersionMigrator(migrator *CRDsMigrator) ProviderOption {
```
ulucinar: So the provider may use this option, or if it maintains previous versions slices properly, the `config.PrepareCRDsMigrator`, right?

pkg/config/crds_migrator.go
```diff
@@ -0,0 +1,275 @@
...
+	"k8s.io/client-go/util/retry"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// The all logic was borrowed from crossplane-runtime. Because the relevant
+// files are in the internal package in crossplane-runtime. When we move them
+// outside the internal package, we can remove the duplication here.
+
+// CRDsMigrator makes sure the CRDs are using the latest storage version.
+type CRDsMigrator struct {
```
ulucinar: nit:

```suggestion
type CRDMigrator struct {
```
This name is I think more idiomatic. What's the upstream (crossplane-runtime) name for this?

cmd/crdsmigrator/main.go
```diff
@@ -0,0 +1,229 @@
...
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/clientcmd"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/yaml"
+
+	"github.com/crossplane/upjet/v2/pkg/config"
+)
+
+var (
+	app = kingpin.New("crds-migrator", "A CLI tool to manually update CRD storage versions for storage version migration")
```
ulucinar: nit:

```suggestion
	app = kingpin.New("crd-migrator", "A CLI tool to manually update CRD storage versions for storage version migration")
```
I think this name is more idiomatic for our context. Please see a comment below for a similar type name change...

cmd/crdsmigrator/main.go
```diff
@@ -0,0 +1,229 @@
...
+
+	// Parse from comma-separated flag (format: "crd1:version1,crd2:version2")
+	if crdNamesFlag != "" {
+		for _, pair := range strings.Split(crdNamesFlag, ",") {
+			trimmed := strings.TrimSpace(pair)
+			if trimmed == "" {
+				continue
+			}
+
+			parts := strings.SplitN(trimmed, ":", 2)
```
ulucinar: I think it's better to use:

```suggestion
			parts := strings.Split(trimmed, ":")
```
so that we don't accept a version like `v1beta1:v1beta2` with the second token.

cmd/crdsmigrator/main.go
```diff
@@ -0,0 +1,229 @@
...
+
+var (
+	app = kingpin.New("crds-migrator", "A CLI tool to manually update CRD storage versions for storage version migration")
+
+	// Global flags
+	kubeconfig = app.Flag("kubeconfig", "Path to kubeconfig file. If not specified, uses in-cluster config or default kubeconfig location").String()
+
+	// Update command
+	updateCmd      = app.Command("update", "Update CRD status to reflect the current storage version")
+	updateCRDNames = updateCmd.Flag("crd-names", "Comma-separated list of CRD:version pairs (e.g., 'buckets.s3.aws.upbound.io:v1beta2,users.iam.aws.upbound.io:v1beta1')").String()
```
ulucinar:
```suggestion
	updateCRDNames = updateCmd.Flag("crd-names", "Comma-separated list of <CRD>:<storage version> pairs (e.g., 'buckets.s3.aws.upbound.io:v1beta2,users.iam.aws.upbound.io:v1beta1')").String()
```

cmd/crdsmigrator/main.go
```diff
@@ -0,0 +1,229 @@
...
+	}
+
+	retryBackoff := wait.Backoff{
+		Duration: time.Duration(*updateDuration) * time.Second,
+		Factor:   *updateFactor,
+		Jitter:   *updateJitter,
+		Steps:    *updateRetries,
+	}
+
+	log.Println("Starting CRD storage version migration", "crd-count", len(crdVersionMap))
```
ulucinar: We may consider using the logging setup that we also use in provider main modules based on `github.com/crossplane/crossplane-runtime/v2/pkg/logging` for a familiar UX.

pkg/config/crds_migrator.go
ulucinar: Should we continue processing the CRDs instead of failing early here?

ulucinar (commented):
Thank you @sergenyalcin for implementing the CRD storage version migrator to be integrated to the Crossplane providers (possibly including non-upjet-based ones). Looks good overall, I think the most important comment above is the one related to the validation on a CRD's `status.storedVersions` after a JSON merge patch, as it may cause a bug.

pkg/config/crds_migrator.go
sergenyalcin: Added as code comment and to PR description. As reference: https://github.com/crossplane/crossplane/blob/main/internal/initializer/crds_migrator.go

pkg/config/crds_migrator.go
sergenyalcin: Done!

pkg/config/crds_migrator.go
sergenyalcin: I've added your suggestion. As far as I see in the crossplane side, there is no retry logic.

pkg/config/crds_migrator.go
sergenyalcin: Done!

pkg/config/crds_migrator.go
sergenyalcin: Done!

pkg/config/crds_migrator.go
sergenyalcin: Done!

pkg/config/crds_migrator.go
sergenyalcin: Done!

pkg/config/crds_migrator.go
sergenyalcin: Done!

pkg/config/crds_migrator.go
sergenyalcin: I've borrowed all this block from crossplane side: https://github.com/crossplane/crossplane/blob/e9d0c23f3fdaa0e526aaf4edf335f24cf6a70ab9/internal/initializer/crds_migrator.go#L111

But, your point makes sense to me so I removed this part.

pkg/config/crds_migrator.go
sergenyalcin: Done!

pkg/config/provider.go
sergenyalcin: Yes, that's right!

pkg/config/crds_migrator.go
sergenyalcin: Done! I've borrowed the name from crossplane but I believe this naming is better.

cmd/crdsmigrator/main.go
sergenyalcin: Done!

cmd/crdsmigrator/main.go
sergenyalcin: Done!

cmd/crdsmigrator/main.go
sergenyalcin: Done!

cmd/crdsmigrator/main.go
sergenyalcin: Done!

pkg/config/crds_migrator.go
sergenyalcin: I implemented this. Also CLI behavior is not failing early. They are consistent now.

sergenyalcin:
@ulucinar I've addressed your all comments. Thanks for your review.

ulucinar (approved):
Thanks @sergenyalcin for the CRD storage version migrator implementation. Let's get this integrated to the GCP v3 provider!

PR #593, opened 2026-01-29
fix(tfjson): Reset Computed flag for SchemaNestingModeSingle blocks with required children

rwwiv:
## Summary

Fixes two issues where `SchemaNestingModeSingle` blocks from Plugin Framework resources were incorrectly handled:

1. **Computed inference bug**: Blocks were incorrectly marked as `Computed=true`, causing them to be excluded from `ForProvider`/`InitProvider` parameters and only appear in `Observation`.

2. **Type representation bug**: Blocks were using `TypeList` instead of `SchemaTypeObject`, causing the generated CRD schema to expect arrays instead of objects.

## Problem

In `tfJSONBlockTypeToV2Schema()`, when `MinItems==0 && MaxItems==0`, the block was being marked as `Computed=true`:

```go
if nb.MinItems == 0 && nb.MaxItems == 0 {
    v2sch.Computed = true
}
```

This heuristic works for SDK v2 collection types (List/Set/Map) where `MinItems=0, MaxItems=0` typically indicates a computed-only collection. However, Plugin Framework resources using `SchemaNestingModeSingle` default to `MinItems=0, MaxItems=0` even for user-configurable blocks.

Additionally, `SchemaNestingModeSingle` blocks were using `schemav2.TypeList`, but according to Terraform Plugin Framework documentation, SingleNestedBlock values should be represented by an object type, not a list.

This caused blocks like `metadata` and `spec` to be:
- Marked `Computed=true`, which makes `IsObservation()` return `true`, excluding them from the generated Parameters struct
- Generated as arrays in the CRD schema, when they should be objects

## Solution

1. Move the `Computed=true` inference to only apply to collection nesting modes (Set/List/Map), not to `SchemaNestingModeSingle`. For single blocks, we now rely solely on `hasRequiredChild()` to determine Required/Optional, leaving Computed=false by default.

2. Use `SchemaTypeObject` instead of `TypeList` for `SchemaNestingModeSingle` blocks, matching the behavior for nested attributes with `SchemaNestingModeSingle` in `tfJSONNestedAttributeTypeToV2Schema()`.

```go
case tfjson.SchemaNestingModeSingle, tfjson.SchemaNestingModeGroup:
    // For SchemaNestingModeSingle (Plugin Framework), we do NOT infer
    // Computed=true from MinItems/MaxItems==0, because Plugin Framework
    // resources default to 0 for these values even for user-configurable blocks.
    // Use SchemaTypeObject to generate an embedded object (not a list),
    // matching the behavior for nested attributes with SchemaNestingModeSingle
    // and the Terraform Plugin Framework documentation which states that
    // SingleNestedBlock values are represented by an object type.
    v2sch.Type = SchemaTypeObject
    v2sch.Required = hasRequiredChild(nb)
    v2sch.Optional = !v2sch.Required
    // Computed remains false (the default)
```

## Testing

Added comprehensive unit tests for `tfJSONBlockTypeToV2Schema` and `hasRequiredChild`:

- `SchemaNestingModeSingle` with required children → `Computed=false`, `Required=true`, `Type=SchemaTypeObject`
- `SchemaNestingModeSingle` with only optional children → `Computed=false`, `Optional=true`, `Type=SchemaTypeObject`
- `SchemaNestingModeSingle` with empty/nil block
- `SchemaNestingModeSingle` with nested blocks containing required children
- `SchemaNestingModeList/Set/Map` with `MinItems=0, MaxItems=0` → `Computed=true` (preserves existing behavior)
- `SchemaNestingModeList` with `MinItems=1` → `Computed=false`

## Impact

This fix is particularly important for Terraform Plugin Framework resources. For example, the Grafana provider's App Platform resources (`grafana_apps_rules_alertrule_v0alpha1`, `grafana_apps_rules_recordingrule_v0alpha1`, etc.) have `metadata` and `spec` blocks that were incorrectly excluded from `forProvider` in the generated CRDs, and were being generated as arrays instead of objects.

**Before this fix:**
```yaml
spec:
  forProvider:
    options: {}  # Only options appeared, metadata/spec excluded
    # If they did appear, they would be arrays: metadata: [{}]
```

**After this fix:**
```yaml
spec:
  forProvider:
    metadata:        # Now an object, not an array
      uid: "..."
      folderUid: "..."
    spec:            # Now an object, not an array
      title: "..."
      expressions: {}
      # ... all other fields
    options: {}
```

SDK v2-based resources are unaffected because they typically use `MinItems=1, MaxItems=1` for required single blocks, so they never hit the `MinItems == 0 && MaxItems == 0` path.

### DIFF
--- a/pkg/types/conversion/tfjson/tfjson.go
+++ b/pkg/types/conversion/tfjson/tfjson.go
@@ -181,7 +181,11 @@ func tfJSONBlockTypeToV2Schema(nb *tfjson.SchemaBlockType) *schemav2.Schema { //
 		// based on whether the block has required children.
 		// Note: Group is similar to Single but guarantees non-null result even
 		// when the block is absent.
-		v2sch.Type = schemav2.TypeList
+		// Use SchemaTypeObject to generate an embedded object (not a list),
+		// matching the behavior for nested attributes with SchemaNestingModeSingle
+		// and the Terraform Plugin Framework documentation which states that
+		// SingleNestedBlock values are represented by an object type.
+		v2sch.Type = SchemaTypeObject
 		v2sch.MinItems = 0
 		v2sch.Required = hasRequiredChild(nb)
 		v2sch.Optional = !v2sch.Required
--- a/pkg/types/conversion/tfjson/tfjson.go
+++ b/pkg/types/conversion/tfjson/tfjson.go
@@ -150,23 +150,39 @@ func tfJSONBlockTypeToV2Schema(nb *tfjson.SchemaBlockType) *schemav2.Schema { //
 	if nb.MinItems == 0 {
 		v2sch.Optional = true
 	}
-	if nb.MinItems == 0 && nb.MaxItems == 0 {
-		v2sch.Computed = true
-	}
 
 	switch nb.NestingMode { //nolint:exhaustive
 	case tfjson.SchemaNestingModeSet:
 		v2sch.Type = schemav2.TypeSet
+		// For collection types (Set/List/Map), infer Computed when MinItems and
+		// MaxItems are both 0, following SDK v2 semantics.
+		if nb.MinItems == 0 && nb.MaxItems == 0 {
+			v2sch.Computed = true
+		}
 	case tfjson.SchemaNestingModeList:
 		v2sch.Type = schemav2.TypeList
+		// For collection types (Set/List/Map), infer Computed when MinItems and
+		// MaxItems are both 0, following SDK v2 semantics.
+		if nb.MinItems == 0 && nb.MaxItems == 0 {
+			v2sch.Computed = true
+		}
 	case tfjson.SchemaNestingModeMap:
 		v2sch.Type = schemav2.TypeMap
-	case tfjson.SchemaNestingModeSingle:
+		// For collection types (Set/List/Map), infer Computed when MinItems and
+		// MaxItems are both 0, following SDK v2 semantics.
+		if nb.MinItems == 0 && nb.MaxItems == 0 {
+			v2sch.Computed = true
+		}
+	case tfjson.SchemaNestingModeSingle, tfjson.SchemaNestingModeGroup:
+		// For SchemaNestingModeSingle and SchemaNestingModeGroup (Plugin
+		// Framework), we do NOT infer Computed=true from MinItems/MaxItems==0,
+		// because Plugin Framework resources default to 0 for these values even
+		// for user-configurable blocks. Instead, we determine Required/Optional
+		// based on whether the block has required children.
+		// Note: Group is similar to Single but guarantees non-null result even
+		// when the block is absent.
 		v2sch.Type = schemav2.TypeList
 		v2sch.MinItems = 0
-		// TODO(erhan): not sure whether we need this
-		// the block itself can be optional, even if some child attribute
-		// or block is required
 		v2sch.Required = hasRequiredChild(nb)
 		v2sch.Optional = !v2sch.Required
 		if v2sch.Required {
--- a/pkg/types/conversion/tfjson/tfjson.go
+++ b/pkg/types/conversion/tfjson/tfjson.go
@@ -147,6 +147,10 @@ func tfJSONBlockTypeToV2Schema(nb *tfjson.SchemaBlockType) *schemav2.Schema { //
 	// https://github.com/hashicorp/terraform-plugin-sdk/blob/6461ac6e9044a44157c4e2c8aec0f1ab7efc2055/helper/schema/core_schema.go#L204
 	v2sch.Computed = false
 	v2sch.Optional = false
+	if nb.MinItems > 0 {
+		v2sch.Required = true
+		v2sch.Optional = false
+	}
 	if nb.MinItems == 0 {
 		v2sch.Optional = true
 	}
@@ -154,24 +158,29 @@ func tfJSONBlockTypeToV2Schema(nb *tfjson.SchemaBlockType) *schemav2.Schema { //
 		v2sch.Computed = true
 	}
 
-	switch nb.NestingMode { //nolint:exhaustive
+	switch nb.NestingMode {
 	case tfjson.SchemaNestingModeSet:
 		v2sch.Type = schemav2.TypeSet
 	case tfjson.SchemaNestingModeList:
 		v2sch.Type = schemav2.TypeList
 	case tfjson.SchemaNestingModeMap:
 		v2sch.Type = schemav2.TypeMap
-	case tfjson.SchemaNestingModeSingle:
-		v2sch.Type = schemav2.TypeList
-		v2sch.MinItems = 0
-		// TODO(erhan): not sure whether we need this
-		// the block itself can be optional, even if some child attribute
-		// or block is required
-		v2sch.Required = hasRequiredChild(nb)
-		v2sch.Optional = !v2sch.Required
-		if v2sch.Required {
-			v2sch.MinItems = 1
-		}
+	case tfjson.SchemaNestingModeSingle, tfjson.SchemaNestingModeGroup:
+		// This is a Plugin Framework-only nesting mode, and
+		// FW schemas never get Min/MaxItems specified in their
+		// core schemas (tfjson schema).
+		// See https://github.com/hashicorp/terraform-plugin-framework/blob/a0219204842978493e5f7742b0c06d5c39951e73/internal/fwschema/block.go#L24
+		// Therefore, the heuristics for determining Required/Optional/Computed
+		// does not make sense and inconclusive. Always make them optional,
+		// so that they generate a configurable spec field in the CRD.
+		// This only affects the CRD generation. Runtime schema validations
+		// are still valid and relevant.
+		// Provider developers can manually override these in their provider,
+		// if they want observation-only, or make them Required.
+		v2sch.Type = SchemaTypeObject
+		v2sch.Required = false
+		v2sch.Optional = true
+		v2sch.Computed = false
 		v2sch.MaxItems = 1
 	default:
 		panic("unhandled nesting mode: " + nb.NestingMode)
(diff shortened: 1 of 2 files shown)


pkg/types/conversion/tfjson/tfjson.go
erhancagirici, line 193: The actual bug is here:
As I mentioned in the main review comment, we initially start with
`Required: false, Optional: true, Computed: true` for all blocks in plugin-fw resources.

If we determine `required: true`, we switch `Optional: False` but computed remains `true`. i.e. we end up with 
`Required: true, Optional: false, Computed: true`
This is actually an [invalid schema configuration](https://github.com/hashicorp/terraform-plugin-sdk/blob/2f989a66c2b39ffc680041793fb30b456195e882/helper/schema/schema.go#L955). A required field cannot be computed.

However, this ends up in generation pipeline and [IsObservation()](https://github.com/crossplane/upjet/blob/dbfccb48a1c47e728eb91afb9aee7457cf9bfc05/pkg/types/builder.go#L515-L520) check treats this as an computed-only field (because computed=true, optional=false)

In summary, adding the following aligns it with the rest of the block types and starts generating those.
```suggestion
			v2sch.Computed = false
```

However, please also see comment for the broader situation on this and a proposed change: https://github.com/crossplane/upjet/pull/593/changes#r3871736101

pkg/types/conversion/tfjson/tfjson.go
```diff
@@ -150,23 +150,43 @@ func tfJSONBlockTypeToV2Schema(nb *tfjson.SchemaBlockType) *schemav2.Schema { //
...
+		// Use SchemaTypeObject to generate an embedded object (not a list),
+		// matching the behavior for nested attributes with SchemaNestingModeSingle
+		// and the Terraform Plugin Framework documentation which states that
+		// SingleNestedBlock values are represented by an object type.
+		v2sch.Type = SchemaTypeObject
 		v2sch.MinItems = 0
-		// TODO(erhan): not sure whether we need this
-		// the block itself can be optional, even if some child attribute
-		// or block is required
 		v2sch.Required = hasRequiredChild(nb)
```
erhancagirici, line 170: following up my previous todo:

The TF Core schema has no notion of required/optional/computed for a "block" field. Only Min/MaxItems. And they are (intentionally) not available for plugin-fw resources.
So, we have no way of inferring whether a block is "Observation"-only or intended to be configurable. plugin-fw [doc comment](https://github.com/hashicorp/terraform-plugin-framework/blob/a0219204842978493e5f7742b0c06d5c39951e73/internal/fwschema/block.go#L24) mentions limited validation capabilities at config time, and says to offload this validation to the provider side.

Additionally, checking the TF SDKv2 code [here](https://github.com/hashicorp/terraform-plugin-sdk/blob/9a14d44996827b9028aa4f3a2422541ce8bbd92e/helper/schema/core_schema.go#L296) -> [here](https://github.com/hashicorp/terraform-plugin-sdk/blob/9a14d44996827b9028aa4f3a2422541ce8bbd92e/helper/schema/core_schema.go#L184), sdk resources never end up with `NestingSingle` block in their core schema, the only exception being the `timeout` block, which upjet skips anyway.
So, I think it is safe to assume this path is only visited by plugin-fw resource blocks.

Lastly, this converted "v2schema" is only for CRD generation purposes and not utilized at runtime. It acts like an Intermediate representation for seeding CRD generation. e.g. the computed/optional etc. 

Considering all of the above, I think `hasRequiredChild` is not a proper heuristic for requiredness. A required child attribute not necessarily mean that the block itself is required.
I propose to treat all `NestingSingle` blocks as optional for CRD generation (so that it always generates a nullable `forProvider.myFooField` ) and let the runtime validate it.

Also, this won't be a breaking change since this path was never generating a required forProvider field anyway, and optionals will just stay the same.

Also, upjet has already config machinery available for explicitly marking the field observe-only or required, so devs can modify according to their needs if they need stricter CRD API validation on this.

TLDR, my suggestion is to have:
```golang
	case tfjson.SchemaNestingModeSingle, tfjson.SchemaNestingModeGroup:
		v2sch.Type = SchemaTypeObject
		v2sch.Required = false
		v2sch.Optional = true
		v2sch.Computed = false
```

pkg/types/conversion/tfjson/tfjson.go
```diff
@@ -150,23 +150,43 @@ func tfJSONBlockTypeToV2Schema(nb *tfjson.SchemaBlockType) *schemav2.Schema { //
...
+		// because Plugin Framework resources default to 0 for these values even
+		// for user-configurable blocks. Instead, we determine Required/Optional
+		// based on whether the block has required children.
+		// Note: Group is similar to Single but guarantees non-null result even
+		// when the block is absent.
+		// Use SchemaTypeObject to generate an embedded object (not a list),
+		// matching the behavior for nested attributes with SchemaNestingModeSingle
+		// and the Terraform Plugin Framework documentation which states that
+		// SingleNestedBlock values are represented by an object type.
+		v2sch.Type = SchemaTypeObject
```
erhancagirici, line 188: note: `SchemaTypeObject` is actually a workaround in upjet to represent plugin-fw schemas in sdkv2 schema structs. This actually leads in an invalid sdkv2 schema struct, and calling member functions like `CoreConfigSchema()` on the resulting `TerraformResource` causes a panic.

Though for plugin-framework resources, this object is CRD-generation purposes only and never utilized at runtime, and this path is framework-only ( as mentioned in https://github.com/crossplane/upjet/pull/593/changes#r3871736101 ), so I think this is acceptable.

pkg/types/conversion/tfjson/tfjson.go
```diff
@@ -150,23 +150,43 @@ func tfJSONBlockTypeToV2Schema(nb *tfjson.SchemaBlockType) *schemav2.Schema { //
 	if nb.MinItems == 0 {
 		v2sch.Optional = true
 	}
-	if nb.MinItems == 0 && nb.MaxItems == 0 {
-		v2sch.Computed = true
-	}
```
erhancagirici, line 155: as mentioned in other comments, let's put this back and fix the `NestingSingle` branch only.

pkg/types/conversion/tfjson/tfjson.go
```diff
@@ -150,23 +150,43 @@ func tfJSONBlockTypeToV2Schema(nb *tfjson.SchemaBlockType) *schemav2.Schema { //
...
 
 	switch nb.NestingMode { //nolint:exhaustive
 	case tfjson.SchemaNestingModeSet:
 		v2sch.Type = schemav2.TypeSet
+		// For collection types (Set/List/Map), infer Computed when MinItems and
+		// MaxItems are both 0, following SDK v2 semantics.
+		if nb.MinItems == 0 && nb.MaxItems == 0 {
+			v2sch.Computed = true
+		}
```
erhancagirici, line 161: see https://github.com/crossplane/upjet/pull/593/changes#r3871862997

pkg/types/conversion/tfjson/tfjson.go
```diff
@@ -150,23 +150,43 @@ func tfJSONBlockTypeToV2Schema(nb *tfjson.SchemaBlockType) *schemav2.Schema { //
...
 	case tfjson.SchemaNestingModeList:
 		v2sch.Type = schemav2.TypeList
+		// For collection types (Set/List/Map), infer Computed when MinItems and
+		// MaxItems are both 0, following SDK v2 semantics.
+		if nb.MinItems == 0 && nb.MaxItems == 0 {
+			v2sch.Computed = true
+		}
```
erhancagirici, line 168: https://github.com/crossplane/upjet/pull/593/changes#r3871862997

pkg/types/conversion/tfjson/tfjson.go
```diff
@@ -150,23 +150,43 @@ func tfJSONBlockTypeToV2Schema(nb *tfjson.SchemaBlockType) *schemav2.Schema { //
...
 	case tfjson.SchemaNestingModeMap:
 		v2sch.Type = schemav2.TypeMap
-	case tfjson.SchemaNestingModeSingle:
-		v2sch.Type = schemav2.TypeList
+		// For collection types (Set/List/Map), infer Computed when MinItems and
+		// MaxItems are both 0, following SDK v2 semantics.
+		if nb.MinItems == 0 && nb.MaxItems == 0 {
+			v2sch.Computed = true
+		}
```
erhancagirici, line 175: see https://github.com/crossplane/upjet/pull/593/changes#r3871862997

erhancagirici (commented):
@rwwiv thanks for reporting and the PR on this. I've investigated this and below are my findings and notes:

Per this doc comment https://github.com/hashicorp/terraform-plugin-framework/blob/a0219204842978493e5f7742b0c06d5c39951e73/internal/fwschema/block.go#L24 looks like we never get `Min/MaxItems` data for blocks at the corresponding TF Core schema for plugin-framework resource schemas. This is true for all block nesting types .
As a result of this, today, the inference/heuristic always ends up with `Optional: true Computed: true Required: false` for all plugin-FW blocks. 

In fact, the TF core block fields have no direct notion of "required,optional,computed"ness.

This currently works out `SchemaNestingModeSet` `SchemaNestingModeList` `SchemaNestingModeMap` for fw resources, they always get a spec.forProvider field in the CRD, with the implication that they are always optional. Which is fine

The actual bug resides in the `SchemaNestingModeSingle` here, please see my comment: https://github.com/crossplane/upjet/pull/593/changes#r3871860815.

on the broader situation: please see https://github.com/crossplane/upjet/pull/593/changes#r3871736101

I'll do a final review after giving it some testing with the existing providers.

erhancagirici (approved):
Thanks again for this PR! I've updated the implementation to make `NestingModeSingle` blocks always optional per my previous [comment](https://github.com/crossplane/upjet/pull/593#discussion_r3871736101), as the heuristic never makes sense for those. `Optional` is the sanest default.

If `Required` or `Computed`-only behavior is desired, these can be still overridden at the provider. 
This only affects the generated CRD, runtime TF schema will still comply/enforce validation.

Issue #559, opened 2025-11-27
Allow something to be Sensitive but also passable as plaintext


Breee:
I have the case that i have properties in my upjet generated provider that i would like to be passable as secret but also as plaintext.

For example. currently we have this

```
    p.AddResourceConfigurator("keycloak_oidc_identity_provider", func(r *config.Resource) {
             
        // Mark fields as sensitive
        if s, ok := r.TerraformResource.Schema["client_id"]; ok {
            s.Sensitive = true
        }
        if s, ok := r.TerraformResource.Schema["client_secret"]; ok {
            s.Sensitive = true
        }
    })
```

This will require everyone to pass clientID and clientSecret as a secret. 
I would like boolean property that allows me to have both. SecretRefs  and Plaintext. 

something like: 

```diff
    p.AddResourceConfigurator("keycloak_oidc_identity_provider", func(r *config.Resource) {
++      r.Sensitive.AllowPlainTextValue = true
        // Mark fields as sensitive
        if s, ok := r.TerraformResource.Schema["client_id"]; ok {
            s.Sensitive = true
        }
        if s, ok := r.TerraformResource.Schema["client_secret"]; ok {
            s.Sensitive = true
        }
    })
```

what do you think?

jonasz-lasut:
Hi @Breee thank you for opening the issue.

After thorough evaluation the maintainer team is against including this feature into upjet.
Providers should be opinionated on which fields are sensitive or not sensitive and configure them appropriately (or use the default sensitivity from terraform provider). The proposed scoping is also too broad as a single per-resource flag configures all present and future `Sensitive` fields to generate both secret references and plain string fields.

We propose to use compositions to satisfy this use-case if needed. XR that builds on top of the resource that should have the sensitive field in both plain and secret forms could expose a string value field that in the composition is used to create a secret and use it as a referenced secret in a managed resource.

Breee:
I think Opinionated decisions are great, in this case this detail is a breaking change for all users, that's why it is until this day not possible to pass a client_id via secret. 

In my case of an identity provider like keycloak, the information of client_id, client_secret, base URLs etc is usually shipped as a single secret using external secret Operator or whatever people use. It's not necessarily a bunch of "secure" values. 

My goal is always to avoid breaking changes and have as much convenience for my users as possible. 

A composition works here, but it's also alot of overhead and just a workaround. (that's 100+ lines yaml to inject a string from a secret into a manifest) 

[...]

PR #610, opened 2026-03-12
feat(issue-559): allow sensitive values to be passable as plaintext

Breee:
# allow sensitive values to be passable as plaintext

Draft for #559 to have a base for discussion.  This shall give you an idea of the requirement i have.
This was the easiest implementation approach I found that is fully backwards compatible. 
However, it does add some complexity to the builder logic and generated code, so I'm open to suggestions on how to make it cleaner or if we want to take a different approach.
In general i need some more flexibility with what i can pass in as secret and what i can pass in as plaintext for the provider-keycloak, as OIDC / SAML clients and tools vary alot in what they support and need.

## Problem

Sensitive fields currently only support secret references. Users must create a Kubernetes Secret even if they just want to pass a value directly. This is inflexible and doesn't match how the underlying Terraform providers work (which accept plaintext).

## Potential Solution

I Added a `AllowPlaintextValue` config flag. When enabled, generates both the regular field AND the secret reference field for sensitive parameters. Users pick which one to use via OR validation (provide one or the other, not both).

Runtime checks the plaintext field first, falls back to secret reference if not set. Its Backwards compatible so existing secret-only resources work unchanged.

## Usage

Provider configuration:
```go
p.AddResourceConfigurator("keycloak_oidc_identity_provider", func(r *config.Resource) {
    r.ShortGroup = "oidc"
    
    // Mark fields as sensitive in Terraform schema
    if s, ok := r.TerraformResource.Schema["client_id"]; ok {
        s.Sensitive = true
    }
    if s, ok := r.TerraformResource.Schema["client_secret"]; ok {
        s.Sensitive = true
    }
    
    // Enable plaintext AND secret reference support
    r.Sensitive.AllowPlaintextValue = true
})
```
See https://github.com/crossplane-contrib/provider-keycloak/blob/feat/upjet-test/config/oidc/config.go#L30

Generated CRD types will then look like this:
```go
type IdentityProviderParameters struct {
    // [...more stuff...]
    // Plaintext field (new!)
    ClientID *string `json:"clientId,omitempty"`
    
    // Secret reference field (existing)
    ClientIDSecretRef v1.SecretKeySelector `json:"clientIdSecretRef,omitempty"`
    
    ClientSecret *string `json:"clientSecret,omitempty"`
    ClientSecretSecretRef *v1.SecretKeySelector `json:"clientSecretSecretRef,omitempty"`
   // [...more stuff...]
}
```
see https://github.com/crossplane-contrib/provider-keycloak/blob/feat/upjet-test/apis/cluster/oidc/v1alpha1/zz_identityprovider_types.go#L43

## Example

After enabling `AllowPlaintextValue`, users can choose:
```yaml
# Option 1: Plaintext (new)
spec:
  forProvider:
    clientId: "my-client-id"
    clientSecret: "my-secret"

# Option 2: Secret references 
spec:
  forProvider:
    clientIdSecretRef:
      name: my-secret
      key: client_id
    clientSecretSecretRef:
      name: my-secret
      key: client_secret
      [... other fields ...]
```

## Testing

I Tested in provider-keycloak with both plaintext and secret reference patterns.
Both resources successfully created and synced in a dev cluster.

### DIFF
--- a/pkg/config/resource.go
+++ b/pkg/config/resource.go
@@ -244,6 +244,12 @@ type Sensitive struct {
 	// fieldPaths keeps the mapping of sensitive fields in Terraform schema with
 	// terraform field path as key and xp field path as value.
 	fieldPaths map[string]string
+
+	// AllowPlaintextValue allows sensitive fields to be passed as plaintext
+	// in addition to secret references. When true, sensitive fields will be
+	// generated as both a regular field and a secret reference field, giving
+	// users the option to provide values directly or via secrets.
+	AllowPlaintextValue bool
 }
 
 // LateInitializer represents configurations that control
--- a/pkg/resource/sensitive.go
+++ b/pkg/resource/sensitive.go
@@ -182,6 +182,14 @@ func GetSensitiveParameters(ctx context.Context, client SecretClient, from resou
 			prefixes = []string{""}
 		}
 
+		// Check if plaintext value already exists in Terraform state
+		// This happens when AllowPlaintextValue is enabled and user provided plaintext
+		existingValue, err := pavedTF.GetValue(tfPath)
+		if err == nil && existingValue != nil {
+			// Plaintext value exists, skip secret reference handling
+			continue
+		}
+
 		// spec.forProvider secret references override the spec.initProvider
 		// references.
 		for _, p := range prefixes {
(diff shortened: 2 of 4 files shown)


vrabbi:
this would be extremely helpful to have

Breee:
I'll rebase onto Main and search in the upjet chat for someone that gives Feedback

PR #623, opened 2026-03-27
feat: add resource identity support to TF Plugin Framework integration

### LINKED ISSUE #2003: [Bug]: Unable to create securitygroupegressrule or securitygroupingressrule in v2.5.0
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Affected Resource(s)

* securitygroupegressrule.ec2.aws.upbound.io/v1beta1
* securitygroupingressrule.ec2.aws.upbound.io/v1beta1

### Resource MRs required to reproduce the bug

```
apiVersion: ec2.aws.upbound.io/v1beta1
kind: SecurityGroupEgressRule
metadata:
  annotations:
    crossplane.io/composition-resource-name: <nameq>
  generateName: <generate_name>
  labels:
    crossplane.io/claim-name: <name>
    crossplane.io/claim-namespace: <namespace>
    crossplane.io/composite: <composite_name>
  name: <name>
  ownerReferences: [...]
 spec:
  deletionPolicy: Delete
  forProvider:
    cidrIpv4: 0.0.0.0/0
    description: <desc>
    ipProtocol: "-1"
    region: eu-west-3
    securityGroupId: <sg_id>
    securityGroupIdRef:
      name: <sg_name_ref>
  initProvider: {}
  managementPolicies:
  - '*'
  providerConfigRef:
    name: <provider_ref>
```

### Steps to Reproduce

When trying to create security rules (ingress or egress) in new provider version (v2.5.0), i've got the following error:
    message: 'observe failed: read resource request failed: Missing Resource Identity
      After Read: The Terraform Provider unexpectedly returned no resource identity
      data after having no errors in the resource read. This is always an issue in
      the Terraform Provider and should be reported to the provider developers.'

Downgrading to previous version (v2.4.0) resolve the issue, so I think it's related to terraform provider bump in last release, and probably to this issue on terraform side:
https://github.com/hashicorp/terraform-provider-aws/issues/45303 

### What happened?

In previous version (v2.4.0) ingress or egress security rules were created successfully.

### Relevant Error Output Snippet

```shell
There is no error log in crossplane pods, only the error message in the managed resources.
```

### Crossplane Version

1.20.4

### Provider Version

2.5.0

### Kubernetes Version

v1.34.0

### Kubernetes Distribution

Vanilla

### Additional Info

_No response_

erikmiller-gusto:

Pass resource identity data through all protov6 request/response cycles (ReadResource, PlanResourceChange, ApplyResourceChange) so that resources declaring an IdentitySchema in terraform-plugin-framework v1.17.0+ receive the required identity data and no longer fail with "Missing Resource Identity After Read".

Fixes crossplane-contrib/provider-upjet-aws#2003

I have:

### DIFF
--- a/pkg/controller/external_tfpluginfw.go
+++ b/pkg/controller/external_tfpluginfw.go
@@ -483,6 +483,25 @@ func (n *terraformPluginFrameworkExternalClient) hasResourceNotFoundDiagnostic(d
 	}
 	return n.config.ExternalName.IsNotFoundDiagnosticFn(diags)
 }
+
+// diagSummaryMissingResourceIdentity is the diagnostic summary returned by
+// terraform-plugin-framework when a resource with an IdentitySchema has its
+// state set to null (resource deleted externally or not yet created) but the
+// provider's identity interceptor skips populating identity for null states.
+const diagSummaryMissingResourceIdentity = "Missing Resource Identity After Read"
+
+// hasMissingResourceIdentityDiagnostic checks whether the diagnostics contain
+// the "Missing Resource Identity After Read" error. We treat this as a
+// resource-not-found condition.
+func hasMissingResourceIdentityDiagnostic(diags []*tfprotov6.Diagnostic) bool {
+	for _, d := range diags {
+		if d.Severity == tfprotov6.DiagnosticSeverityError && d.Summary == diagSummaryMissingResourceIdentity {
+			return true
+		}
+	}
+	return false
+}
+
 func (n *terraformPluginFrameworkExternalClient) Observe(ctx context.Context, mg xpresource.Managed) (managed.ExternalObservation, error) { //nolint:gocyclo
 	n.logger.Debug("Observing the external resource")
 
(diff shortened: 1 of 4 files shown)


erikmiller-gusto:
Have also confirmed that this works and fixed the same issue in the provider-upjet-cloudflare

aditmeno:
Thanks for the PR — the identity propagation fix is necessary but not sufficient on its own. We tested it in production and hit the same `Missing Resource Identity After Read` error on `aws_s3_bucket_lifecycle_configuration` (and likely any resource with an `IdentitySchema` that gets deleted externally or doesn't yet exist).

## Root Cause

The PR's comment in `nofork_store.go` assumes:

> after a restart, the first Observe sends nil CurrentIdentity, and the provider's identity interceptor populates NewIdentity from state/client data in the ReadResourceResponse, rehydrating the tracker.

This is only true **when the resource still exists**. When the resource has been deleted externally (or hasn't been created yet):

1. The provider's `Read` sets `response.State` to null (resource gone)
2. The AWS provider's identity interceptor ([`identity_interceptor.go`](https://github.com/hashicorp/terraform-provider-aws/blob/main/internal/provider/framework/identity_interceptor.go)) checks `response.State.Raw.IsNull()` and **skips populating identity**
3. The `terraform-plugin-framework` server ([`server_readresource.go` L201-209](https://github.com/hashicorp/terraform-plugin-framework/blob/v1.17.0/internal/fwserver/server_readresource.go#L201-L209)) validates that identity must be non-null after a successful Read — and returns the `Missing Resource Identity After Read` error diagnostic

## Additional Fix Needed

In `Observe()`, the `Missing Resource Identity After Read` diagnostic should be treated as a **resource-not-found** condition (similar to how `isResourceNotFoundDiags` already works). Here's the change we applied on top of this PR:

```go
// hasMissingResourceIdentityDiagnostic checks whether the diagnostics contain
// the "Missing Resource Identity After Read" error that terraform-plugin-framework
// returns when a resource with an IdentitySchema has its state set to null
// (resource deleted externally) but the identity interceptor skips populating
// identity for null states. We treat this as a resource-not-found condition.
func hasMissingResourceIdentityDiagnostic(diags []*tfprotov6.Diagnostic) bool {
	for _, d := range diags {
		if d.Severity == tfprotov6.DiagnosticSeverityError && d.Summary == "Missing Resource Identity After Read" {
			return true
		}
	}
	return false
}
```

And in `Observe()`, alongside the existing `isResourceNotFoundDiags` check:

```go
isResourceNotFoundDiags := n.hasResourceNotFoundDiagnostic(readResponse.Diagnostics)
isMissingIdentityDiags := hasMissingResourceIdentityDiagnostic(readResponse.Diagnostics)
if fatalDiags := getFatalDiagnostics(readResponse.Diagnostics); fatalDiags != nil {
    if !isResourceNotFoundDiags && !isMissingIdentityDiags {
        n.opTracker.ResetReconstructedFrameworkTFState()
        return managed.ExternalObservation{}, errors.Wrap(fatalDiags, "read resource request failed")
    }
    if isResourceNotFoundDiags {
        n.logger.Debug("TF ReadResource returned error diagnostics, but XP resource was configured to treat them as `Resource not exists`. Skipping", "skippedDiags", fatalDiags)
    }
    if isMissingIdentityDiags {
        n.logger.Debug("TF ReadResource returned Missing Resource Identity diagnostic, treating as resource not found", "skippedDiags", fatalDiags)
        isResourceNotFoundDiags = true
    }
}
```

Without this, any Plugin Framework resource with an `IdentitySchema` will fail on first reconciliation (resource doesn't exist yet) or when the external resource is deleted out-of-band.

erikmiller-gusto:
@aditmeno I've updated the PR to address that.  Will also run that through my own tests to confirm

erikmiller-gusto:
@aditmeno was able to confirm the external deletion flow on the aws Bucketlifecycleconfiguration

erhancagirici (approved):
@erikmiller-gusto thanks for contribution 🎉 I have added a commit that identity-related data is only passed when resource supports identity to further guardrail.

Tested with AWS provider with the affected resources and LGTM

PR #635, opened 2026-04-27
fix: set dotted map keys with correct nesting

erhancagirici:

When a `SecretReference` is used to populate an MR attribute of `map[string]string` type, keys of secret data become the map keys. When the key have dots, `fieldpath` splits the key at every dot, producing a wronly-nested object, instead of a flat string to string map entry.

The fix switches to bracket notation (`foo_attribute[bar.with.dot]`), which preserves the full key string as a single path segment.

Addresses https://github.com/crossplane-contrib/provider-upjet-aws/issues/1946

I have:

Consuming upjet from this PR in `provider-upjet-aws` with the reproducer at https://github.com/crossplane-contrib/provider-upjet-aws/issues/1946

### DIFF
--- a/pkg/controller/external_tfpluginsdk.go
+++ b/pkg/controller/external_tfpluginsdk.go
@@ -221,9 +221,14 @@ func (c *TerraformPluginSDKConnector) applyHCLParserToParam(sc *schema.Schema, p
 			}
 		}
 	case schema.TypeString:
+		paramStr, ok := param.(string)
+		if !ok {
+			c.logger.Debug("expected parameter value to be string", "parameterType", fmt.Sprintf("%T", param))
+			return param
+		}
 		// For String types check if it is an HCL string and process
-		if isHCLSnippetPattern.MatchString(param.(string)) {
-			hclProccessedParam, err := processHCLParam(param.(string))
+		if isHCLSnippetPattern.MatchString(paramStr) {
+			hclProccessedParam, err := processHCLParam(paramStr)
 			if err != nil {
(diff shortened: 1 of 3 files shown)


ulucinar (approved):
Thank you @erhancagirici for fixing this ancient bug!

When consuming this in `crossplane-contrib/provider-aws-upjet`, let's also add a regression (up)test in that repo, in addition to the unit tests we introduce with this PR.

Looking at the AWS Apache [Airflow configuration guide](https://docs.aws.amazon.com/mwaa/latest/userguide/configuring-env-variables.html#configuring-env-variables-reference), I wonder whether we will be able to handle non-string configuration parameters (seems like not all configuration options are strings but I did not check the AWS SDK).

We don't need to block this PR (as the fix it introduces is solid) but let's check things on the AWS side. We can discuss how to proceed if non-string configuration options are not working in the consumption PR.

erhancagirici:
> Looking at the AWS Apache [Airflow configuration guide](https://docs.aws.amazon.com/mwaa/latest/userguide/configuring-env-variables.html#configuring-env-variables-reference), I wonder whether we will be able to handle non-string configuration parameters (seems like not all configuration options are strings but I did not check the AWS SDK).

@ulucinar for this particular resource and any other with sensitive map fields in TF SDKv2, the underlying TF SDKv2 schema is always map of string to string (or more precisely `primitive`), i.e TF SDKv2 schema anyways won't allow non-primitive(e.g nested objects) values in maps by SDKv2 design, regardless of the AWS API.

PR #636, opened 2026-04-28
feat: API roundtrip testing library 

erhancagirici:

This PR introduces `pkg/apitesting/roundtrip`, a reusable testing utility library, that providers built on upjet
can drop into their test suites to automatically verify the roundtrip correctness of their generated API types.

Similar to the approach in the Kubernetes ecosystem, the library utilizes fuzz-based round-trip testing: generate
randomly filled objects and assert they survive the full encode/decode or conversion cycle unchanged.
This library extracts that common boilerplate into a single, maintained place so providers can opt in with minimal setup.

Two categories of invariants are covered:

- **Serialization** — every registered API type can be JSON-encoded and decoded back to an identical object. This catches incorrect field types, and other schema mismatches that would cause silent data loss when Kubernetes stores or returns a resource.
- **Version conversion** — every multi-version type survives a full hub↔spoke round trip (`spoke → hub → spoke` and `hub → spoke → hub`) with no data loss.
- This ensures that upjet's generated conversion functions faithfully preserve all fields across API versions.

Note: The tests focus on the serialization/conversion machinery aspect.
The test data is fuzz-based and does not focus on "semantically valid" API objects of the corresponding external provider API.  

### How it works

**Serialization test**

This test utilizes upstream k8s "k8s.io/apimachinery/pkg/api/apitesting/roundtrip" for serialization tests.

1. Discover all API types registered in the scheme (respecting include/exclude filters).
2. For each type, generate a randomly filled instance using the fuzzer.
3. Encode the object to JSON and decode it back into a new instance.
4. Assert the two instances are equal.

**Conversion test**

1. Discover all API groups and GroupKinds registered in the scheme.
2. For each GroupKind with more than one version, identify which version is the hub and which are spokes (based on the `conversion.Hub` / `conversion.Convertible` interfaces).
   - Skips `GroupKind`s with one available version
3. For every hub↔spoke pair, run two sub-tests:
   - **spoke → hub → spoke**: fill a spoke object, convert to hub, convert back to a new spoke, compare.
   - **hub → spoke → hub**: fill a hub object, convert to spoke, convert back to a new hub, compare.
4. Each sub-test repeats the cycle N times with freshly fuzzed/randomized data (configurable, default 5 iterations).
5. Assert the final object is identical to the original using `go-cmp` with any registered comparison options.

### Wiring at providers

Provider wires the library once in a test file, supplying the upjet provider object and a runtime scheme with types registered:
The test runner discovers all types and API versions registered in the scheme automatically — no per-resource configuration is needed.
Optionally, some test behavior can be customized. See `Customization` section.
```go
func TestRoundTrip(t *testing.T) {
	// 
    provider, _ := config.GetProvider(t.Context(), schema, false)
	providerNamespaced := config.GetProviderNamespaced(t.Context(), schema, false)

    testScheme := runtime.NewScheme()
    clusterapis.AddToScheme(testScheme)
	namespacedapis.AddToScheme(testScheme)

    rt, _ := roundtrip.NewRoundTripTest(provider, providerNamespaced, testScheme
	// ... customization options 
	)
    t.Run("Serialization", rt.TestSerializationRoundtrip)
    t.Run("Conversion",    rt.TestConversionRoundtrip)
}
```

### Customisation

When provider-specific quirks require it, the library exposes options to tune behavior without modifying shared code:

- **API Filtering** — include or exclude specific API groups or GroupKinds from testing.
- **Fuzzer tuning** — configure nil probability, slice sizes, iteration count, or register custom fill functions for fields that must match a specific value domain (e.g. enum strings).
- **Multiple fuzzer passes** — register several fuzzer configurations that run in sequence, e.g. one dense pass and one sparse pass, to improve coverage.
- **Comparison options** — plug in `go-cmp` options to handle known legitimate asymmetries in conversions

### Changes

- New package `pkg/apitesting/roundtrip` with three files:
  - `roundtrip.go` — core test runner, options, and parallelism-safe sub-test structure
  - `fuzzer.go` — Fuzzer helpers
  - `cmp.go` — k8s object comparison helpers
- Documentation added under `docs/api-roundtrip-testing.md`.

I have:

### DIFF
--- a/docs/api-roundtrip-testing.md
+++ b/docs/api-roundtrip-testing.md
@@ -1,0 +1,200 @@
+<!--
+SPDX-FileCopyrightText: 2026 The Crossplane Authors <https://crossplane.io>
+
+SPDX-License-Identifier: CC-BY-4.0
+-->
+
+# API roundtrip testing
+
+A testing utility library for verifying that Crossplane provider managed resources
+correctly survive two kinds of round trips:
+
+| Test | What it checks |
+|---|---|
+| **Serialization** | Every registered type can be JSON-encoded and decoded back to an identical object. |
+| **Conversion** | Every multi-version type survives spoke→hub→spoke and hub→spoke→hub conversion with no data loss. |
+
+The library builds on `k8s.io/apimachinery`'s fuzzing/round-trip infrastructure and
+`sigs.k8s.io/randfill` to generate random objects, then exercises the conversion
+functions registered by `pkg/controller/conversion.RegisterConversions`.
+
+---
+
+## How it works
+
+### Serialization round-trip
+
+`TestSerializationRoundtrip` delegates to the upstream
+`roundtrip.RoundTripExternalTypesWithoutProtobuf` helper. For every type in the
+scheme it:
+
+1. Creates a randomly filled instance using the fuzzer.
+2. Marshals it to JSON.
+3. Unmarshals back to a new instance.
+4. Asserts the two instances are equal.
+
+### Conversion round-trip
+
+`TestConversionRoundtrip` discovers every API group in the scheme, finds the hub
+version and all spoke versions for each `GroupKind`, then runs two sub-tests per
+(hub, spoke) pair:
+
+- **spoke → hub → spoke**: fill a spoke, convert to hub, convert back to a new spoke, compare.
+- **hub → spoke → hub**: fill a hub, convert to spoke, convert back to a new hub, compare.
+
+Each sub-test repeats the cycle `defaultFuzzIterations` (5) times with fresh random data.
+Sub-tests are run in parallel per group to keep the suite fast.
+
+---
+
+## Usage
+
+### Minimal setup
+
+```go
+// e2e/roundtrip/roundtrip_test.go
+package roundtrip
+
+import (
+    "testing"
+
+    "github.com/crossplane/upjet/v2/pkg/apitesting/roundtrip"
+    "k8s.io/apimachinery/pkg/runtime"
+
+    clusterapis    "github.com/upbound/provider-foo/apis/cluster"
+    namespacedapis "github.com/upbound/provider-foo/apis/namespaced"
+    "github.com/upbound/provider-foo/config"
+    "github.com/upbound/provider-foo/xpprovider"
+)
+
+func TestRoundTrip(t *testing.T) {
+    schema, err := xpprovider.GetProviderSchema(t.Context())
+    if err != nil {
+        t.Fatalf("GetProviderSchema: %s", err)
+    }
+
+    provider, err := config.GetProvider(t.Context(), schema, false)
+    if err != nil {
+        t.Fatalf("GetProvider: %s", err)
+    }
+
+    providerNamespaced, err := config.GetNamespacedProvider(t.Context(), schema, false)
+    if err != nil {
+        t.Fatalf("GetNamespacedProvider: %s", err)
+    }
+
+    testScheme := runtime.NewScheme()
+    if err := clusterapis.AddToScheme(testScheme); err != nil {
+        t.Fatalf("cluster-scoped apis AddToScheme: %s", err)
+    }
+    if err := namespacedapis.AddToScheme(testScheme); err != nil {
+        t.Fatalf("namespaced apis AddToScheme: %s", err)
+    }
+
+    rt, err := roundtrip.NewRoundTripTest(provider, providerNamespaced, testScheme)
+    if err != nil {
+        t.Fatalf("NewRoundTripTest: %s", err)
+    }
+
+    t.Run("TestSerializationRoundtrip", rt.TestSerializationRoundtrip)
+    t.Run("TestConversionRoundtrip",    rt.TestConversionRoundtrip)
+}
+```
+
+Pass `nil` for `providerNamespaced` if the provider only exposes cluster-scoped resources.
+
+---
+
+## Options
+
+All options are passed to `NewRoundTripTest` as variadic `TestOption` arguments.
+
+### Codec
+
+| Option | Description |
+|---|---|
+| `WithCodecFactory(c)` | Override the codec factory derived from the scheme. |
+
+### Filtering which kinds to test
+
+| Option | Description |
+|---|---|
+| `WithIncludeGroups(groups...)` | Only test these API groups. |
+| `WithIncludeGroupKinds(gks...)` | Only test these GroupKinds. |
+| `WithExcludeGroups(groups...)` | Skip these API groups. |
+| `WithExcludeGroupKinds(gks...)` | Skip these GroupKinds. |
+
+When no include filter is set, all groups registered in the scheme are tested
+(minus `defaultIgnoredKinds` and the empty/core group).
+
+### Fuzzer configuration
+
+`WithFuzzerConfig(opts ...FuzzerOption)` adds a fuzzer configuration.  Each
+registered configuration is run in sequence for every (kind, version-pair),
+accumulating coverage across different fuzz parameters.  When no
+`WithFuzzerConfig` call is made a single default configuration is used
+(NilChance≈0.2, NumElements 0–3, 10 iterations).
+
+Multiple calls each add a **distinct** configuration:
+
+```go
+rt, _ := roundtrip.NewRoundTripTest(provider, nil,
+    roundtrip.WithScheme(testScheme),
+    // First pass: no nil pointers, 20 iterations
+    roundtrip.WithFuzzerConfig(
+        roundtrip.FuzzerNilChance(0),
+        roundtrip.FuzzerIterations(20),
+    ),
+    // Second pass: sparse data, 5 iterations
+    roundtrip.WithFuzzerConfig(
+        roundtrip.FuzzerNilChance(0.8),
+        roundtrip.FuzzerNumElements(0, 1),
+        roundtrip.FuzzerIterations(5),
+    ),
+)
+```
+
+Available `FuzzerOption` constructors:
+
+| Constructor | Description |
+|---|---|
+| `FuzzerIterations(n)` | Number of fuzz-fill + round-trip cycles for this config. Default: 10. |
+| `FuzzerNilChance(p)` | Probability [0,1] that pointer fields are left nil. Default: ~0.2. |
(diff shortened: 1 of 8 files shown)


pkg/apitesting/roundtrip/roundtrip.go
```diff
@@ -0,0 +1,718 @@
...
+	}
+	return rt, nil
+}
+
+// setupTestInfrastructure performs expensive one-time setup shared across all
+// sub-tests: it rebuilds the codec factory from the (potentially caller-supplied)
+// scheme and registers hub↔spoke conversions for both cluster-scoped and
+// namespaced providers.
+func (rt *RoundTripTest) setupTestInfrastructure() error {
+	rt.codecFactory = serializer.NewCodecFactory(rt.scheme)
```
sergenyalcin: If I understand correctly, we always (unconditionally) re-assign the `rt.codecFactory` here. Is the following scenario possible:

- User set the `WithCodecFactory` `TestOption` when creating the new `RoundTripTest` object.
- Line 345 applied this option.
- Line 354 called `setupTestInfrastructure` and it re-assigns the `rt.codecFactory`.

Is it a valid scenario? If yes, the `WithCodecFactory` `TestOption` may become inoperable.

docs/api-roundtrip-testing.md
sergenyalcin, line 1: This doc points to the following references, but I couldn't find them in the codebase. Btw, I am assuming the roundtrip library is the one added here, not an external one.
  - roundtrip.EquateNilPtr()                                                                                                                                  
  - roundtrip.NoNilElementsInPointerSlices(...)
  - roundtrip.WithScheme(...)

pkg/apitesting/roundtrip/roundtrip.go
sergenyalcin, line 1: nit: As a future improvement, we may consider adding unit tests for this library. It will also help to increase the codecov report coverage.

pkg/apitesting/roundtrip/roundtrip.go
```diff
@@ -0,0 +1,718 @@
...
+}
+
+// TestSerializationRoundtrip verifies that every type registered in the scheme
+// and passing the include/exclude filters survives a JSON encode→decode cycle
+// with no data loss.  It delegates to the upstream k8s roundtrip helper.
+// The same WithIncludeGroups/WithExcludeGroups/WithIncludeGroupKinds/
+// WithExcludeGroupKinds filters that apply to TestConversionRoundtrip also
+// apply here.  The first fuzzer configuration is used to build the fuzzer.
+func (rt *RoundTripTest) TestSerializationRoundtrip(t *testing.T) {
+	objFuzzer := rt.getFuzzer(rt.fuzzerConfigs[0])
```
sergenyalcin: Question: Is this intentional? If a user registers two fuzzer configurations with `WithFuzzerConfig`, conversion tests run both, but serialization tests only run the first. During reviews, I am a bit confused here, so we may at least consider adding more detailed code comments for future readers. What do you think?

pkg/apitesting/roundtrip/roundtrip.go
```diff
@@ -0,0 +1,718 @@
...
+}
+
+// setupTestInfrastructure performs expensive one-time setup shared across all
+// sub-tests: it rebuilds the codec factory from the (potentially caller-supplied)
+// scheme and registers hub↔spoke conversions for both cluster-scoped and
+// namespaced providers.
+func (rt *RoundTripTest) setupTestInfrastructure() error {
+	rt.codecFactory = serializer.NewCodecFactory(rt.scheme)
+	if err := ujconversion.RegisterConversions(rt.providerCluster, rt.providerNamespaced, rt.scheme); err != nil {
+		return fmt.Errorf("failed to register conversions: %w", err)
```
sergenyalcin: nit: We may consider using the `errors` package also here.

pkg/apitesting/roundtrip/roundtrip.go
```diff
@@ -0,0 +1,718 @@
...
+// config — a temporary workaround for singleton-list and pointer-slice
+// handling in upjet conversions — and the appropriate scope function
+// (namespaced or cluster-scoped) for ObjectMeta.
+//
+// Callers must not share the returned slice across goroutines: randfill.Filler
+// is not goroutine-safe.  Call getFillers once per kind sub-test.
+func (rt *RoundTripTest) getFillers(namespaced bool) []fillerWithIterations {
+	fillers := make([]fillerWithIterations, 0, len(rt.fuzzerConfigs))
+	for _, cfg := range rt.fuzzerConfigs {
+		f := rt.getFuzzer(cfg).NumElements(0, 1).NilChance(0)
```
sergenyalcin: If I can follow the code correctly, `getFuzzer` applies the user's `FuzzerNilChance(p)` and `FuzzerNumElements(min, max)` configurations. But then `getFillers` overrides with hardcoded (0,1) and 0. A user calling these configurations would observe a test that silently ignores their settings. I am not sure this is a valid scenariot but if yes, at least we may consider documenting it.

pkg/apitesting/roundtrip/roundtrip.go
```diff
@@ -0,0 +1,718 @@
...
+		if group == "" {
+			// skip K8s core API group
+			continue
+		}
+		namespaced := rt.providerNamespaced != nil && rt.providerNamespaced.RootGroup != "" &&
+			strings.HasSuffix(group, rt.providerNamespaced.RootGroup)
+
+		t.Run(group, func(t *testing.T) {
+			t.Parallel()
+			for _, gk := range kinds.UnsortedList() {
```
sergenyalcin, line 531: Question: Do we need sorting the list here?

sergenyalcin (commented):
Thank you very much, @erhancagirici, for this library. This is very helpful. It will increase the confidence level in the generated APIs. I've left a few comments to discuss and consider.

pkg/apitesting/roundtrip/roundtrip.go
erhancagirici: we could have sorted here, though I think it won't be much significant.
We are running the subtests in parallel and test logs are flushed at the subtest end. Therefore the test output order will be random, depending on test running time.

pkg/apitesting/roundtrip/roundtrip.go
erhancagirici: your observation is correct, this was a workaround for conversion roundtrip tests until we get a reliable solution. There are two challenges:
- During fuzzing, currently, we are unable to determine whether a given slice field corresponds to a singleton list or a regular list. When max element is >=2 , singleton lists can be filled with multiple elements, failing the conversions & tests.
- Other challenge is: Upjet generated runtime structs has pointer slices for primitive types, e.g you can see fields with type `[]*string`. In practice, this lets you have nil elements in slices, something like:
```go
type Foo struct {
   Bar []*string
}
// ...
myFoo := Foo{
  Bar: []string{"fizz", nil, "buzz"}
}
// or 
myFoo := Foo{
  Bar: []string{ nil } // slice is length 1 with a nil element
}
```
I am not sure this is fully intended or a valid case in real-world MR APIs.
Combining this with singleton lists, roundtripping a singleton-list of length 1 with a nil element, ends up with a zero-value element.

Anyways, removed the overrides at this point of code, and moved it to default values when not specified.
Users explicit configuration will not be overridden.
- Providers that have singleton lists, should not configure max element (documented this)
- For the "nil challenge" consumers should configure `EquateEmptyAndSingleZeroSlice()` custom comparison

pkg/apitesting/roundtrip/roundtrip.go
erhancagirici: done 👍

pkg/apitesting/roundtrip/roundtrip.go
erhancagirici: was partially intentional, but I think it makes sense to run this multiple times as well. Updated 👍

docs/api-roundtrip-testing.md
erhancagirici: those were leftovers after a refactoring, updated 👍

pkg/apitesting/roundtrip/roundtrip.go
erhancagirici: 👍 updated to skip CodecFactory construction when it is provided via options.

pkg/apitesting/roundtrip/roundtrip.go
```diff
@@ -0,0 +1,725 @@
...
+		if gvk.Version == runtime.APIVersionInternal {
+			continue
+		}
+		groupKinds, ok := allowed[gvk.Group]
+		if !ok || !groupKinds.Has(gvk.GroupKind()) {
+			result[gvk] = true
+		}
+	}
+	return result
+}
```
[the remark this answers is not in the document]
erhancagirici: this is intended, if no filter is supplied we want to test the whole scheme. dismiss this review.

pkg/apitesting/roundtrip/roundtrip.go
```diff
@@ -0,0 +1,725 @@
...
+	useCustomCodec bool
+	// fuzzerConfigs is the list of fuzzer configurations to run per
+	// (kind, version-pair).  Each configuration produces an independent
+	// randfill.Filler and runs its own iteration count.  At least one entry is
+	// always present after NewRoundTripTest; callers add more with
+	// WithFuzzerConfig.
+	fuzzerConfigs []fuzzerOptions
+	// extraFuzzFns are additional randfill-compatible fuzzer functions appended
+	// to every fuzzer built from fuzzerConfigs.  Add them with WithExtraFuzzFuncs.
+	extraFuzzFns []any
```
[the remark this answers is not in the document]
erhancagirici: done

docs/api-roundtrip-testing.md
```diff
@@ -0,0 +1,250 @@
...
+	StateCompleted State = "COMPLETED"
+)
+
+var validStates = []State{ StatePending, StateRunning, StateCompleted }
+
+type MyWorkload {
+	Name string
+	State State // enum-like field with restricted set of valid values
+	PendingJobCount int 
+}
```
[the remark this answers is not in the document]
erhancagirici: done

sergenyalcin (approved):
Thanks @erhancagirici LGTM!

PR #640, opened 2026-05-04
Check for previous null values in filtered diffs

stevendborrelli:
## Summary

`filteredDiffExists` in `pkg/controller/external_tfpluginfw.go` discarded any
diff where the planned value (`Value1`) was null, on the assumption that a null
planned value means "field was never specified." This is correct for fields that
were never set, but it is wrong for Optional non-Computed attributes that a
user explicitly removes: the prior state is non-null, the planned state is null,
and that transition is a real diff that must trigger an Update.

for example: 

1. User removes a nested attribute from their Claim
2. Composition patches null attribute to the MR's spec.forProvider
3. Upjet computes proposed state attribute = null (correct)
4. Upjet sends proposed state to TF provider's PlanResourceChange
5. TF provider returns planned state with attribute = null (correct)
6. Upjet diffs planned state vs prior state: Value1 (planned) = null, Value2 (prior) = non-null
7. filteredDiffExists filters out the diff because Value1 is null
8. Upjet concludes no diff exists — no Update is triggered
9. Block remains on the cluster

The result was that removing an optional nested attribute through a Crossplane
claim had no effect — Upjet concluded the resource was up-to-date and never
called the provider's Update RPC.

## Fix

Add a second keep-condition to the filter: retain a diff when the planned value
(`Value1`) is null **and** the prior value (`Value2`) is non-null. This is the
unambiguous signature of an explicit removal.

The existing behavior is preserved for all other cases:
- `Value1` nil (child attributes of a null parent object) → still filtered
- `Value1` unknown (computed fields) → still filtered  
- Both values null (field never specified) → still filtered

Fixes #

I have:

Added table-driven tests in `TestFilteredDiffExists` cover the core removal case,
the "never specified" case that must stay filtered, and the nested-object removal
scenario where child-attribute diffs have nil `Value1` but the parent-level diff
carries the null/non-null transition.

### DIFF
--- a/pkg/controller/external_tfpluginfw.go
+++ b/pkg/controller/external_tfpluginfw.go
@@ -357,10 +357,20 @@ func (c *TerraformPluginFrameworkConnector) getResourceConfigTerraformValue(ctx 
 // when only computed attributes or not-specified argument diffs
 // exist in the raw diff and no actual diff exists in the
 // parametrizable attributes.
+// Exception: a diff where the prior value (Value2) is non-null and the
+// planned value (Value1) is null represents an explicit removal of a
+// previously-set optional attribute and must not be filtered.
 func (n *terraformPluginFrameworkExternalClient) filteredDiffExists(rawDiff []tftypes.ValueDiff) bool {
 	filteredDiff := make([]tftypes.ValueDiff, 0)
 	for _, diff := range rawDiff {
+		// Keep diffs where the planned value is non-null and known.
 		if diff.Value1 != nil && diff.Value1.IsKnown() && !diff.Value1.IsNull() {
+			filteredDiff = append(filteredDiff, diff)
+			continue
+		}
+		// Keep diffs where the prior value was non-null and the planned value
+		// is null — this is an explicit removal of a previously-set attribute.
+		if diff.Value1 != nil && diff.Value1.IsNull() && diff.Value2 != nil && !diff.Value2.IsNull() {
 			filteredDiff = append(filteredDiff, diff)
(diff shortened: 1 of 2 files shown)


pkg/controller/external_tfpluginfw.go
```diff
@@ -357,16 +357,53 @@ func (c *TerraformPluginFrameworkConnector) getResourceConfigTerraformValue(ctx
...
 // exist in the raw diff and no actual diff exists in the
 // parametrizable attributes.
-func (n *terraformPluginFrameworkExternalClient) filteredDiffExists(rawDiff []tftypes.ValueDiff) bool {
+// Exceptions:
+//   - A diff where the prior value (Value2) is non-null and the
+//     planned value (Value1) is null, and the diff does not belong to
+//     an attribute which is computed-only or one of its ancestors is
+//     computed-only. Such a case is taken as a previously set value being
+//     unset and is thus, not filtered.
+func (n *terraformPluginFrameworkExternalClient) filteredDiffExists(ctx context.Context, rawDiff []tftypes.ValueDiff) bool {
```
erhancagirici, line 366: a general side-note on this:
this function was implemented before we had a ProposedState implementation https://github.com/crossplane/upjet/blob/ee84659adb232bbab031561af08add2c5eea651a/pkg/controller/external_tfpluginfw.go#L381

to suppress some spurious diffs. After ProposedState, It is very likely that this function is not needed anymore.

erhancagirici (approved):
thanks! LGTM 👍

PR #641, opened 2026-05-04
fix panic when null prior in proposedNewAttributes

stevendborrelli:

- Fixes a panic (`AttributeName("..."): missing value type`) in `proposedNewAttributes` ([pkg/controller/proposed_state.go:495](pkg/controller/proposed_state.go#L495)) triggered during `Observe` when a `SingleNestedAttribute`'s prior TF state value is a typed null
- Derives the null placeholder's type from the schema attribute definition rather than from the (empty) prior state map
- Adds a regression test covering the null-prior + non-null-config path

### Root Cause

`proposedNewAttributes` handles the case where `prior.IsNull()` by constructing null placeholder values for each attribute in the loop. The original code read the type from `priorMap[name]`:

```go
priorV = tftypes.NewValue(priorMap[name].Type(), nil)
```

However, prior.As(&priorMap) silently returns an error and leaves priorMap empty when prior is null. A Go map miss on an empty map returns a zero-value tftypes.Value, and .Type() on a zero-value returns nil. Passing nil to tftypes.NewValue causes the panic.

This affects any resource where:

An optional SingleNestedAttribute (e.g. an addon block) is absent from the TF provider's Read response, causing a typed null in TF state
The user then sets that attribute in their Crossplane claim
Upjet's Observe path calls proposedState to compute the planned diff — triggering the panic on every reconciliation loop
Because Go map iteration order is non-deterministic, a different child attribute name appears in the panic message each time (external_traffic_policy, service_annotations, node_selector, version, etc.).

Fixes #

I have:

Added `TestProposedNewAttributes_NullPriorNestedAttribute` in [proposed_state_test.go](pkg/controller/proposed_state_test.go) reproduces the exact scenario: top-level resource exists (non-null prior), optional nested attribute is a typed null in prior state, user sets the attribute in config

### DIFF
--- a/pkg/controller/proposed_state.go
+++ b/pkg/controller/proposed_state.go
@@ -490,7 +490,9 @@ func proposedNewAttributes(attrs map[string]rschema.Attribute, prior, config tft
 	for name, attr := range attrs {
 		var priorV tftypes.Value
 		if prior.IsNull() {
-			priorV = tftypes.NewValue(priorMap[name].Type(), nil)
+			// prior.As silently fails on null values, leaving priorMap empty.
+			// Derive the null value's type from the schema instead of priorMap.
+			priorV = tftypes.NewValue(attr.GetType().TerraformType(context.TODO()), nil)
 		} else {
 			priorV = priorMap[name]
 		}
--- a/pkg/controller/proposed_state_test.go
+++ b/pkg/controller/proposed_state_test.go
@@ -7,65 +7,95 @@ package controller
 import (
 	"testing"
 
+	"github.com/google/go-cmp/cmp"
 	rschema "github.com/hashicorp/terraform-plugin-framework/resource/schema"
 	"github.com/hashicorp/terraform-plugin-go/tftypes"
 )
 
-// TestProposedNewAttributes_NullPriorNestedAttribute is a regression test for a
-// panic that occurred when a SingleNestedAttribute's prior value was a typed
-// null (e.g. the TF provider returned null for an unset optional addon).
-//
-// proposedNewAttributes previously called priorMap[name].Type() where priorMap
-// was empty (prior.As silently fails on null), yielding a zero-value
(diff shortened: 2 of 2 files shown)


stevendborrelli:
@coderabbitai review

stevendborrelli:
@coderabbitai review

pkg/controller/proposed_state.go
```diff
@@ -490,7 +490,9 @@ func proposedNewAttributes(attrs map[string]rschema.Attribute, prior, config tft
 	for name, attr := range attrs {
 		var priorV tftypes.Value
 		if prior.IsNull() {
-			priorV = tftypes.NewValue(priorMap[name].Type(), nil)
+			// prior.As silently fails on null values, leaving priorMap empty.
+			// Derive the null value's type from the schema instead of priorMap.
```
ulucinar, line 494: I believe we had better also check for errors from `tftypes.Value.As` above. But the approach taken here for reading the attribute's Terraform type directly from the resource schema is more robust and we can implement the error check with a follow-up PR (it's a different concern than this fix's target).

pkg/controller/proposed_state_test.go
```diff
@@ -0,0 +1,71 @@
...
+			"replicas":                tftypes.NewValue(tftypes.Number, nil),
+		}),
+	})
+
+	// Must not panic.
+	result := proposedState(schema, prior, config)
+
+	if !result.IsKnown() || result.IsNull() {
+		t.Errorf("expected a known non-null result, got %s", result)
+	}
```
ulucinar, line 70: nit: We can implement a stricter check here, i.e., that asserts the proposed state actually matches the config.

ulucinar (approved):
Thank you @stevendborrelli for fixing this. I've left a couple of comments that we can address with a follow-up PR, no need to block this one.

PR #657, opened 2026-05-14
fix(examples): flatten SchemaTypeObject blocks in generated example manifests

### LINKED ISSUE #656: ApplyAPIConverters does not flatten NestingModeSingle (SchemaTypeObject) blocks in generated examples
## Summary

`ApplyAPIConverters` (added in #538) correctly flattens single-element arrays for `TypeList`/`TypeSet` with `MaxItems=1` fields in generated example manifests. However, it does not handle `SchemaTypeObject` fields (Terraform plugin framework's `NestingModeSingle` blocks), which also need flattening.

## Background

The HCL-to-JSON scraper always wraps Terraform blocks in arrays per HCL spec. For `NestingModeSingle` blocks, upjet maps these to `SchemaTypeObject` (ValueType 9) and generates CRD types as embedded objects (pointer-to-struct), not slices. But the generated example manifests retain the array syntax from the scraper, producing invalid YAML that fails with "unknown field" errors when applied.

## Why this doesn't affect AWS/GCP/Azure providers

Those providers use the older SDK v2 `TypeList + MaxItems=1` pattern. The `SingletonListEmbedder` + `ApplyAPIConverters` pipeline handles that correctly because:
1. `SingletonListEmbedder` registers conversion paths with `[*]` wildcards for `TypeList`/`TypeSet` fields
2. `conversion.Convert` uses `fieldpath.Pave` to traverse these paths and flatten arrays

For `SchemaTypeObject` fields, `SingletonListEmbedder` correctly skips them (they're already objects in the CRD). But the traverser does NOT add `[*]` wildcards to `SchemaTypeObject` paths, so `conversion.Convert` fails with "not an object" when trying to traverse the arrays that remain in the scraped examples.

## Who is affected

Any upjet-based provider whose Terraform provider uses the plugin framework `NestingModeSingle` pattern instead of the older `TypeList + MaxItems=1` pattern. The `grafana/crossplane-provider-grafana` provider is the first to hit this (grafana/crossplane-provider-grafana#549).

## Workaround

We added a post-processing step in our generator that uses upjet's schema traverser to collect CRD paths for `SchemaTypeObject` fields, then walks generated example YAML files and flattens single-element arrays at those paths (grafana/crossplane-provider-grafana#557).

## Expected behavior

`ApplyAPIConverters` (or a related utility) should also handle `SchemaTypeObject` fields by flattening single-element arrays in generated examples to plain objects, matching the CRD's pointer-to-struct type.

## Related

- #538 - Added `ApplyAPIConverters` for `TypeList`/`TypeSet` with `MaxItems=1`
- #593 - `NestingModeSingle` Computed flag
- #345 - Wrong types in generated examples

Duologic:
## Summary

- Adds a post-processing step to `ApplyAPIConverters` that flattens single-element arrays at `SchemaTypeObject` (NestingModeSingle) field paths in generated example YAML manifests
- Uses upjet's existing schema traverser to collect CRD paths for `SchemaTypeObject` fields per resource
- Adds tests for both the path collection and the flattening logic

## Background

The HCL-to-JSON scraper wraps all Terraform blocks in arrays per HCL spec. For `NestingModeSingle` blocks, upjet generates CRD types as embedded objects (pointer-to-struct), not slices. However, the generated example manifests retained the array syntax from the scraper, producing invalid YAML that fails with "unknown field" errors when applied.

`ApplyAPIConverters` (added in #538) only handles `TypeList`/`TypeSet` with `MaxItems=1` via `SingletonListEmbedder`. `SchemaTypeObject` fields were not covered because:
1. `SingletonListEmbedder.VisitResource` filters on `Type == TypeList || TypeSet`, skipping `SchemaTypeObject` (ValueType 9)
2. The traverser doesn't add `[*]` wildcards for `SchemaTypeObject` paths, so they can't be passed through `conversion.Convert` the same way

## Implementation

A separate lenient flattening function is used instead of reusing `conversion.Convert`, because:
- `Convert` errors on non-slice values (correct for runtime), but example manifests may already have objects at some paths
- `SchemaTypeObject` paths require **lexical** ordering (parents before children), opposite to `Convert`'s reverse-lexical ordering for `ToEmbeddedObject`, because without `[*]` wildcards child paths are unresolvable while the parent is still an array

## Who is affected

Any upjet-based provider whose Terraform provider uses the plugin framework `NestingModeSingle` pattern. The first provider to hit this is `grafana/crossplane-provider-grafana` (grafana/crossplane-provider-grafana#549).

Fixes #656

### DIFF
--- a/pkg/examples/conversion/example_conversions.go
+++ b/pkg/examples/conversion/example_conversions.go
@@ -11,8 +11,10 @@ import (
 	"log"
 	"os"
 	"path/filepath"
+	"sort"
 	"strings"
 
+	"github.com/crossplane/crossplane-runtime/v2/pkg/fieldpath"
 	"github.com/pkg/errors"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	k8sschema "k8s.io/apimachinery/pkg/runtime/schema"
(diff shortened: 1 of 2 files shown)


Duologic:
Superseded by #753, which preserves the array-shaped manifests through reference resolution and flattens only the copies serialized to YAML.

PR #665, opened 2026-06-02
Add Support for ReconciliationPolicy Configurations

ulucinar:

This PR introduces the `ReconciliationPolicy.v1alpha1` type that currently allows the configuration of the exponential failure rate limiter via the `ExponentialFailureRateLimiter` struct type. It also introduces an `EncapsulatingRateLimiter` which encapsulates a specified default rate limiter and allows overriding the rate limiter to be used with a specific managed resource (MR). If the rate limiter for a specific MR is not overridden, then the given default rate limiter is used. 

Providers can instantiate a new `reconciliationpolicy.ExponentialFailureRateLimiter`, which is an exponential failure rate limiter that allows overriding the rate limiter parameters for specific MRs, using `reconciliationpolicy.NewExponentialFailureRateLimiter` as follows:
```go
rl := reconciliationpolicy.NewExponentialFailureRateLimiter(time.Second, 60*time.Second)
```
Here, the default base delay is 1s and the default max delay is 60s.

When a provider uses the `reconciliationpolicy.ExponentialFailureRateLimiter`, it must also set up the `reconciliationpolicy.Finalizer` via the `reconciliationpolicy.NewFinalizer` that allows chaining a child finalizer. And the provider must chain the actual MR reconciler with a `reconciliationpolicy.Reconciler` using the `reconciliationpolicy.NewReconciler` function as follows:
```go
func Setup(mgr ctrl.Manager, o tjcontroller.Options) error {
...
	return ctrl.NewControllerManagedBy(mgr).
		Named(name).
		WithOptions(co).
		WithEventFilter(xpresource.DesiredStateChanged()).
		Watches(&v1beta1.AlternateContact{}, eventHandler).
		Complete(reconciliationpolicy.NewReconciler(
			ratelimiter.NewReconciler(name, r, o.GlobalRateLimiter),
			mgr,
			v1beta1.AlternateContact_GroupVersionKind,
			reconciliationpolicy.WithRateLimiter(rl),
			reconciliationpolicy.WithSource(clients.ReconciliationPolicy),
		))
```

**NOTE**: The `ReconciliationPolicy` API is currently an alpha API that might be altered or even completely removed in the future.

**Note**: If you update the exponential failure rate limiter parameters for an MR, the retry counters of the referencing MRs will be reset, their reconcilers will start backing-off with delays starting from the configured new base-delay with a cap at the configured max-delay. The MR will be reconciled at its next scheduled time (if the MR has been in an error state, probably with an increased back-off delay according to the old (before the update) rate limiter parameters from the configuration source, and if its reconciler encounters an error during the next reconciliation, the reconciler will back-off initially with the new base-delay, and if the error condition continues, will continue backing-off with exponentially growing delays, up to the new max-delay.
TLDR; changing the rate limiter parameters in the configuration source resets the back-off counters of the MRs using that configuration source. Also, syntactical changes in the base-delay & max-delay that correspond to the same durations, e.g., updating a `1s` to a `1000ms`, are *not* considered as changes, meaning that the retry counters (of the MRs using that configuration source) will *not* be reset.

**Note**: A restart of the provider will reset all retry counters and reconcilers will back-off starting from the initial base-delay.

I have:

~- [ ] Added `backport release-x.y` labels to auto-backport this PR if necessary.~

Tested against `crossplane-contrib/provider-upjet-aws` `ProviderConfig` API. Please see https://github.com/crossplane-contrib/provider-upjet-aws/pull/2096.


pkg/internal/ratelimiter/encapsulating.go
```diff
@@ -0,0 +1,131 @@
...
+	defer c.mu.RUnlock()
+
+	c.getRateLimiterFor(req).Forget(req)
+}
+
+func (c *EncapsulatingRateLimiter[K]) NumRequeues(req reconcile.Request) int {
+	c.mu.RLock()
+	defer c.mu.RUnlock()
+
+	return c.getRateLimiterFor(req).NumRequeues(req)
```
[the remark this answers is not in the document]
ulucinar: But it's a programming error if `Add` is called with a `nil` rate limiter. Also using the default rate limiter instead of a `nil` rate limiter is not the desired behavior.

pkg/reconciler/reconciliationpolicy/finalizer.go
```diff
@@ -0,0 +1,60 @@
...
+	if cf.targets.exponentialFailureRateLimiter != nil {
+		cf.targets.exponentialFailureRateLimiter.Remove(
+			reconcile.Request{
+				NamespacedName: types.NamespacedName{
+					Namespace: obj.GetNamespace(),
+					Name:      obj.GetName(),
+				},
+			})
+	}
+	return cf.Finalizer.RemoveFinalizer(ctx, obj)
```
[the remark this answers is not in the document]
ulucinar: This is a valid point. We had better preserve the retry state in the associated exponential failure rate limiter in case the inner finalizer remove fails. Otherwise we reset to the base delay during a retry (of the finalizer removal).

pkg/reconciler/reconciliationpolicy/reconciler.go
```diff
@@ -0,0 +1,133 @@
...
+	if r.source == nil || r.targets.exponentialFailureRateLimiter == nil {
+		return nil
+	}
+
+	mg := resource.MustCreateObject(r.gvk, r.manager.GetScheme()).(resource.Managed)
+	if err := r.manager.GetClient().Get(ctx, req.NamespacedName, mg); err != nil {
+		// There's no need to requeue if the request no longer exists.
+		// Otherwise, request will be requeued because an error is returned.
+		// log.Debug("Cannot get managed resource", "error", err)
+		return errors.Wrap(resource.IgnoreNotFound(err), errGetManaged)
```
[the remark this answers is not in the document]
ulucinar: See my response above. Again, a panicking call to `resource.MustCreateObject` in case the GVK isn't registered in the scheme or the type doesn't implement `resource.Managed` should be treated as a programming error. That would be a permanent wiring error in the provider implementation, not a runtime conditional error.

apis/configuration/v1alpha1/reconciliation_policy.go
```diff
@@ -0,0 +1,39 @@
...
+	// +kubebuilder:validation:XValidation:rule="duration(self) >= duration('1s')",message="baseDelay must be at least 1s"
+	BaseDelay *metav1.Duration `json:"baseDelay,omitempty"`
+}
+
+// ReconciliationPolicy
+//
+// +kubebuilder:object:generate=true
+type ReconciliationPolicy struct {
+	// +optional
+	ExponentialFailureRateLimiter *ExponentialFailureRateLimiter `json:"exponentialFailureRateLimiter"`
```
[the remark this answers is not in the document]
ulucinar: Done.

pkg/reconciler/reconciliationpolicy/reconciler.go
```diff
@@ -0,0 +1,146 @@
...
+func (r *Reconciler) setRateLimiter(ctx context.Context, req reconcile.Request) error {
+	if r.source == nil || r.targets.exponentialFailureRateLimiter == nil {
+		return nil
+	}
+
+	mg := resource.MustCreateObject(r.gvk, r.manager.GetScheme()).(resource.Managed)
+	if err := r.manager.GetClient().Get(ctx, req.NamespacedName, mg); err != nil {
+		// There's no need to requeue if the request no longer exists.
+		// Otherwise, request will be requeued because an error is returned.
+		// log.Debug("Cannot get managed resource", "error", err)
```
sergenyalcin: I think the debug log statement on this line is a leftover from the implementation.

pkg/reconciler/reconciliationpolicy/reconciler.go
```diff
@@ -0,0 +1,146 @@
...
+
+	for _, opt := range o {
+		opt(r)
+	}
+
+	return r
+}
+
+// Reconcile the given request subject to the reconciler's Configurations.
+func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reconcile.Result, error) {
```
sergenyalcin, line 81: nit: If I haven’t missed anything, we aren’t logging that we’re using an external rate limiter instead of the default rate limiter. We can assume the user/controller is aware of this, but someone monitoring the logs who sees a `requeue-after=4m32s` value where it should normally be ~60 seconds would have no clue as to why. In this context, such a log message could be useful during troubleshooting. If we’ve already added this log somewhere, we can ignore it.

pkg/reconciler/reconciliationpolicy/reconciler.go
sergenyalcin, line 1: Worth a comment and release note entry: the override registry is in-memory, so a controller/configuration restart resets everything (like backoff history). The user may not expect this. So I believe documenting this will help to users to understand the mechanism of the new policy API.

sergenyalcin (approved):
Thanks @ulucinar LGTM!

pkg/reconciler/reconciliationpolicy/reconciler.go
ulucinar: Done. Thank you for catching this 🙏

pkg/reconciler/reconciliationpolicy/reconciler.go
ulucinar: It's a little bit tricky I think. We could use debug logging but I'm afraid it will pollute the reconciliation logs. I think we may need to implement some metrics for the delays at some point in the future (controller-runtime or client-go do not provider any metrics on back-off delays probably due to cardinality concerns). But we could at least collect some aggregate metrics and/or we can add some debug logging if users ask for it. If we had some further log levels, I would log at lower levels.

pkg/reconciler/reconciliationpolicy/reconciler.go
ulucinar: Done. Added a note section in the PR description. We can also reuse it in the release notes.

PR #668, opened 2026-06-09
Make the Default Rate Limiter for the Event Handler Configurable

ulucinar:

This PR proposes the `handler.WithDefaultRateLimiter` `handler.Option` so that when initializing a new `handler.EventHandler` via `NewEventHandler`, one can now specify the default rate limiter to be used with the event handler when enqueuing reconciliation requests within the asynchronous callbacks:
```go
...
rl := reconciliationpolicy.NewExponentialFailureRateLimiter(time.Second, 60*time.Second)
eventHandler := handler.NewEventHandler(handler.WithDefaultRateLimiter(rl), handler.WithLogger(o.Logger.WithValues("gvk", v1beta2.Analyzer_GroupVersionKind)))
...
```

The need for this has surfaced when @jonasz-lasut made observations in https://github.com/crossplane-contrib/provider-upjet-azuread/pull/343.

This allows the event handler to use the same rate limiter with the associated controller.

This PR also addresses an issue where multiple reconciliation requests can be enqueued for the same failing async create/update/delete operation. The Terraform plugin SDKv2 & framework external clients now enqueue reconciliation requests in their async callback handlers when:
- The async operation has succeeded: This is needed so that the MR status conditions are properly updated immediately, without waiting for the next poll period expiration.
- The async operation has failed and there's no cached error to be returned back to the managed reconciler: When the async operation fails, the request needs to be enqueued with an exponential back-off. If there already exists a cached error to be returned to the managed reconciler, the callback handlers for the async create/update/delete operations no longer enqueue reconciliation requests because the managed reconciler will already enqueue a rate-limited request.

This PR also updates the signatures of the `controller.CallbackProvider` interface methods so that they now all accept a `requestReconcile` parameter:
```go
// CallbackProvider provides functions that can be called with the result of
// async operations.
type CallbackProvider interface {
	// Create signals completion of the async Create operation for
	// the given name. If requestReconcile is set, an immediate or rate-limited
	// reconcile request will be enqueued, depending on the error state.
	Create(name types.NamespacedName, requestReconcile bool) terraform.CallbackFn
	// Update signals completion of the async Update operation for
	// the given name. If requestReconcile is set, an immediate or rate-limited
	// reconcile request will be enqueued, depending on the error state.
	Update(name types.NamespacedName, requestReconcile bool) terraform.CallbackFn
	// Destroy signals completion of the async Destroy operation for
	// the given name. If requestReconcile is set, an immediate or rate-limited
	// reconcile request will be enqueued, depending on the error state.
	Destroy(name types.NamespacedName, requestReconcile bool) terraform.CallbackFn
}
```
This allows the caller to specify whether a rate-limited or immediate re-queue request is desired or not. 

**Note**: This PR introduces breaking changes in the `controller.CallbackProvider` interface. Unless a provider has implemented its own `CallbackProvider`, it should not be affected from these breaking changes though. We've already updated the existing `CallbackProvider` implementations that upjet ships. The rest of the existing behavior & APIs should remain backward-compatible with this PR.

**Note**: The behavior of the TF CLI-based external client has not changed with this PR. It always re-queues reconciliation requests, as before. Optimizations for the CLI-based external client are the topic of future PRs. Its behavior should NOT have  changed with this PR.

I have:

The event handler behavior has also been manually validated against crossplane-contrib/provider-upjet-azuread. Please also see https://github.com/crossplane-contrib/provider-upjet-aws/pull/2104.

### DIFF
--- a/pkg/controller/external_async_tfpluginsdk.go
+++ b/pkg/controller/external_async_tfpluginsdk.go
@@ -152,6 +152,7 @@ func (n *terraformPluginSDKAsyncExternal) Create(_ context.Context, mg xpresourc
 		var ph panicHandler
 		defer cancel()
 		defer func() { // Finishing operations
+			currentErr := n.opTracker.LastOperation.Error()
 			err := tferrors.NewAsyncCreateFailed(ph.err)
 			n.opTracker.LastOperation.SetError(err)
 			n.opTracker.logger.Debug("Async create ended.", "error", err, "tfID", n.opTracker.GetTfID())
@@ -161,7 +162,11 @@ func (n *terraformPluginSDKAsyncExternal) Create(_ context.Context, mg xpresourc
 				Namespace: mg.GetNamespace(),
 				Name:      mg.GetName(),
 			}
-			if cErr := n.callback.Create(name)(err, ctx); cErr != nil {
+			// we request an immediate reconcile upon success to set the status, or
+			// in case of failure (err != nil), if there's no cached error.
+			// If there already exists a cached error, managed reconciler
+			// will already requeue.
+			if cErr := n.callback.Create(name, err == nil || currentErr == nil)(err, ctx); cErr != nil {
 				n.opTracker.logger.Info("Async create callback failed", "error", cErr.Error())
 			}
 		}()
@@ -189,6 +194,7 @@ func (n *terraformPluginSDKAsyncExternal) Update(_ context.Context, mg xpresourc
 		var ph panicHandler
 		defer cancel()
 		defer func() { // Finishing operations
+			currentErr := n.opTracker.LastOperation.Error()
 			err := tferrors.NewAsyncUpdateFailed(ph.err)
 			n.opTracker.LastOperation.SetError(err)
 			n.opTracker.logger.Debug("Async update ended.", "error", err, "tfID", n.opTracker.GetTfID())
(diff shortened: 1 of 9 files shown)


pkg/controller/external_async_tfpluginsdk.go
```diff
@@ -198,7 +204,7 @@ func (n *terraformPluginSDKAsyncExternal) Update(_ context.Context, mg xpresourc
 				Namespace: mg.GetNamespace(),
 				Name:      mg.GetName(),
 			}
-			if cErr := n.callback.Update(name)(err, ctx); cErr != nil {
+			if cErr := n.callback.Update(name, err == nil || currentErr == nil)(err, ctx); cErr != nil {
```
sergenyalcin, line 207: Very nit: In `external_async_tfpluginsdk.go`, the `Create` and `Delete` closures both get code comment for the new `requestReconcile` logic. `Update` in the same file silently gets only the code change with no comment. All three operations in `external_async_tfpluginfw.go` have the comment. We may consider adding a comment also this.

sergenyalcin (approved):
Thanks @ulucinar LGTM!

jonasz-lasut (approved):
LGTM, I've tested it with AzureAD and two issues that were noticeable earlier have been removed: base delay is now correct and there's no multiple reconciles getting triggered at once.

pkg/controller/external_async_tfpluginsdk.go
ulucinar: Sorry, I rushed and forgot about this comment. I'm planning to address this in a follow-up PR.

ulucinar: Thank you @sergenyalcin for catching this. Please see: https://github.com/crossplane/upjet/pull/670/changes/6c7a1929273c453b0dacd82105ff7b70b88db941
Marking this as resolved and making sure those changes get merged...

PR #677, opened 2026-06-16
feat: add getImportIDFn

### LINKED ISSUE #676: Add optional GetImportIDFn to ExternalName for providers where import ID differs from state ID

`GetIDFn` in `config.ExternalName` serves dual purpose: its return value is used both as the `terraform import` ID argument and as `d.Id()` in the reconstructed tfstate for `terraform refresh`.

Some Terraform providers use different ID formats for import vs state, making it impossible to satisfy both consumers with a single function.

For instance, with `provider-mongodbatlas` we identified in [`#127`](https://github.com/crossplane-contrib/provider-mongodbatlas/issues/127) that the upstream Terraform provider uses a custom `conversion.EncodeStateID` that base64-encodes compound keys into `d.Id()`:

- State d.Id(): `cHJvamVjdF9pZA==:NjY2NjY2NjY2NjY2NjY2NjY2NjY2NjY2-cm9sZV9uYW1l:dGVzdHhwLXJ3` (base64-encoded key:value, joined by `-`).
- Import format: 666666666666666666666666-testxp-rw (plain values joined by `-`).

That is:

- If `GetIDFn` returns the plain import format, terraform import works ✅ but terraform refresh breaks ❌  (DecodeStateID gets garbage).
- If `GetIDFn` returns the base64 state format, terraform refresh works ✅ but terraform import breaks ❌ (import function can't parse base64).

Add an optional `GetImportIDFn` field to config.ExternalName, which optionally returns the ID to pass to `terraform import`.
When nil, `GetIDFn` is used (current behavior, backwards compatible).

fernandezcuesta:

Some Terraform providers store resource IDs in a different format than what their import function expects. Until now, upjet used a single function (`GetIDFn`) to produce the ID for both purposes, making it impossible to support these providers correctly.

This PR introduces an optional `GetImportIDFn` field to `ExternalName`. When set, upjet uses it to produce the ID passed to `terraform import`, while continuing to use `GetIDFn` for the state file used by `terraform refresh`. When unset, the existing behavior is preserved.

Fixes #676

I have:

- ~[ ] Added `backport release-x.y` labels to auto-backport this PR if necessary.~

Add unit tests.

Built [provider-mongodbatlas](https://github.com/crossplane-contrib/provider-mongodbatlas/pull/128) 1.2.0-rc.1 on top of these changes, now I can set a custom function for `GetImportIDFn` for those resources where the upstream TF provider works as explained above.


vrabbi:
this would be huge for us as well in many providers

jonasz-lasut (approved):
Thank you for the contribution @fernandezcuesta ! The change seems to be fully backwards-compatible and does not change the behavior without explicit `GetImportIDFn`, LGTM

PR #678, opened 2026-06-17
Decouple conversion webhook registration from leader election

sergenyalcin:

Conversion webhook registration was accidentally gated behind leader election through the `SetupGated`/gate mechanism. `SetupGated` stores a closure that only fires on the elected leader, so follower pods never called `ctrl.NewWebhookManagedBy` and their webhook servers returned 404 for every conversion request. With two replicas and `--leader-election`, some conversion webhook calls fail.

The fix separates webhook registration from reconciler setup:

- **`controller.go.tmpl`**: adds `SetupWebhookWithManager` — a standalone function that registers the conversion webhook for a single resource kind. Removes the `if o.StartWebhooks { ... }` block from `Setup`; webhook registration is no longer the reconciler's responsibility.

- **`setup.go.tmpl`**: adds the `SetupWebhookWithManager{{ .Group }}` aggregator, following the same pattern as `Setup{{ .Group }}` and `SetupGated{{ .Group }}`, so providers can register all webhooks for a group in a single call.

Providers call `SetupWebhookWithManager_<group>` once before `mgr.Start()` on every pod, independent of the gate and leader election. Reconciler setup remains behind the gate and runs only on the leader.

See the generated provider with these changes: https://github.com/crossplane-contrib/provider-upjet-gcp/pull/953 and https://github.com/crossplane-contrib/provider-upjet-aws/pull/2122

I have:

Tested locally by sending webhook calls to the webhook handlers.

### DIFF
--- a/pkg/pipeline/templates/controller.go.tmpl
+++ b/pkg/pipeline/templates/controller.go.tmpl
@@ -23,6 +23,15 @@ import (
 	{{ .Imports }}
 )
 
+// SetupWebhookWithManager registers the conversion webhook for {{ .CRD.Kind }}.
+func SetupWebhookWithManager(mgr ctrl.Manager) error {
+	if err := ctrl.NewWebhookManagedBy(mgr, &{{ .TypePackageAlias }}{{ .CRD.Kind }}{}).
+		Complete(); err != nil {
+		return errors.Wrap(err, "cannot register webhook for the kind {{ .TypePackageAlias }}{{ .CRD.Kind }}")
+	}
+	return nil
+}
+
 // SetupGated adds a controller that reconciles {{ .CRD.Kind }} managed resources.
 func SetupGated(mgr ctrl.Manager, o tjcontroller.Options) error {
 	o.Options.Gate.Register(func() {
@@ -120,15 +129,6 @@ func Setup(mgr ctrl.Manager, o tjcontroller.Options) error {
 	{{- end}}
 	if o.MetricOptions != nil {
 		opts = append(opts, managed.WithMetricRecorder(o.MetricOptions.MRMetrics))
-	}
-
-	// register webhooks for the kind {{ .TypePackageAlias }}{{ .CRD.Kind }}
-	// if they're enabled.
-	if o.StartWebhooks {
-		if err := ctrl.NewWebhookManagedBy(mgr, &{{ .TypePackageAlias }}{{ .CRD.Kind }}{}).
-			Complete(); err != nil {
-			return errors.Wrap(err, "cannot register webhook for the kind {{ .TypePackageAlias }}{{ .CRD.Kind }}")
-		}
 	}
 
 	if o.MetricOptions != nil && o.MetricOptions.MRStateMetrics != nil {
--- a/pkg/pipeline/templates/setup.go.tmpl
+++ b/pkg/pipeline/templates/setup.go.tmpl
@@ -39,3 +39,17 @@ func SetupGated{{ .Group }}(mgr ctrl.Manager, o controller.Options) error {
 	}
 	return nil
 }
+
+// SetupWebhookWithManager{{ .Group }} registers conversion webhooks for all resource kinds in the group.
+func SetupWebhookWithManager{{ .Group }}(mgr ctrl.Manager) error {
+	for _, setup := range []func(ctrl.Manager) error{
+		{{- range $alias := .Aliases }}
+		{{ $alias }}SetupWebhookWithManager,
+		{{- end }}
+	} {
+		if err := setup(mgr); err != nil {
+			return err
+		}
+	}
+	return nil
+}


pkg/pipeline/templates/controller.go.tmpl
```diff
@@ -23,6 +23,15 @@ import (
 	{{ .Imports }}
 )
 
+// SetupWebhookWithManager registers the conversion webhook for {{ .CRD.Kind }}.
+func SetupWebhookWithManager(mgr ctrl.Manager) error {
+	if err := ctrl.NewWebhookManagedBy(mgr, &{{ .TypePackageAlias }}{{ .CRD.Kind }}{}).
+		Complete(); err != nil {
+		return errors.Wrap(err, "cannot register webhook for the kind {{ .TypePackageAlias }}{{ .CRD.Kind }}")
```
ulucinar, line 30: In case the same (version, kind) exists in a different group:

```suggestion
		return errors.Wrapf(err, "cannot register webhook for the kind %s", {{ .TypePackageAlias }}{{ .CRD.Kind }}_GroupVersionKind.String())
```

ulucinar (approved):
Thank you @sergenyalcin for debugging this issue & fixing it, lgtm. Expressed some concerns about the breaking changes below, let's discuss.

pkg/pipeline/templates/controller.go.tmpl
```diff
@@ -122,15 +131,6 @@ func Setup(mgr ctrl.Manager, o tjcontroller.Options) error {
 		opts = append(opts, managed.WithMetricRecorder(o.MetricOptions.MRMetrics))
 	}
 
-	// register webhooks for the kind {{ .TypePackageAlias }}{{ .CRD.Kind }}
-	// if they're enabled.
-	if o.StartWebhooks {
```
ulucinar, line 127: One concern here is some providers out in the wild might still be relying on this condition (and using this template shipped with upjet). Maybe we need to document the change in the release notes.

pkg/pipeline/templates/controller.go.tmpl
```diff
@@ -23,6 +23,15 @@ import (
 	{{ .Imports }}
 )
 
+// SetupWebhookWithManager registers the conversion webhook for {{ .CRD.Kind }}.
+func SetupWebhookWithManager(mgr ctrl.Manager) error {
```
ulucinar, line 27: Please see the comment below. We need to document the newly generated webhook setup functions and ask the provider authors to call them. Not all providers will have this bug (not all of them will be using CRD-gating). What could happen is that although they set the `controller.Options.StartWebhooks` to `True`, their webhooks won't be started. The good news is that they will get a change to see the changes (especially the removal of the previously generated condition section in `Setup` function for the webhooks and the newly generated `SetupWebhookWithManager` function) when they run `make generate`, so in my opinion, not a huge deal. But we had better still explain the changes and instruct the provider authors how to call the newly generated `SetupWebhookWithManager_<API group>` function(s) in some release notes. What do you think?

erhancagirici: just to double-check: AFAIU, in "gated" setups, when there are MRAPs disabling a particular `Kind`, this will register its webhook unconditionally, although the CRD won't exist in cluster.

I think this will still work with `controller-runtime`, but we should test it to make sure

pkg/pipeline/templates/controller.go.tmpl
erhancagirici: nit: newly introduced `SetupWebhookWithManager(mgr ctrl.Manager)` can accept `o tjcontroller.Options`, keeping the logic intact

pkg/pipeline/templates/setup.go.tmpl
```diff
@@ -38,4 +38,18 @@ func SetupGated{{ .Group }}(mgr ctrl.Manager, o controller.Options) error {
 		}
 	}
 	return nil
+}
+
+// SetupWebhookWithManager{{ .Group }} registers conversion webhooks for all resource kinds in the group.
+func SetupWebhookWithManager{{ .Group }}(mgr ctrl.Manager) error {
+	for _, setup := range []func(ctrl.Manager) error{
+		{{- range $alias := .Aliases }}
+		{{ $alias }}SetupWebhookWithManager,
```
erhancagirici, line 47: nit: another alternative could be to add `{{ $alias }}SetupWebhookWithManager,` into `Setup_<apiGroup>` and `SetupGated_<apiGroup>` templates:
this would:
- keep `Setup_<apiGroup>` logically the same
- decouple Webhook registration for `SetupGated_<apiGroup>`, the actual setup is still gated, webhooks are registered without gate.
- no extra call to `SetupWebhookWithManager_<apiGroup>` required from provider developers

On the other hand, the current approach has the advantage of keeping webhook registration clean and explicit.

pkg/pipeline/templates/controller.go.tmpl
sergenyalcin: I'll add a detailed section about this change to the Release Notes.

pkg/pipeline/templates/controller.go.tmpl
sergenyalcin: I'll add an instruction to the Release Notes.

@erhancagirici, I've tested this and the webhook registration doesn't be an issue for the unregistered CRDs or controller-runtime.

pkg/pipeline/templates/setup.go.tmpl
sergenyalcin: Yes, this was an option, but I preferred it because of the clear separation.

PR #683, opened 2026-06-24
Add support for Terraform runtime conversions in CLI-based external client

ulucinar:

This PR adds support for Terraform runtime conversions in the Terraform CLI-based external client. @sergenyalcin is working on a Crossplane provider for f5xc based on https://github.com/volterraedge/terraform-provider-volterra and it's CLI-based. Currently, when we try to convert the singleton lists in that provider's APIs to embedded objects, the reconciliations fail at runtime.

I have:

Tested against a Crossplane provider based on https://github.com/volterraedge/terraform-provider-volterra.

### DIFF
--- a/pkg/controller/external.go
+++ b/pkg/controller/external.go
@@ -245,9 +245,6 @@ func (e *external) Observe(ctx context.Context, mg xpresource.Managed) (managed.
 	if err := json.JSParser.Unmarshal(res.State.GetAttributes(), &tfstate); err != nil {
 		return managed.ExternalObservation{}, errors.Wrap(err, "cannot unmarshal state attributes")
 	}
-	if err := tr.SetObservation(tfstate); err != nil {
-		return managed.ExternalObservation{}, errors.Wrap(err, "cannot set observation")
-	}
 
 	// NOTE(lsviben) although the annotations were supposed to be set and the
 	// managed resource updated during the Create step, we are checking and
@@ -276,9 +273,23 @@ func (e *external) Observe(ctx context.Context, mg xpresource.Managed) (managed.
 		return managed.ExternalObservation{}, errors.Wrap(err, "cannot get connection details")
 	}
 
+	// Apply Terraform conversions.
+	tfstate, err = e.config.ApplyTFConversions(tfstate, config.FromTerraform)
+	if err != nil {
+		return managed.ExternalObservation{}, errors.Wrap(err, "cannot apply terraform conversions")
+	}
+	// Set the MR state using any Terraform runtime conversions applied.
+	if err := tr.SetObservation(tfstate); err != nil {
+		return managed.ExternalObservation{}, errors.Wrap(err, "cannot set observation")
+	}
+
 	var lateInitedParams bool
 	if policyHasLateInit {
-		lateInitedParams, err = tr.LateInitialize(res.State.GetAttributes())
+		buff, err := json.JSParser.Marshal(tfstate)
+		if err != nil {
+			return managed.ExternalObservation{}, errors.Wrap(err, "cannot marshal state attributes")
+		}
+		lateInitedParams, err = tr.LateInitialize(buff)
 		if err != nil {
 			return managed.ExternalObservation{}, errors.Wrap(err, "cannot late initialize parameters")
 		}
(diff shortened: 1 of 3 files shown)


pkg/terraform/files.go
```diff
@@ -110,22 +110,28 @@ func NewFileProducer(ctx context.Context, client resource.SecretClient, dir stri
...
+	convertedParams, err := cfg.ApplyTFConversions(params, config.ToTerraform)
 	if err != nil {
 		return nil, errors.Wrap(err, "cannot apply terraform parameter conversions")
 	}
-	fp.parameters = params
+	if err = resource.GetSensitiveParameters(ctx, client, tr, convertedParams, tr.GetConnectionDetailsMapping()); err != nil {
+		return nil, errors.Wrap(err, "cannot get sensitive parameters")
+	}
+	fp.Config.ExternalName.SetIdentifierArgumentFn(params, meta.GetExternalName(tr))
+	fp.parameters = convertedParams
```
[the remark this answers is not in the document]
ulucinar: Done.

pkg/controller/external.go
```diff
@@ -245,7 +245,12 @@ func (e *external) Observe(ctx context.Context, mg xpresource.Managed) (managed.
 	if err := json.JSParser.Unmarshal(res.State.GetAttributes(), &tfstate); err != nil {
 		return managed.ExternalObservation{}, errors.Wrap(err, "cannot unmarshal state attributes")
 	}
-	if err := tr.SetObservation(tfstate); err != nil {
+	// Apply Terraform conversions.
+	tfConvertedState, err := e.config.ApplyTFConversions(tfstate, config.FromTerraform)
```
sergenyalcin: @ulucinar I think there might be a subtle issue with the `ApplyTFConversions` call in `Observe`. Let me explain this.

`ApplyTFConversions` mutates the supplied map in place. In the background, `conversion.Convert` operates on a `fieldpath.Paved` backed by the original map and updates values directly via `setValue()`, then returns the same map instance.

Because of that, after:

```go
tfConvertedState, err := e.config.ApplyTFConversions(tfstate, config.FromTerraform)
```

`tfstate` and `tfConvertedState` are actually the same object. Here, the variable names give the impression that `tfstate` still contains the state in Terraform format, whilst `tfConvertedState` contains the state in converted CRD format, and this might be assumed. However, at this point, both references point to the converted representation.

This may be relevant for the `GetConnectionDetails()` function. At the very least, we may need to consider this. The paths returned by `GetConnectionDetailsMapping()` are Terraform schema paths (e.g., `x[0].y`). `GetSensitiveAttributes()` matches these paths against the provided state via `ExpandWildcards()`. Once the state has been transformed, these Terraform paths may no longer be available in the expected format (e.g., list → embedded object). Consequently, the matching may fail and the sensitive field may be silently skipped.

The sdkv2 implementation already handles this by extracting connection details before applying TF conversions:

```go
connDetails, err = resource.GetConnectionDetails(stateValueMap, ...)
...
stateValueMap, err = n.config.ApplyTFConversions(...)
```

If this problem is real, I think we may consider following the same ordering here: performing connection detail extraction (and anything else that depends on Terraform-native state) before `ApplyTFConversions()`, then using the converted state for observation, late-init, etc.

I think we may also consider dropping the separate tfConvertedState variable and just reassigning tfstate, since the conversion mutates the underlying map anyway.

I just wanted to be sure that we are aware of this case and that it is considered during the PR. @ulucinar, if you've already thought this and I am missing something in this analysis, I think we are good to go.

sergenyalcin (commented):
Thanks @ulucinar for working this. I’ve left a comment for discussion.

pkg/controller/external.go
ulucinar: Hi @sergenyalcin,
Agreed, thanks for catching this. I think the concern is real given how `fieldpath.Paved` is implemented, they are effectively sharing the same underlying map. I was under the false impression that we are keeping the converted state and the original state separate, hence had not assigned the converted state back to the `tfstate` variable, and used the new `tfConvertedState` variable. We both need to pass the original state to the `resource.SetCriticalAnnotations` & `resource.GetConnectionDetails` (and this is also reflected in the current code, they receive the `tfstate` variable). The dangerous part is both of those functions also receive the `Terraformed` resource, `tr`, but looks like we don't access the MR status in either of those places, so it should be safe to move the `SetObservation` call. 
This is quite dangerous. For now, I've updated the Go doc of `Resource.ApplyTFConversions`.

sergenyalcin (approved):
Thanks @ulucinar LGTM!

PR #684, opened 2026-06-25
fix(types): omit XValidation for required fields with references

jonasz-lasut:
When a field is marked as required but also has a reference configured, the field becomes satisfiable via *Ref/*Selector mechanisms. Explicitly set f.Required=false in NewReferenceField so no XValidation rule is emitted for such fields.

Fixes issue observed in https://github.com/crossplane-contrib/provider-upjet-gcp/issues/952

I have:

- [X] Read and followed Upjet's [contribution process].
- [X] Run `make reviewable` to ensure this PR is ready for review.
~~- [ ] Added `backport release-x.y` labels to auto-backport this PR if necessary.~~

Added unit test that validates this behavior

### DIFF
--- a/pkg/types/field.go
+++ b/pkg/types/field.go
@@ -349,6 +349,7 @@ func NewReferenceField(g *Builder, cfg *config.Resource, r *resource, sch *schem
 
 	f.Comment.Reference = *ref
 	f.Schema.Optional = true
+	f.Required = false // referenced fields are satisfiable via *Ref/*Selector; do not force-require
 
 	return f, nil
 }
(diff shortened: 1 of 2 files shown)


pkg/types/field.go
```diff
@@ -349,6 +349,7 @@ func NewReferenceField(g *Builder, cfg *config.Resource, r *resource, sch *schem
 
 	f.Comment.Reference = *ref
 	f.Schema.Optional = true
+	f.Required = false // referenced fields are satisfiable via *Ref/*Selector; do not force-require
```
ulucinar, line 352: Have we observed any behavior changes in a provider consuming this change? We may consider in the future a solution that does not rely on TF schema changes but it's a broader topic.

ulucinar (approved):
Thank you @jonasz-lasut, lgtm.

pkg/types/field.go
jonasz-lasut: I did not  observe any behavior changes during E2E tests with the upjet-gcp resource tfrom the linked issue outside of the expected lack of admission failure

PR #685, opened 2026-06-25
Set InstanceState's RawConfig & RawPlan if Resource is not found

ulucinar:

This PR proposes to set `InstanceState`'s `RawConfig` & `RawPlan` in case the TF state has been cached previously with a zero-valued `RawConfig` or `RawPlan`, and the external resource is now not found. Currently, if there's a cached Terraform state in the tracker store but the external client's `Observe` finds out that the external resource does not (yet) exist, `RawConfig` and `RawPlan` may stay `nil`. And this may panic the underlying Terraform provider in case it tries to read an attribute from the `RawConfig` or `RawPlan`, such as the Terraform AWS provider's tagging interceptor.

We have not been able to reproduce this issue but a review of the external client implementation reveals this can potentially happen right after a successful `Create`. And we've received the following logs from an older version (`v1.21.0`) of the AWS provider, where the Terraform provider accesses the `RawPlan` in `verify.SetTagsDiff`:

```
E0612 09:33:22.750019       1 runtime.go:77] Observed a panic: value is not an object
goroutine 10118 [running]:
k8s.io/apimachinery/pkg/util/runtime.logPanic({0x1b6d6d00, 0x21fef880})
	k8s.io/apimachinery@v0.30.0/pkg/util/runtime/runtime.go:75 +0x85
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Reconcile.func1()
	sigs.k8s.io/controller-runtime@v0.18.2/pkg/internal/controller/controller.go:103 +0xb2
panic({0x1b6d6d00?, 0x21fef880?})
	runtime/panic.go:785 +0x132
github.com/hashicorp/go-cty/cty.Value.GetAttr({{{0x0?, 0x0?}}, {0x0?, 0x0?}}, {0x1ed33191, 0x4})
	github.com/hashicorp/go-cty@v1.4.1-0.20200723130312-85980079f637/cty/value_ops.go:695 +0x2fd
github.com/hashicorp/terraform-provider-aws/internal/verify.SetTagsDiff({0x2232a818, 0xc0072ecf60}, 0xc01ec8cb40, {0x1ed18740?, 0xc0048a9d00?})
	github.com/hashicorp/terraform-provider-aws@v0.0.0-00010101000000-000000000000/internal/verify/diff.go:38 +0x165
github.com/hashicorp/terraform-provider-aws/internal/provider.New.(*wrappedResource).CustomizeDiff.func6({0x2232a8c0?, 0xc00c503730?}, 0xc01ec8cb40, {0x1ed18740, 0xc0048a9d00})
	github.com/hashicorp/terraform-provider-aws@v0.0.0-00010101000000-000000000000/internal/provider/intercept.go:186 +0x63
github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema.schemaMap.Diff(0xc00179c720, {0x2232a8c0, 0xc00c503730}, 0xc006803380, 0xc001748690, 0xc001750678, {0x1ed18740, 0xc0048a9d00}, 0x0)
	github.com/hashicorp/terraform-plugin-sdk/v2@v2.35.0/helper/schema/schema.go:698 +0x4b4
github.com/crossplane/upjet/pkg/controller.(*terraformPluginSDKExternal).getResourceDataDiff(0xc009bc12c0, {0x2252f4c0, 0xc008f6a308}, {0x2232a8c0, 0xc00c503730}, 0xc006803380, 0x0)
	github.com/crossplane/upjet@v1.5.0/pkg/controller/external_tfpluginsdk.go:421 +0xb5
github.com/crossplane/upjet/pkg/controller.(*terraformPluginSDKExternal).Observe(0xc009bc12c0, {0x2232a8c0, 0xc00c503730}, {0x224f5b78, 0xc008f6a308})
	github.com/crossplane/upjet@v1.5.0/pkg/controller/external_tfpluginsdk.go:504 +0x851
github.com/crossplane/upjet/pkg/controller.(*terraformPluginSDKAsyncExternal).Observe(0xc017e66ce0, {0x2232a8c0, 0xc00c503730}, {0x224f5b78, 0xc008f6a308})
	github.com/crossplane/upjet@v1.5.0/pkg/controller/external_async_tfpluginsdk.go:127 +0xc5
github.com/crossplane/crossplane-runtime/pkg/reconciler/managed.(*Reconciler).Reconcile(0xc00d1a3b00, {0x2232a818, 0xc01e7fc6c0}, {{{0x0, 0x0}, {0xc01b9e97d0, 0x29}}})
	github.com/crossplane/crossplane-runtime@v1.17.0/pkg/reconciler/managed/reconciler.go:973 +0x211f
github.com/crossplane/crossplane-runtime/pkg/ratelimiter.(*Reconciler).Reconcile(0xc00d236a00, {0x2232a818, 0xc01e7fc6c0}, {{{0x0?, 0x5?}, {0xc01b9e97d0?, 0xc00c1c7d10?}}})
	github.com/crossplane/crossplane-runtime@v1.17.0/pkg/ratelimiter/reconciler.go:54 +0x151
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Reconcile(0x2235bad0?, {0x2232a818?, 0xc01e7fc6c0?}, {{{0x0?, 0xb?}, {0xc01b9e97d0?, 0x0?}}})
	sigs.k8s.io/controller-runtime@v0.18.2/pkg/internal/controller/controller.go:114 +0xa5
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).reconcileHandler(0xc010384630, {0x2232a850, 0xc001603900}, {0x1d6390a0, 0xc002e5c400})
	sigs.k8s.io/controller-runtime@v0.18.2/pkg/internal/controller/controller.go:311 +0x39c
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem(0xc010384630, {0x2232a850, 0xc001603900})
	sigs.k8s.io/controller-runtime@v0.18.2/pkg/internal/controller/controller.go:261 +0x19d
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Start.func2.2()
	sigs.k8s.io/controller-runtime@v0.18.2/pkg/internal/controller/controller.go:222 +0x73
created by sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Start.func2 in goroutine 2655
	sigs.k8s.io/controller-runtime@v0.18.2/pkg/internal/controller/controller.go:218 +0x46c
```

I have:

I've added a regression test that setups up the panic conditions.

### DIFF
--- a/pkg/controller/external_tfpluginsdk.go
+++ b/pkg/controller/external_tfpluginsdk.go
@@ -521,6 +521,18 @@ func (n *terraformPluginSDKExternal) Observe(ctx context.Context, mg xpresource.
 	} else if diffState != nil {
 		diffState.Attributes = nil
 		diffState.ID = ""
+		// We still need to populate a non-nil RawPlan & RawConfig in InstanceState
+		// as the Terraform provider being called with this state may assume that
+		// they are not nil. An example is Terraform AWS provider's
+		// transparent tagging interceptor, which calls
+		// ResourceDiff.GetRawPlan().GetAttr("tags"), which will panic on the
+		// zero cty.Value with "value is not an object".
+		if diffState.RawPlan.IsNull() {
+			diffState.RawPlan = n.rawConfig
+		}
+		if diffState.RawConfig.IsNull() {
+			diffState.RawConfig = n.rawConfig
+		}
 	}
 
 	n.instanceDiff = nil
--- a/pkg/controller/external_tfpluginsdk_test.go
+++ b/pkg/controller/external_tfpluginsdk_test.go
@@ -168,7 +168,7 @@ func TestTerraformPluginSDKConnect(t *testing.T) {
 	for name, tc := range cases {
 		t.Run(name, func(t *testing.T) {
 			c := NewTerraformPluginSDKConnector(nil, tc.args.setupFn, tc.args.cfg, tc.args.ots, WithTerraformPluginSDKLogger(logTest))
-			_, err := c.Connect(context.TODO(), &tc.args.obj)
+			_, err := c.Connect(t.Context(), &tc.args.obj)
 			if diff := cmp.Diff(tc.want.err, err, test.EquateErrors()); diff != "" {
 				t.Errorf("\n%s\nConnect(...): -want error, +got error:\n", diff)
 			}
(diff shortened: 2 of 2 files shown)


pkg/controller/external_tfpluginsdk.go
```diff
@@ -521,6 +521,14 @@ func (n *terraformPluginSDKExternal) Observe(ctx context.Context, mg xpresource.
 	} else if diffState != nil {
 		diffState.Attributes = nil
 		diffState.ID = ""
+		// We still need to populate a non-nil RawPlan & RawConfig in InstanceState
+		// as the Terraform provider being called with this state may assume that
+		// they are not nil. An example is Terraform AWS provider's
+		// transparent tagging interceptor, which calls
+		// ResourceDiff.GetRawPlan().GetAttr("tags"), which will panic on the
+		// zero cty.Value with "value is not an object".
+		diffState.RawPlan = n.rawConfig
```
sergenyalcin: Although this change may seem minor, I feel it has the potential to result in an important change in behavior. When we make changes to our relationships with TF internals, we usually introduce this as part of a testing/validation effort across providers. It would be advisable to assess whether such a need arises specifically in relation to this change.

Another thing I have regarding this change is that, if there really was an issue with such high potential to affect such a wide area, why did it not come to light earlier? I think there is a possibility that the problem you described in the description could have been apparent before. Could this perhaps point to a resource-specific issue? (As we have experienced before with certain resources that relied directly on this type of TF internal.)

Another point that occurred to me is that we’re using our existing `RawConfig` for `RawPlan`. Do you think this is safe? I don’t have a great context for these two types, so your explanation of how to use `RawPlan` and `RawConfig` will be very helpful.

If I’ve understood correctly, this change will affect all cases where the resource does not exist/not found (in Observe call), covering a rather broad scope. At this stage, we could make various assumptions regarding the scope of impact; however, as I mentioned at the beginning, do you also think that, as with previous changes to TF internals, providers should make testing and validation efforts? Do you think this change is on that scale?

sergenyalcin (commented):
Thanks @ulucinar, I have left a comment for discussion.

pkg/controller/external_tfpluginsdk.go
ulucinar: > Another thing I have regarding this change is that, if there really was an issue with such high potential to affect such a wide area, why did it not come to light earlier? I think there is a possibility that the problem you described in the description could have been apparent before. Could this perhaps point to a resource-specific issue? (As we have experienced before with certain resources that relied directly on this type of TF internal.)

It's not quite a wide case. What we are addressing here is rather an edge case in which:
- A create has previously succeeded and the op tracker cache has been populated with the TF state.
- The external resource got deleted and we have TF state in the op tracker cache with a zero-valued `RawPlan` or `RawConfig`.
- `Observe` is called again.
Yesterday, we received some further feedback from the user:
```
...
The panic fires on managed resources stuck in Terminating that still hold the
[finalizer.managedresource.crossplane.io](http://finalizer.managedresource.crossplane.io/) finalizer.
...
```
They are saying the panic occurs during a deletion, which increases my confidence in the described scenario.

> Although this change may seem minor, I feel it has the potential to result in an important change in behavior. When we make changes to our relationships with TF internals, we usually introduce this as part of a testing/validation effort across providers. It would be advisable to assess whether such a need arises specifically in relation to this change.

>  ... do you also think that, as with previous changes to TF internals, providers should make testing and validation efforts?

Yes, I agree that we should be careful when introducing these changes as, in general, we don't know how a Terraform resource implementation makes use of these changes. The reassuring aspects related to these changes are:
- They kick in when the resource does not exist, not in the update path.
- And they don't kick in if the op tracker cache is cold. The initial create path should not be affected.
- The edge case we are targeting is when the external resource previously existed but then somehow got deleted (out-of-band or in-band).

> Another point that occurred to me is that we’re using our existing RawConfig for RawPlan. Do you think this is safe? I don’t have a great context for these two types, so your explanation of how to use RawPlan and RawConfig will be very helpful.

I believe what we do here is more appropriate for `RawConfig`, putting the user config into the corresponding field. `RawPlan` is the proposed post-plan value (config+computed+unknown). If the resource does not exist, `n.rawConfig` is a minimal non-nil value we can use in place. But as we already discussed above, it depends on the Terraform resource implementation where to read an attribute from. For a non-existing resource, the plan is to provision the resource from scratch as dictated by the given resource config. In our specific case, the resource implementation panics because the tagging interceptor being used reads the `tags` attribute from the plan as a customize diff implementation. So, we need to set a non-zero `RawPlan` as the underlying interceptor implementation does not do a nil-check. 

I've also added `{RawPlan,RawConfig}.IsNull` checks to make sure we don't override non-zero values, in which case the panic should not occur.

FWIW, this issue is also related to what we do in external client's `Create`, as we've been discussing for a long time.

sergenyalcin (approved):
Thanks @ulucinar LGTM!

PR #694, opened 2026-07-05
feat(design): a plan service for upjet providers

haarchri:

upjet providers compute a Terraform diff on every reconcile and throw the detail away. This one-pager proposes exposing that machinery as a versioned gRPC `PlanService`: desired and live resource in, provider-computed diff out (action, field changes in CRD paths, replacement info, diagnostics).

The service is stateless and credential-free. Prior state comes from `status.atProvider` in the request, and the server never calls a cloud API, so it runs anywhere the provider image runs. It covers all three execution modes (Plugin SDKv2, Plugin Framework, and the Terraform CLI workspace architecture) and is served from a hidden `internal plan-server` subcommand that providers pick up from a template change and `make generate`.

Companion proposal in crossplane/cli (link follows) drives these plan servers to deliver `crossplane project simulate` and `crossplane resource simulate`.

Fixes #

I have:

### DIFF
--- a/design/one-pager-upjet-plan-service.md
+++ b/design/one-pager-upjet-plan-service.md
@@ -1,0 +1,373 @@
+# A Plan Service for Upjet Providers
+
+* Owner: Christopher Haar (@haarchri)
+* Reviewers: Upjet Maintainers, Upjet Community
+* Status: Draft
+
+## Background
+
+Crossplane users want the equivalent of `terraform plan`: a preview of what a
+change would do to their external resources before they apply it. A change
+might be an edited managed resource (MR), a Composition update that re-renders
+MRs, or a provider upgrade. In every case the question is the same: *if this
+spec is applied, what changes, and is any of it destructive?*
+
+Today there is no way to ask an upjet provider that question. The diff
+machinery exists, it runs on every reconcile, but it is only reachable by
+actually reconciling, which means actually applying. Tooling that wants a
+preview is left approximating: comparing `spec.forProvider` against
+`status.atProvider` field by field. That approximation cannot see fields that
+force replacement, values the provider computes, defaults the provider
+injects, or `CustomizeDiff` logic. It tells you *something* changed; it cannot
+tell you the change is one Terraform could only apply by destroying and
+recreating your database, and therefore one the provider will refuse to
+apply at all.
+
+The accurate answer already lives inside every upjet provider. Upjet embeds
+the upstream Terraform provider and calls its diff machinery in-process during
+every reconcile: SDKv2 resources compute an `InstanceDiff` through the
+resource schema's `Diff`, and Terraform Plugin Framework resources call
+`PlanResourceChange`. Resources on the older Terraform CLI architecture go
+further still: their `Observe` runs an actual `terraform plan` in a
+workspace on every reconcile, then reduces the output to a single up-to-date
+boolean. The field-level answer users want is computed today, and thrown
+away. The same code that decides what `Update` will do can answer what
+`Update` *would* do. It is just not exposed.
+
+I want to expose it: let a client hand an upjet provider a desired resource
+and its live state, and get back the provider's own diff (action, field
+changes, replacement info, diagnostics) without the provider touching the
+cloud or a cluster.
+
+### Prior Art: How the Crossplane CLI Drives Functions
+
+`crossplane composition render` solved a similar problem for compositions.
+Rather than requiring a control plane to see what a composition produces, the
+CLI runs function packages locally as containers and talks to them over a
+versioned gRPC protocol (`RunFunctionRequest` / `RunFunctionResponse`). The
+package *is* the runtime; the CLI orchestrates.
+
+Provider packages can work the same way for planning. A provider image
+already contains everything needed to compute a diff: the embedded Terraform
+provider, the resource schemas, and upjet's conversion machinery between CRD
+shape and Terraform shape. What is missing is an entrypoint that serves diffs
+over gRPC instead of reconciling, and a protocol to ask for them.
+
+A companion one-pager in crossplane/cli proposes the client side: a `crossplane
+project simulate` command that resolves a project's provider dependencies, runs
+each provider image as a local plan server, and drives this protocol. This
+document proposes the provider side in upjet.
+
+## Goals
+
+- Expose upjet's diff machinery as a gRPC plan service any client can drive.
+- Support every execution mode upjet has: the in-process Terraform Plugin
+  SDKv2 and Plugin Framework clients, and the Terraform CLI workspace
+  architecture that predates them, so a plan covers every resource a
+  provider serves.
+- Make the service stateless and credential-free. All state arrives in the
+  request; the server never calls a cloud API and never reads a cluster.
+- Report diffs in Crossplane terms, CRD field paths rather than Terraform
+  attribute paths, so clients can render them against the user's YAML.
+
+It is not a goal to expose plans as an in-cluster API (a `Plan` CRD), to
+refresh state from the cloud at plan time, or to implement diffing for
+native (non-upjet) providers, though the protocol is deliberately shaped so
+they can implement it themselves. See Alternatives.
+
+## Proposal
+
+Add a versioned plan protocol under `proto/plan/v1alpha1` and a
+provider-agnostic implementation under `pkg/plan`, mirroring how composition
+functions pair a protocol with a runtime. A provider serves the protocol from
+a `plan-server` subcommand that clients like the Crossplane CLI run locally.
+
+### The Protocol
+
+A `PlanService` with two RPCs:
+
+```protobuf
+service PlanService {
+  // Plan computes a diff between the desired resource and the live resource.
+  rpc Plan(PlanRequest) returns (PlanResponse);
+
+  // GetInfo returns metadata about this plan server, including which API
+  // groups it supports, so clients can route resources to the right server.
+  rpc GetInfo(GetInfoRequest) returns (GetInfoResponse);
+}
+
+message PlanRequest {
+  // The desired managed resource, as full JSON (apiVersion, kind, metadata,
+  // spec).
+  google.protobuf.Struct desired_resource = 1;
+
+  // The live resource, with status.atProvider populated. Empty for a
+  // resource that does not exist yet.
+  google.protobuf.Struct live_resource = 2;
+}
+
+message PlanResponse {
+  string action = 1;                  // no-op | create | update | replace | delete
+  repeated FieldChange changes = 2;
+  bool requires_replace = 3;
+  repeated string replace_fields = 4; // CRD paths, e.g. spec.forProvider.region
+  repeated Diagnostic diagnostics = 5;
+  string error = 6;
+  google.protobuf.Timestamp computed_at = 7;
+}
+
+message FieldChange {
+  string field = 1;          // CRD path, e.g. spec.forProvider.deletionWindowInDays
+  string current_value = 2;  // Empty on create or when unknown.
+  string desired_value = 3;  // Empty when known only after apply.
+  bool requires_replace = 4;
+  bool sensitive = 5;        // Values are redacted when true.
+}
+```
+
+The request carries everything the server needs. Desired state is the MR the
+user wants to apply. Prior state is reconstructed from the live resource's
+`status.atProvider`, the provider's own record of the external resource from
+its last observation. Nothing else is consulted: no cluster reads, no cloud
+reads. That one decision buys most of the properties I care about:
+
+* **Credential-free.** The server needs no cloud credentials, no
+  `ProviderConfig`, no kubeconfig. Anyone who can pull the provider image can
+  compute a plan.
+* **Read-only by construction.** There is no code path that could mutate an
+  external system, because there is no code path that reaches one. The safety
+  property is structural, not enforced by care.
+* **Runs anywhere.** A laptop, CI, a cluster sidecar, anywhere the image
+  runs and the caller can supply the two structs.
+
+The cost is freshness: the diff is computed against the state the provider
+observed at its last reconcile, not the cloud's state right now. That is the
+same trade `terraform plan -refresh=false` makes, and the right default here.
+Live MRs are freshly observed on every poll interval anyway, and the client
+knows how stale `status.atProvider` is. A refresh mode could be added to the
+protocol later without breaking it (see Alternatives).
+
+One value in the `action` enum needs Crossplane words. `replace` is
+Terraform's answer, destroy and recreate, but a Crossplane provider never
+destroys an external resource to update it. When a reconcile computes a diff
+that requires replacement, the provider refuses to apply it: the MR turns
+unsynced and stays that way until a human intervenes, by recreating the
+resource deliberately or reverting the change. A `replace` in a plan is
+therefore not a prediction of automatic recreation. It is a warning that the
+change cannot be applied as written, which is exactly the surprise a preview
+exists to catch, and why `requires_replace` and `replace_fields` are
+first-class in the response rather than derivable details.
+
+`GetInfo` exists so clients route by fact instead of by convention. A client
+with a fleet of plan servers asks each which API groups it serves
+(`s3.aws.upbound.io`, `kms.aws.upbound.io`, ...) and routes each resource to
+the right one. The obvious alternative. inspecting each provider package's
+CRDs to derive its groups and it assumes every
+client has package-inspection machinery and a package to inspect. Asking the running
+server is one RPC, doubles as the readiness check after startup, and
+`GetInfoResponse` can grow fields later, capability flags
+without breaking existing clients.
+
+### The Engine
+
+`pkg/plan` provides the implementation, built from pieces upjet already has:
+
+* A **`Server`** implements the gRPC service. It is constructed from the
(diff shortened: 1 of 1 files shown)


haarchri:
We demonstrated the end to end flow, the CLI driving provider plan servers against a live control plane, in the Crossplane community meeting: see the
[recording](https://www.youtube.com/watch?v=zPJQpiIAYyQ&t=1100s) from 18:20 to 39:15.

design/one-pager-upjet-plan-service.md
```diff
@@ -0,0 +1,373 @@
...
+  string action = 1;                  // no-op | create | update | replace | delete
+  repeated FieldChange changes = 2;
+  bool requires_replace = 3;
+  repeated string replace_fields = 4; // CRD paths, e.g. spec.forProvider.region
+  repeated Diagnostic diagnostics = 5;
+  string error = 6;
+  google.protobuf.Timestamp computed_at = 7;
+}
+
+message FieldChange {
```
sergenyalcin, line 119: One thing I wondered about is the choice of modelling `current_value` and `desired_value` as plain strings.

For diff rendering, it seems like this makes a few important states indistinguishable. For example an empty string, null, and values that are only known after apply (unknown). That could make it difficult for clients to present certain transitions accurately.

Would it be worth considering a typed representation here instead?

The main reason I bring this up is that changing these value types later would be a protocol-breaking change, so it feels like something that’s worth validating before the API settles.

design/one-pager-upjet-plan-service.md
```diff
@@ -0,0 +1,373 @@
...
+  // spec).
+  google.protobuf.Struct desired_resource = 1;
+
+  // The live resource, with status.atProvider populated. Empty for a
+  // resource that does not exist yet.
+  google.protobuf.Struct live_resource = 2;
+}
+
+message PlanResponse {
+  string action = 1;                  // no-op | create | update | replace | delete
```
sergenyalcin, line 110: One thing I wasn’t quite understand from the current design is how a caller is expected to express a delete.

`PlanRequest` currently carries a desired resource and a live resource, but I couldn’t find where “the desired state is absence” is represented.

I may be missing something, but would it be worth making that contract explicit in the protocol?

If the intended convention is something like an empty or nil `desired_resource`, documenting that explicitly would probably help avoid different client implementations making different assumptions.

design/one-pager-upjet-plan-service.md
```diff
@@ -0,0 +1,373 @@
...
+  // Plan computes a diff between the desired resource and the live resource.
+  rpc Plan(PlanRequest) returns (PlanResponse);
+
+  // GetInfo returns metadata about this plan server, including which API
+  // groups it supports, so clients can route resources to the right server.
+  rpc GetInfo(GetInfoRequest) returns (GetInfoResponse);
+}
+
+message PlanRequest {
+  // The desired managed resource, as full JSON (apiVersion, kind, metadata,
```
sergenyalcin, line 100: One question I had here is around multi-version CRDs.

The server resolves a GVK to the Terraform resource type, but I couldn’t find how schema version differences are expected to be handled when generating field paths.

For example, if a caller sends a `v1beta1` resource but the generated mappings correspond to `v1beta2`, which schema are `FieldChange.field` paths expected to follow?

There may already be a clear implementation strategy for this, but it feels like an important part of the protocol contract to spell out explicitly.

design/one-pager-upjet-plan-service.md
```diff
@@ -0,0 +1,373 @@
...
+  list indices, map keys, for both SDKv2 and Framework schemas), redacts
+  sensitive values, and marks computed values as known-after-apply. It also
+  filters changes the user did not author: fields injected by the provider's
+  configuration injector and Crossplane's own system tags (`crossplane-kind`,
+  `crossplane-name`, `crossplane-providerconfig`). Without that filter every
+  plan for a fresh resource reports noise the user cannot act on.
+
+The executor needs a `terraform.Setup` to get provider metadata (and, for
+Framework resources, a configured provider). Because the server never reaches
+the cloud, this setup is minimal: the provider's singleton metadata plus
```
sergenyalcin, line 224: The idea of a “minimal setup” makes sense to me, but I wasn’t completely sure what that contract includes.

Some providers may still require configuration such as a region, project ID or endpoint before they can initialize successfully, even if credentials themselves aren’t needed.

Would it be worth describing that setup contract a bit more explicitly, as well as how initialization failures are expected to surface back to clients? I am aware that this may be related to provider's itself but at least we may consider mentioning this and putting some examples.

design/one-pager-upjet-plan-service.md
```diff
@@ -0,0 +1,373 @@
...
+whatever configuration the schema requires (a region, say, taken from the
+resource itself). It deliberately does not run the provider's normal
+credential resolution.
+
+Sensitive *inputs* deserve a note. A desired spec can reference sensitive
```
sergenyalcin, line 229: One thing I wondered about is whether returning those fields as unknown could unintentionally give clients the impression that “nothing changes”, even though the reconciler may later resolve those values and produce a real diff.

Would it make sense for the protocol to surface that limitation explicitly, for example through a caveat that clients can present alongside the plan?

design/one-pager-upjet-plan-service.md
```diff
@@ -0,0 +1,373 @@
...
+  repeated Diagnostic diagnostics = 5;
+  string error = 6;
+  google.protobuf.Timestamp computed_at = 7;
+}
+
+message FieldChange {
+  string field = 1;          // CRD path, e.g. spec.forProvider.deletionWindowInDays
+  string current_value = 2;  // Empty on create or when unknown.
+  string desired_value = 3;  // Empty when known only after apply.
+  bool requires_replace = 4;
```
sergenyalcin, line 123: The part I was wondering about is the client-side.

Should the protocol also describe how clients are expected to present a replace action?

Otherwise it feels possible for different clients to render the same response with quite different UX, even though the underlying semantics are the same.

design/one-pager-upjet-plan-service.md
```diff
@@ -0,0 +1,373 @@
...
+  * **Framework resources:** build prior, config, and proposed
+    `tftypes.Value`s from the same inputs and call `PlanResourceChange` on an
+    in-process protov6 provider server, the identical call Terraform itself
+    makes to plan.
+  * **Terraform CLI resources:** the workspace architecture already has the
+    pieces. Upjet synthesizes `main.tf.json` from the desired parameters and
+    `terraform.tfstate` from the live observation, and runs `terraform plan`
+    on every `Observe` today. The executor reuses that synthesis, runs
+    `terraform plan -refresh=false` to a plan file, and reads it back with
+    `terraform show -json`. The JSON plan's `resource_changes` carry the
```
sergenyalcin, line 205: This is probably just a documentation detail, but I thought it was worth pointing out since this document will likely be used as implementation guidance.

Unless I’m mistaken, `terraform show -json` reads the current state unless a plan file is provided.

I think the command here is intended to be:

```bash
terraform show -json <planfile>
```

design/one-pager-upjet-plan-service.md
```diff
@@ -0,0 +1,373 @@
...
+* A **`Server`** implements the gRPC service. It is constructed from the
+  provider's scheme, its root API group, and its generated resource
+  configuration map, the same `map[string]*config.Resource` the provider's
+  controllers are built from. It resolves an incoming resource's GVK to its
+  Terraform resource type (handling both cluster-scoped and namespaced v2 API
+  groups), looks up the resource's configuration, and hands off to the
+  executor. Plan failures are reported in `PlanResponse.error` rather than as
+  gRPC errors, so one bad resource in a batch does not look like a broken
+  server.
+* An **`Executor`** computes the diff, choosing the path from the resource's
```
sergenyalcin, line 184: One question that came to mind, though, is whether layers above the Terraform diff (transforms, reference resolution, etc.) can still change the effective desired state before reconciliation.

If that’s possible, would it be useful for the protocol to give clients some indication that the reported plan may not fully match the eventual reconcile behavior? Maybe better is finding a solution that will make closer to the reconcile machinery during calculating diff.

sergenyalcin (commented):
Thanks, @haarchri, for this design. The core insight here is solid, and the user value is clear. I've left comments that I think should be considered before moving forward.

design/one-pager-upjet-plan-service.md
haarchri: replaced the plain strings with a typed `FieldValue`: a oneof of a concrete `google.protobuf.Value` (so explicit null and "" stay distinct), unknown for known-after-apply, and sensitive for redacted values.

An unset `FieldValue` means the field is absent on that side, which is how a create differs from an update to null. These are the same states Terraforms own diff machinery distinguishes (nulls, unknowns, and sensitive marks), so the converter maps onto them directly instead of inventing a string encoding. Sensitivity also moved from a per-change flag into the value itself, since Terraform marks each side independently

design/one-pager-upjet-plan-service.md
haarchri: I removed delete from the action enum and made `desired_resource` required. A diff of desired against live can only conclude no-op, create, update, or replace; a delete only exists when the caller already knows the desired state is absence, and at that point the provider has nothing to add.

design/one-pager-upjet-plan-service.md
haarchri: Inbound the server converts the whole object spoke -> hub with the generated `ConvertTo`/`ConvertFrom` (the same `conversion.RoundTrip` the webhook runs).

Outbound there is no object to convert, only paths, and those map back mechanically because upjet conversions are declared as data, so the converter re-inserts the [0]s (or renames) at the declared paths. The one undeclarable case, a hand-written custom converter ?!

design/one-pager-upjet-plan-service.md
haarchri: true provider-level can come from three places
- the resource itself (an AWS resource carries its region in spec.forProvider.region, the common case)
- the referenced ProviderConfig
- per-provider defaults 
 
added the ProviderConfig case, since two resources can reference two ProviderConfigs with different projects, this is per-request data, so `PlanRequest` gains an optional `provider_config` field. The caller fetches the referenced ProviderConfig with its own RBAC and includes it; the server reads non-credential fields only (a GCP project ID, an Azure subscription, a custom endpoint)

design/one-pager-upjet-plan-service.md
haarchri: `FieldValue` gained an unresolved state, distinct from known-after-apply, and the server reports every secret-referenced parameter it could not evaluate as a `FieldChange` with an unresolved desired value.

Clients render that as a caveat alongside the plan and count it separately from no-ops, so a plan can honestly say `no changes, except the fields I could not see`. If it turns out callers want fully evaluated plans, the request can gain an explicit way to supply resolved Secret values later... thats additive ?!

design/one-pager-upjet-plan-service.md
haarchri: The companion CLI one-pager defines one concrete rendering (the `[-/+]` marker, replaces counted on their own in the summary, a warning that the MR would wedge unsynced) that other clients can adopt as a reference, but only the meaning binds them.

design/one-pager-upjet-plan-service.md
haarchri: Fixed

design/one-pager-upjet-plan-service.md
haarchri: everything that runs without a cluster (parameter merging, Terraform conversions, external-name functions, the schema diff with CustomizeDiff) is the reconciles own code, so those layers cannot diverge. 

The cluster-dependent layers are closed upfront by the client, which is the party with cluster access: resolved references are already persisted in a live resources spec, so the client merges them into the desired resource before sending (like it merges external-name and status), and admission defaults come from a server-side dry-run. What remains unresolvable, a reference on a resource that doesnot exist yet or a Secret value (no RBAC), surfaces per field as unresolved the explicit indication you suggested, attached to the exact field rather than a blanket caveat.. or ?

design/one-pager-upjet-plan-service.md
```diff
@@ -0,0 +1,667 @@
...
+composed resource with no rendered counterpart is reported as a delete,
+without calling a plan server. This also keeps the contract simple:
+`desired_resource` is always required, and a request without one fails as
+`INVALID_ARGUMENT`.
+
+Provider discovery is deliberately kept out of the protocol. A client needs to
+know which provider serves a resource before it knows which plan server to
+start or call, and that information is already part of the provider package.
+
+### Provider Discovery
```
ulucinar, line 261:
```suggestion
### Provider API Groups Discovery
```

ulucinar, line 261: Let's also add a discussion for adding a new feature identifier (like `DiffServer`) to the provider's metadata, so that the Crossplane CLI can query whether the new feature is supported.

design/one-pager-upjet-plan-service.md
```diff
@@ -0,0 +1,667 @@
...
+against provider-family-aws, extracting group/kind information took about 180 ms
+for a 208-CRD service package and 3.9 s for the 2,045-CRD monolith. So the
+`package.yaml` processing isn't the performance bottleneck.  The main factor
+affecting performance will be on the network side when the images are captured.
+However, this is a step that must be carried out in any case.
+More importantly, `provider-family-aws-ec2:v2.0.0` carries `package.yaml` in a
+0.91 MB compressed layer while the full image is 242 MB. Across 30 AWS service
+packages, the package layers total roughly 10--27 MB; pulling the full images
+for discovery is about 7.2 GB because the ~236--237 MB provider binary layers
+do not deduplicate across service packages.
```
ulucinar, line 291: Thanks for including this discussion. We had better be careful about the network transfers we require as it looks like the network will be the bottleneck for large configurations with many provider dependencies.

design/one-pager-upjet-plan-service.md
```diff
@@ -0,0 +1,667 @@
+<!--
+SPDX-FileCopyrightText: 2026 The Crossplane Authors <https://crossplane.io>
+
+SPDX-License-Identifier: Apache-2.0
+-->
+
+# A Plan Service for Upjet Providers
```
ulucinar, line 7: We should stress that whatever APIs we introduce in this context will be unstable APIs , protobuf/gRPC/package metadata APIs/upjet APIs, etc.

design/one-pager-upjet-plan-service.md
```diff
@@ -0,0 +1,667 @@
...
+a `plan-server` subcommand that clients like the Crossplane CLI run locally.
+
+### The Protocol
+
+A `PlanService` with one RPC:
+
+```protobuf
+service PlanService {
+  // Plan computes a diff between the desired resource and the live resource.
+  rpc Plan(PlanRequest) returns (PlanResponse);
```
ulucinar, line 98: The `Plan` implementation had better, for now, reject a request that attempts to calculate diff between different API versions of an MR. 

In future iterations, we can look into extending the in-memory client with upjet's CRD API conversion registry so that the in-memory client becomes capable of carrying out these conversions.

design/one-pager-upjet-plan-service.md
```diff
@@ -0,0 +1,667 @@
...
+  repeated string replace_fields = 4; // CRD paths, e.g. spec.forProvider.region
+  repeated Diagnostic diagnostics = 5;
+  string error = 6;
+  google.protobuf.Timestamp computed_at = 7;
+}
+
+message FieldChange {
+  // CRD path in the schema of the request's apiVersion,
+  // e.g. spec.forProvider.deletionWindowInDays.
+  string field = 1;
```
ulucinar, line 131: `field` sound more like a field name, better to be explicit here:
```suggestion
  string field_path = 1;
```

design/one-pager-upjet-plan-service.md
ulucinar, line 261: We should also capture the fact that even if an (upjet-based) Crossplane provider has the `DiffServer` feature, it should not mean that the provider will support diff'ing for all of its resources.

Rationale: The Terraform resource's diff implementation could, in theory, depend on a properly populated SDK client/configuration meta...

In those cases, we should gracefully report the failure via the `PlanResponse` protobuf message. In these failure cases, the Crossplane CLI should be reverting back to a raw diff'ing mode (like raw YAML diffing between desired and actual MR manifests), just like it would do for a provider that does not have the `DiffServer` capability (and should be making it explicit to the user that the diff was calculated without the actual provider's support).

design/one-pager-upjet-plan-service.md
```diff
@@ -0,0 +1,667 @@
...
+provider-agnostic implementation under `pkg/plan`, mirroring how composition
+functions pair a protocol with a runtime. A provider serves the protocol from
+a `plan-server` subcommand that clients like the Crossplane CLI run locally.
+
+### The Protocol
+
+A `PlanService` with one RPC:
+
+```protobuf
+service PlanService {
```
ulucinar, line 96: We should install a panic handler for the provider gRPC server we running running.

design/one-pager-upjet-plan-service.md
sergenyalcin: Done!

sergenyalcin: Done!

PR #695, opened 2026-07-12
Fix resolving namespaced conversions to wrong provider in RoundTrip

sergenyalcin:

`registry.GetConversions` selects between `providerCluster` and `providerNamespaced` by checking `tr.GetNamespace()` on the passed Terraformed object. In RoundTrip, all three conversion stages (`PrioritizedManagedConversion`, `PavedConversion`, `ManagedConversion`) call `GetConversions(dst)`, but `dst` is a freshly constructed object with an empty `ObjectMeta` at the first point, its namespace has not yet been populated. As a result, `GetConversions` always fell through to `providerCluster`, causing namespaced resources to have their conversions resolved from the wrong provider.

The fix copies the namespace from `src` to `dst` before the first `GetConversions` call, mirroring the existing check for `src.GetNamespace() != ""` to keep cluster-scoped resources unaffected. This ensures all three conversion loops operate against the correct provider for namespaced resources.

We did not see this behavior before, because we don't have multi-versions CRDs and conversions for namespaced resources yet.

I have:

### DIFF
--- a/pkg/controller/conversion/functions.go
+++ b/pkg/controller/conversion/functions.go
@@ -43,6 +43,12 @@ func (r *registry) RoundTrip(dst, src resource.Terraformed) error { //nolint:goc
 			src.GetObjectKind().SetGroupVersionKind(gvk)
 		}
 	}
+	// Propagate namespace before any GetConversions call: registry.GetConversions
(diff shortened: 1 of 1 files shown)


jonasz-lasut (approved):
Thank you @sergenyalcin LGTM!

erhancagirici (approved):

PR #698, opened 2026-07-15
fw-external-client: fix TF unexpected identity change errors at Observe

erhancagirici:

Fixes spurious "unexpected identity change" errors raised by
TF provider during `Observe` for Plugin Framework resources that support
[resource identity](https://developer.hashicorp.com/terraform/plugin/framework/resources/identity)

### Root cause

During `Observe` calls, TF Plugin Framework resource identity is fed to TF `ReadResource` RPC (as `CurrentIdentity`) and TF `ReadResourceResponse` returns the `NewIdentity`
When the current identity does not match the returned NewIdentity, TF `ReadResource` RPC fails the read with an "unexpected identity change" diagnostics.

- Some code paths can produce incomplete/invalid identities, e.g. the external API Create operation failed for some reason, say due to a permission/config issue eventually got resolved.
- these incomplete/invalid identities are stored as the CurrentIdentity in the opTracker cache
- Subsequent reconciles use it
- Say at some point, the original API failure reason is resolved (e.g. external API become available or user fixed their MR spec/config)
- This results in subsequent `Observe` operations to fail due to a spurious "identity change", even if the newly returned identity is valid (i.e invalid prior identity -> valid  identity transition)
- The MR becomes stuck at `Observe` step

A provider restart fixes it by rehydrating the cache, but still inconvenient.
An example reported issue is https://github.com/crossplane-contrib/provider-upjet-aws/issues/2135

This only affects the following resources in following particular conditions:
- TF plugin framework resource using TF framework external client
- Resource supports TF Resource Identity
- Resource encounters an external API failure during create
- Produces an incomplete/stale identity as a result of failure
  - has an indeterministic provider-assigned identifier 

### Fix

Reset or do not store potentially invalid TF resource identities.

- During XP `Create` after `ApplyResource` RPC calls with fatal diagnostics: Treat the returned identity as invalid/partial. Still preserves the returned (partial) state, so that is still available for recovery, rehydration
- At `Observe` resource not found case: this is to cover the case for
 FW resources with `FrameworkResourceWithComputedIdentifier`, that uses a stub/placeholder identifier for the very-first `Observe` to trigger an initial not-found. This expectedly returns a nil state, but can return a non-empty FW Identity with placeholder value. We don't want to store this, otherwise when the resource is actually created, this will cause an identity change from `placeholder -> real` identity.
- At `ResetReconstructedState`: This now also clear the associated identity of a reconstructed state.
- At `FrameworkResourceWithComputedIdentifier`: This is a complementary change. Do not return the placeholder value as a valid external-name at `GetExternalNameFn`. Placeholder identifiers can only exist at "input states" of TF Observe, should never exist in resulting TF states (after Read or Apply ), which `GetExternalNameFn` is only called for in FW resources. Instead error out if found.

Addresses https://github.com/crossplane-contrib/provider-upjet-aws/issues/2135

I have:

- Added unit tests
- test branch at provider-aws https://github.com/crossplane-contrib/provider-upjet-aws/pull/2148
  - uptest all FW resources with Identity Support
- Manually reproduce https://github.com/crossplane-contrib/provider-upjet-aws/issues/2135 and tested recovery without a pod restart

### DIFF
--- a/pkg/controller/external_tfpluginfw.go
+++ b/pkg/controller/external_tfpluginfw.go
@@ -618,6 +618,20 @@ func (n *terraformPluginFrameworkExternalClient) Observe(ctx context.Context, mg
 		} else {
 			stateValueMap = conv.(map[string]any)
 		}
+	} else if n.supportsIdentity() {
+		// For FW resources that use stub/placeholder identifiers
+		// in their external name, the initial Observe returns a
+		// "not-found" with nil TF state as expected, however they might
+		// return a non-empty TF identity with the placeholder identifier.
+		// Discard the TF identity for non-existent resource observations
+		// otherwise subsequent reconciles detect a superfluous TF identity
+		// change, failing the Observe.
+		//
+		// This also clears the TF identity for other not-found cases
+		// like resource was fully established but was deleted out-of-band.
+		// This is fine as Upjet in fact does not rely on TF resource
+		// identities and rehydrates them in subsequent reconciles
+		n.opTracker.SetFrameworkIdentity(nil)
 	}
 
 	// TODO(cem): Consider skipping diff calculation to avoid potential config
@@ -742,9 +756,15 @@ func (n *terraformPluginFrameworkExternalClient) Create(ctx context.Context, mg 
 		// In the following reconciles, this helps to track the external
 		// resource, rather than try to recreate that might cause leaking.
 		n.opTracker.SetFrameworkTFState(applyResponse.NewState)
-		if n.supportsIdentity() {
-			n.opTracker.SetFrameworkIdentity(applyResponse.NewIdentity)
-		}
+		// note: do not store returned TF Framework Identity here.
+		// https://github.com/crossplane-contrib/provider-upjet-aws/issues/2135
+		// The returned TF identity might be partial/garbage (missing some fields).
+		// When subsequent observe recovers the resource via state, it detects an
+		// superfluous identity change garbage->valid and errors out.
+		//
+		// Instead, ignore the returned TF identity and let subsequent
+		// reconciliations rehydrate TF identity via state
+
 		return managed.ExternalCreation{}, errors.Wrap(fatalDiags, "resource creation call returned error diags")
 	}
 
--- a/pkg/config/externalname.go
+++ b/pkg/config/externalname.go
@@ -229,7 +229,7 @@ func FrameworkResourceWithComputedIdentifier(identifier, placeholder string) Ext
 		WithGetExternalNameFn(func(fn GetExternalNameFn, tfState map[string]any) (string, error) {
 			if id, ok := tfState[identifier]; ok {
 				idStr := fmt.Sprintf("%v", id)
-				if len(idStr) > 0 {
+				if len(idStr) > 0 && idStr != placeholder {
 					return idStr, nil
 				}
 			}
--- a/pkg/config/externalname_test.go
+++ b/pkg/config/externalname_test.go
@@ -277,6 +277,82 @@ func TestTemplatedGetIDFn(t *testing.T) {
 	}
 }
 
+func TestFrameworkResourceWithComputedIdentifierGetExternalNameFn(t *testing.T) {
+	const (
+		identifier  = "arn"
+		placeholder = "upjet-placeholder-id"
+	)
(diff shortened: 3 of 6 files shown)


pkg/controller/external_tfpluginfw.go
```diff
@@ -742,9 +756,15 @@ func (n *terraformPluginFrameworkExternalClient) Create(ctx context.Context, mg
 		// In the following reconciles, this helps to track the external
 		// resource, rather than try to recreate that might cause leaking.
 		n.opTracker.SetFrameworkTFState(applyResponse.NewState)
-		if n.supportsIdentity() {
-			n.opTracker.SetFrameworkIdentity(applyResponse.NewIdentity)
-		}
```
ulucinar, line 747: What if the failure observed during creation is not due to a misconfiguration but, for example, is due to a timeout or similar? Aren't we losing track of the external resource (in cases of generated identifiers) in those scenarios?

ulucinar (commented):
Hi @erhancagirici,
Thank you for working on a fix on unexpected identity change errors. I've a question for the changes in the create path. Could you please take a look?

pkg/controller/external_tfpluginfw.go
erhancagirici: Short answer: In theory possible, but in practice depends on the provider/resource implementation. And in the context of today's upjet, the gain/covered scenarios are (relatively) not much due to how TF identity is used. 

Quick primer:
In TF `ApplyResource` RPC, both the `NewState` and `NewIdentity` in `ApplyResourceResponse`, are first initialized to `PlannedState` and `PlannedIdentity` by default by the FW server, before hitting any resource-specific logic (see [source](https://github.com/hashicorp/terraform-plugin-framework/blob/1b5d79f3864a4d5f741d20b16593e4afec6f0939/internal/fwserver/server_createresource.go#L103-L124)) . 

Some breakdown of scenarios:
We continue to persist the partial state `applyResponse.NewState`. If set by the resource's Create implementation, it will be available for keeping track/recovering external name.

Saving the partial identity, would have an added benefit only when following cases all hold: 
- Resource's `Create` implementation returns a garbage partial state, but a valid partial identity, i.e. returned Identity has more info than the state.
- The resource has an undeterministic (provider-assigned, computed) identifier
  - For deterministic identifiers, external name configurations already handles this.

In practice, I observe that `Identity` is mostly derived from resulting `State`. For example: `terraform-provider-aws` implements a [generic interceptor](https://github.com/hashicorp/terraform-provider-aws/blob/214babcbc7ecbc19c5d75e0215c42c540d64703c/internal/provider/framework/identity_interceptor.go#L51) that runs after CRUD RPC calls, to populate Identity from resulting state. [Here](https://github.com/hashicorp/terraform-provider-aws/blob/214babcbc7ecbc19c5d75e0215c42c540d64703c/internal/provider/framework/identity_interceptor.go#L88), it is the `OnError` path of the interceptor, with the same logic.
IIUC this means, saving the partial identity, would never have additional benefits for AWS provider.

Of course, that does not mean this case is invalid but I'd consider them to be comparatively rare and very resource dependent. TF Identity is still not mandatory for TF resources to implement. And if the resource's developer was handled the partial identity cases for undeterministic IDs, I'd expect them to reflect that information to State as well, or not handle both at all.  Don't think it is worth the complexity here, open to ideas here though. 

Another point is how TF Identity is used today in Upjet. Upjet mostly rely on external-name machinery for keeping track of resource identities, and the TF identity impl. mostly targets to satisfy the TF protocol. `TF Identities` are somewhat ephemeral in the current implementation. Because, unlike the State, they are not "persisted" into the MR and `opTracker` cache-only. A provider restart clears all TF Identity-related data and it is rehydrated via next Observe's result verbatim. This means a restart can invalidate its core functionality.

In summary, we'll basically be at the "pre-identity" behavior.

|Partial State | Partial Identity | Notes |
|--|--|--|
| Garbage | Garbage | No way to reliably recover anyway, same as today's behavior |
| Valid       | Garbage | Recover via state, same as today   | 
| Valid       | Valid        | Recover via state , same as today |
| Garbage | Valid        |  The case explained above. |

ulucinar (approved):
Thank you @erhancagirici, lgtm.

PR #711, opened 2026-07-24
Switch to dynamic storage version migration

sergenyalcin:

Follow up #588. Switches to dynamic approach instead of static configuraiton.

I have:

Tested in provider-upjet-gcp.

### DIFF
--- a/cmd/crdmigrator/example.yaml
+++ b/cmd/crdmigrator/example.yaml
@@ -2,8 +2,18 @@
 #
 # SPDX-License-Identifier: CC0-1.0
 
+# Static mode: provide explicit CRD-to-storage-version mappings.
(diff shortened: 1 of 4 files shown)


jonasz-lasut (approved):
Thank you @sergenyalcin lgtm!
One quick question: what's the proposed model of running the dynamic migrations? Do we plan to add it as an init container for the provider, run it on SetupGated or some other approach? I'm not that up to speed with the proposed design

sergenyalcin:
Thanks @jonasz-lasut for your review. As you pointed this is a library in the upjet and integration is related to provider's implementation. But the suggested deployment model is init container approach. I'll open a PR in provider-upjet-gcp and it will show the all picture end-to-end. And also let me add generic guides (in another PR) to upjet for sharing the details about the suggested deployment model and flow.

PR #724, opened 2026-08-06
Add config.Resource.OverrideScalarFieldType

ulucinar:

This PR adds the `config.Resource.OverrideScalarFieldType` configuration method that allows the generated type of a scalar CRD field at a given Terraform path to be overridden. An example invocation is:
```go
func Configure(p *config.Provider) {
	p.AddResourceConfigurator("azuread_application", func(r *config.Resource) {
		r.OverrideScalarFieldType("prevent_duplicate_names", types.NewStringOrBoolType())
                 ...
```

The Terraform field path should omit any wildcards, e.g., `x.y` even `x` is a collection type.

Breaking CRD scalar field type changes in upjet-based providers, such as bool->string or int->string are common, when the underlying Terraform provider makes changes in the Terraform resource's schema. For instance, in [crossplane-contrib/provider-upjet-aws:v1.19.0](https://github.com/crossplane-contrib/provider-upjet-aws/releases/tag/v1.19.0), the Terraform provider changed the type of the `at_rest_encryption_enabled` field from a bool to a nullable bool (which is a string type). The provider did not publish a new CRD API version with the necessary API converters and the Terraform runtime converters (probably because that introduces some extra complexity) and instead documented the breaking changes in its release notes.

It's also a common practice not to introduce new CRD API versions with the major version bumps of the providers when the API versioning policies allow such breaking changes.

However, when the CR field is stored in etcd as a different type than what the CRD Go struct type now declares in a new version of the provider, the UX is seriously broken. The provider simply fails to start its informer cache as it cannot decode the object it reads from etcd, fails with something like:
```
json: cannot unmarshal bool into Go struct field ReplicationGroupParameters.items.spec.forProvider.atRestEncryptionEnabled
```

This PR introduces the internal generic `StringOrPrimitive` type and the public `StringOrBool` type so that provider authors can better handle such breaking changes with a much better and expected UX. When the provider configures (either through manual means, or via the means of the auto-registration framework) a type override for the affected field, the controller-runtime then becomes able to properly decode the (now) broken object into its new memory representation and no fatal errors will be observed. Clients of the affected CRD API will though still need to be migrated to use the new type, this will only prevent a crash loop of the provider.

The existing API clients will fail as expected with the proper API validation errors when they try to use the old field type, until they are migrated to use the new type for the affected field.

**Note**: The current implementation only supports overriding the generated types for scalar fields; collections and composites are currently not supported. It's a build time error to specify a Terraform path that points to a collection or composite when calling `OverrideScalarFieldType`. 

In the future, we may consider adding support for other primitive type overrides and/or adding support for type overrides of collections/composites. Supporting overrides for non-scalars is more complex to handle in upjet's code generation pipelines and is not needed as of now.

I have:

~~- [ ] Added `backport release-x.y` labels to auto-backport this PR if necessary.~~

Tested against `crossplane-contrib/provider-upjet-azure`. The runtime behavior of the provider has been tested successfully for an update with a breaking scalar field type change (bool -> string as discussed above). The provider no longer crashes as intended.

### DIFF
--- a/pkg/config/resource.go
+++ b/pkg/config/resource.go
@@ -7,6 +7,7 @@ package config
 import (
 	"context"
 	"fmt"
+	"go/types"
 	"strings"
 	"time"
 
@@ -707,6 +708,11 @@ type Resource struct {
 	// the Terraform Plugin SDKv2 client.
 	useTerraformPluginFrameworkClient bool
 
+	// overrideGeneratedFieldType allows to manually override the type for the
+	// generated field of a Resource at the specified Terraform path.
+	// We only support type overrides for scalar fields currently.
+	overrideGeneratedFieldType map[string]types.Type
+
 	// OverrideFieldNames allows to manually override the relevant field name to
 	// avoid possible Go struct name conflicts that may occur after Multiversion
 	// CRDs support. During field generation, there may be fields with the same
@@ -1178,6 +1184,33 @@ func (r *Resource) RemoveSingletonListConversion(tfPath string) bool {
 		}
 	}
 	return false
+}
+
+// OverrideScalarFieldType allows to manually override the type for the
+// generated scalar field of a Resource at the specified Terraform path.
+// The path is a Terraform field path without the wildcard segments, e.g.,
+// "x.y", even if "x" is a collection type.
+// We only support overriding types for scalar fields as of now.
+// Trying to override the type generated for a non-scalar path will result in
+// a generation-time error.
+func (r *Resource) OverrideScalarFieldType(path string, t types.Type) {
+	r.overrideGeneratedFieldType[path] = t
+}
+
+// FieldTypeOverrideConfiguration represents a configuration for a set of type
+// overrides at a specific Terraform path.
+type FieldTypeOverrideConfiguration struct {
+	ParameterTypeOverride types.Type
+}
+
+// FieldTypeOverride returns the type override configuration for the specified
+// path. The path is a Terraform field path without the wildcard segments,
+// e.g., "x.y", even if "x" is a collection type.
+// Note: This accessor is meant to be only used by the code generator.
+func (r *Resource) FieldTypeOverride(path string) FieldTypeOverrideConfiguration {
+	return FieldTypeOverrideConfiguration{
+		ParameterTypeOverride: r.overrideGeneratedFieldType[path],
+	}
 }
 
 // SetEmbeddedObject sets the EmbeddedObject for the specified key.
--- a/pkg/config/resource.go
+++ b/pkg/config/resource.go
@@ -11,6 +11,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/crossplane/crossplane-runtime/v2/pkg/errors"
 	"github.com/crossplane/crossplane-runtime/v2/pkg/fieldpath"
 	"github.com/crossplane/crossplane-runtime/v2/pkg/reconciler/managed"
 	xpresource "github.com/crossplane/crossplane-runtime/v2/pkg/resource"
@@ -21,7 +22,6 @@ import (
 	"github.com/hashicorp/terraform-plugin-go/tftypes"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/terraform"
-	"github.com/pkg/errors"
 	"k8s.io/apimachinery/pkg/util/json"
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/utils/ptr"
@@ -1192,9 +1192,22 @@ func (r *Resource) RemoveSingletonListConversion(tfPath string) bool {
 // "x.y", even if "x" is a collection type.
 // We only support overriding types for scalar fields as of now.
 // Trying to override the type generated for a non-scalar path will result in
-// a generation-time error.
-func (r *Resource) OverrideScalarFieldType(path string, t types.Type) {
-	r.overrideGeneratedFieldType[path] = t
+// an error.
+func (r *Resource) OverrideScalarFieldType(path string, t types.Type) error {
+	if r.TerraformResource == nil {
+		return errors.Errorf("resource %q does not have a valid Terraform resource schema", r.Name)
+	}
+	s := GetSchema(r.TerraformResource, path)
+	if s == nil {
+		return errors.Errorf("path %s is not valid for the Terraform resource schema of %q", path, r.Name)
+	}
+	switch s.Type { //nolint:exhaustive // The default case handles the error cases (non-scalar paths) already.
+	case schema.TypeBool, schema.TypeFloat, schema.TypeInt, schema.TypeString:
+		r.overrideGeneratedFieldType[path] = t
+	default:
+		return errors.Errorf("field at path %q with Terraform type %s is not scalar, only scalar field types can be overridden", path, s.Type.String())
+	}
+	return nil
 }
 
 // FieldTypeOverrideConfiguration represents a configuration for a set of type
--- a/pkg/types/builder.go
+++ b/pkg/types/builder.go
@@ -225,7 +225,7 @@ func (g *Builder) buildSchema(f *Field, cfg *config.Resource, names []string, cp
 			return o.ParameterTypeOverride, nil, nil
 		default:
 			return nil, nil, errors.Errorf(
-				"OverrideScalarFieldType at %q is only supported for scalar fields, got Terraform type %s",
+				"field at path %q with Terraform type %s specified for OverrideScalarFieldType is not scalar, only scalar field types can be overridden",
 				cpath, f.Schema.Type.String())
 		}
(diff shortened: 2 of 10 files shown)


pkg/types/builder.go
```diff
@@ -219,6 +219,17 @@ func (g *Builder) AddToBuilder(typeNames *TypeNames, r *resource) (*types.Named,
...
 
 func (g *Builder) buildSchema(f *Field, cfg *config.Resource, names []string, cpath string, r *resource) (types.Type, types.Type, error) { //nolint:gocyclo
+	if o := cfg.FieldTypeOverride(cpath); o.ParameterTypeOverride != nil {
+		switch f.Schema.Type { //nolint:exhaustive // The default case handles the error cases (non-scalar paths) already.
+		case schema.TypeBool, schema.TypeFloat, schema.TypeInt, schema.TypeString:
+			return o.ParameterTypeOverride, nil, nil
+		default:
+			return nil, nil, errors.Errorf(
+				"OverrideScalarFieldType at %q is only supported for scalar fields, got Terraform type %s",
+				cpath, f.Schema.Type.String())
```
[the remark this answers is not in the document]
ulucinar: Done.

pkg/types/internal/string.go
```diff
@@ -0,0 +1,51 @@
...
+	errInvalidValue = "value must be a JSON string or %T, got %s"
+)
+
+// Primitive is a type constraint that represents the supported primitive types
+// for StringOrPrimitive.
+type Primitive interface {
+	~bool |
+		~int | ~int8 | ~int16 | ~int32 | ~int64 |
+		~uint | ~uint8 | ~uint16 | ~uint32 | ~uint64
+}
```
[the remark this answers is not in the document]
ulucinar: Done. Updated the doc comment accordingly.

pkg/types/builder_test.go
```diff
@@ -873,3 +873,95 @@ func TestBuild(t *testing.T) {
...
+							},
+						},
+					},
+				}, nil, nil)
+				r.Kind = ""
+				r.OverrideScalarFieldType("settings", NewStringOrBoolType())
+				return r
+			},
+			want: want{
+				errContains: `OverrideScalarFieldType at "settings" is only supported for scalar fields, got Terraform type TypeList`,
```
sergenyalcin: It looks like the error message changed in the last commit, but this test wasn’t updated accordingly. So the unit tests fail.

pkg/config/resource.go
```diff
@@ -1180,6 +1186,33 @@ func (r *Resource) RemoveSingletonListConversion(tfPath string) bool {
...
+// Trying to override the type generated for a non-scalar path will result in
+// a generation-time error.
+func (r *Resource) OverrideScalarFieldType(path string, t types.Type) {
+	r.overrideGeneratedFieldType[path] = t
+}
+
+// FieldTypeOverrideConfiguration represents a configuration for a set of type
+// overrides at a specific Terraform path.
+type FieldTypeOverrideConfiguration struct {
+	ParameterTypeOverride types.Type
```
sergenyalcin, line 1216: nit: I just wanted to discuss the `ParameterTypeOverride` is the right name here. This override applies to `forProvider`, `initProvider`, and `atProvider`, not just parameters in terms of spec parameters. Maybe I am overthinking here but do you think there might be some confusion for those to use this? Do you think something like `TypeOverride` or `FieldTypeOverride` might better reflect its scope?

pkg/config/resource.go
```diff
@@ -1180,6 +1186,33 @@ func (r *Resource) RemoveSingletonListConversion(tfPath string) bool {
...
 }
 
+// OverrideScalarFieldType allows to manually override the type for the
+// generated scalar field of a Resource at the specified Terraform path.
+// The path is a Terraform field path without the wildcard segments, e.g.,
+// "x.y", even if "x" is a collection type.
+// We only support overriding types for scalar fields as of now.
+// Trying to override the type generated for a non-scalar path will result in
+// a generation-time error.
+func (r *Resource) OverrideScalarFieldType(path string, t types.Type) {
```
sergenyalcin: One thing that came to mind: should we validate that path actually exists in the Terraform schema? Right now a typo would silently become a no-op. It may never match anything during code generation. If we intentionally want to keep this like that, it might at least be worth documenting that invalid paths fail. What do you think?

pkg/config/resource.go
```diff
@@ -1180,6 +1186,33 @@ func (r *Resource) RemoveSingletonListConversion(tfPath string) bool {
...
+	r.overrideGeneratedFieldType[path] = t
+}
+
+// FieldTypeOverrideConfiguration represents a configuration for a set of type
+// overrides at a specific Terraform path.
+type FieldTypeOverrideConfiguration struct {
```
sergenyalcin, line 1215: One thing I thought about is exposing `FieldTypeOverrideConfiguration` as part of the public API. The The doc of the `FieldTypeOverride` field is says that it is only to be used by the code generator, but the returned type is exported. With this state, I am not sure this is just a internal plumbing point. It seems a general use API. What do you think?

I didn't think the alternative for this but still wanted to discuss do we really want to export them for general use instead of really generation internal?

sergenyalcin (commented):
Thanks @ulucinar for this PR. I've left a few comments for discussion.

pkg/types/builder_test.go
ulucinar: :) Done. Sorry for that, it was in response to another review comment about that error message.

pkg/config/resource.go
ulucinar: If, in the future, we decide to add support for overrides for non-scalar field types (complex or collection types), then we will need to extend this `FieldTypeOverrideConfiguration` so that it contains init parameters and observation parameters. The current name of `ParameterTypeOverride` reflects how upjet internally names `spec.forProvider` parameters. Also, this struct is not meant to be consumed by the providers directly, it's only consumed in the code generation pipeline. Normally we would like to hide this struct but Go does not have something like "friend-package" access (the code generation packages need to access this configuration). So, I would argue we may go towards something more specific like `ForProviderParameterTypeOverride` but considering the existing naming, I would keep it as is.

pkg/config/resource.go
ulucinar: > One thing that came to mind: should we validate that path actually exists in the Terraform schema?

I definitely agree we should do that. I considered implementing a validation in addition to the scalar type check we already have but the issue is the available canonical paths are not readily available to us. This is also the reason we don't have such checks for other similar configuration options like `config.Resource.AddSingletonListConversion` and others. I was thinking we could try to address this cross-cutting validation concern in a future PR, because this already surfaced in this PR's discussions, let me check what we can do about it...

pkg/config/resource.go
ulucinar: I was initially thinking exposing the `config.Resource.overrideGeneratedFieldType` (as `config.Resource.OverrideGeneratedFieldType`) so that we would not need a separate `config.Resource.OverrideScalarFieldType` configuration method. However, in the future, we may need to support overriding types of non-scalar (complex/collection) fields and if we simply expose `overrideGeneratedFieldType` as our configuration API, it also directly exposes upjet code generation pipeline's implementation details. 

Currently, the only "intended" and exposed configuration API is the `config.Resource.OverrideScalarFieldType`, which is what a provider needs to specify to be able to override the generated type for a scalar field, the field's path and the type to be used for the override. Supporting overrides for complex types or collection types in more involved because such fields can differ across the forProvider, initProvider and atProvider API trees. For scalars, they always have the same exact type (and that's the reason the `OverrideScalarFieldType` method accepts a single type parameter, not separate types for the 3 API trees we have). If we had exposed `overrideGeneratedFieldType` directly, it would not be future proof but the `OverrideScalarFieldType` is future proof because it's not a generic override API, i.e., it's not expected to handle non-scalar field type overrides. Also having the configured performed via `OverrideScalarFieldType` (instead of directly setting the configuration data structures) would allow us to do some validation in the future.

The cost of hiding the implementation detail from the provider authors is then we need to expose that configuration via an accessor because the code generation pipeline that consumes this configuration lives in the package `pkg/types`. Go does not have something like a "friend package" concept and an `internal` package won't help because `overrideGeneratedFieldType` is not exported. `FieldTypeOverrideConfiguration` and `FieldTypeOverride` should be of no use to the provider authors but we cannot also hide them from the provider authors. That's why I just added a Go doc that tells `FieldTypeOverride` is meant to be used by the code generator.

sergenyalcin (approved):
LGTM @ulucinar

pkg/config/resource.go
ulucinar: Added 3 assertions in `config.Resource.OverrideScalarFieldType`:
- The `config.Resource` has actually a Terraform resource schema associated (`nil` check)
- The specified path exists
- And the specified path is actually a scalar field's path

We still keep the scalar path assertion in the code-generation pipeline itself. We may have the resource schema mutated by a schema traverser or by another mutator, though we've effectively moved the generation-time validation to the "configuration time".

PR #727, opened 2026-08-09
fix(types): make secret reference fields optional in initProvider

### LINKED ISSUE #456: Secret in InitProvider is required but should be optional
The Property IdentityProviderSpec.InitProvider.ClientSecretSecretRef
(see https://github.com/crossplane-contrib/provider-keycloak/blame/11f3432fc6cbf1f12168f6e3ba425f79ad0d5710/apis/oidc/v1alpha1/zz_identityprovider_types.go#L55-L57)

is a required field since updating upjet to to v1.4.0
(see https://github.com/crossplane/upjet/releases/tag/v1.4.0 -> Support for Secret References in spec.initProvider)

InitProvider properties are set if the value of the property should only be taken into account during creation and not during update
(see https://docs.crossplane.io/latest/concepts/managed-resources/#initprovider)

But from Keycloaks point of view the client secret of an OIDC IdentityProvider is NOT immutable.

It should be possible to set ClientSecretSecretRef via forProvider and skip setting it at the initProvider

Additional Notes:
forProvider defines the field "ClientSecretSecretRef" also as required.
Which leads to the situation that I have to specify both (initProvider and forProvider), which might run into the following issue: https://github.com/crossplane/upjet/issues/299

ref: https://github.com/crossplane-contrib/provider-keycloak/issues/113

pujitha24:
Fixes #456

I have:

Extended the table-driven `TestBuild` cases `Sensitive_Fields` and `Sensitive_Fields_namespaced` in `pkg/types/builder_test.go` to assert on the generated `initProvider` struct, not just `forProvider`. Confirmed via `git stash` on `pkg/types/field.go` alone that these assertions FAIL against the pre-fix code and PASS with the fix applied:

```
go test ./pkg/types/... -run TestBuild -v
```

Also ran, all passing:

```
go build ./...
go vet ./pkg/...
go test ./pkg/...
```

---
AI assistance: this change was drafted with Claude Code.

### DIFF
--- a/pkg/types/builder_test.go
+++ b/pkg/types/builder_test.go
@@ -309,6 +309,7 @@ func TestBuild(t *testing.T) {
 	}
 	type want struct {
 		forProvider     string
+		initProvider    string
 		atProvider      string
 		validationRules string
 		err             error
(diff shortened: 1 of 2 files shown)


pujitha24:
This has been sitting green and rebased for a couple weeks now — happy to make any changes that would help move review along.

